PR #2, opened 2021-02-09
Minor checks

abreis:
Hi Jorge, I fixed a few things here and there while going through your proposal, mostly spelling.

I'm too new to the `arrow` codebase to be able to tell the second-order effects of what you propose here, so I have no technical comments to offer. That said, I found myself in agreement with everything you argue.
 
Hope it gets accepted. Cheers


jorgecarleitao:
Thanks a lot for taking the time to read and PR this, @abreis  :)

PR #6, opened 2021-03-10
Macro natives and linting

elferherrera:
While I was studying you code I found several lint hints that could be applied to your code. I hope you find them useful and want to integrate to your branch


jorgecarleitao:
Hi @elferherrera ; thanks a lot! Could you describe the process over which you found them? It would be advantageous to place it on the CI.

elferherrera:
If you are using vscode and rust analyzer you will get these suggestions for free. I used to use nvim but I couldn't get the suggestions to work as well as they work in vscode. In the past month I have been coding with vscode and it has helped me a lot. 

You can find the same suggestions using `cargo clippy`. In the code there are still 25 warnings if you run clippy. I didn't change them all because I think you would need to see if they make sense. These include functions with too many arguments, or to replace `while let` for a `for` loop, or to add safety explanations in unsafe code.

PR #8, opened 2021-03-14
Dense tensor definition

elferherrera:
Initial definition of dense tensor. It includes constructor and functions to interact with it

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 target
 venv
+
+.vscode/
--- a/src/tensor/dense.rs
+++ b/src/tensor/dense.rs
@@ -1,0 +1,492 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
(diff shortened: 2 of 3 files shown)


jorgecarleitao:
Hi @elferherrera , thanks a lot for this! 

I looked it up for integration files to perform integration tests, but I have not found any. I can see that the arrow crate itself also has no integration tests for tensors, which suggests that this feature is not well maintained.

A concern I have is that we would offer tensor support, but then have no process of verifying its compatibility against other implementations, which IMO is the hallmark of the arrow format. It is true that the arrow crate itself has no support, but without integration tests, there is no proof that what it is done there is in spec, right?

Do you have an idea on how we should handle this?

elferherrera:
Hi. Actually that is something I want to write about in the mailing list. I know they have dense and sparse tensor definitions in C++ implementation but I couldn't find how they are testing it. I have an idea of how to add it to the new IPC format you have but I was wondering how we can test it with the C++ side.

I guess it will have to be a joint effort with the C++ team, unless they are not interested in this feature.

elferherrera:
What do you think we should do with these tensors?. I would like to write more implementations for dense and sparse tensors, especially to be able to use them with IPC. However, I dont think they will be consider them soon into main Arrow.

jorgecarleitao:
Firstly, thanks a lot for driving that in the mailing list..

I think that we should not have added them to the master arrow crate in the first place, as there is no guarantee that they can be used throughout the ecosystem. IMO we should not invest time on them without IPC tests, as we risk drifting from the specification, which IMO is detrimental to the arrow project as a whole.

I think that the path forward in this direction is to write a google docs describing the json-integration for tensors and share them on the mailing list, so that folks from C++ and the spec itself can take a look and see whether the format would make sense. Note that these are very simple files in json. E.g. the validity is just "true" / "false".

PR #30, opened 2021-03-24
Fix typos in guide

Jibbow:
Me again, haha :)

When reading the guide I noticed some typos. This PR fixes them.

Let's see if the new GitHub Actions stuff works as expected ;)

### DIFF
--- a/guide/book.toml
+++ b/guide/book.toml
@@ -3,4 +3,4 @@ authors = ["Jorge C. Leitao"]
 language = "en"
 multilingual = false
 src = "src"
-title = "arrow 2 documentation"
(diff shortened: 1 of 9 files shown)


Jibbow:
Also, if you want, I could put some more work into the guide. Like, not a full rewrite, but just some improvements here and there.
The crate itself looks so great, but it feels like the guide falls a bit behind.

jorgecarleitao:
Thanks @Jibbow ! Sure, fire it away. :) The JSON read and write may still receive a re-design; other parts are stabler at this point and thus benefit from better documentation for sure and the guide is a great place to put it.

PR #32, opened 2021-04-03
Primitive arithmetic macro

elferherrera:
Definition of macro for primitive arithmetic matching. The macro is used to downcast `lhs` and `rhs` arrays to a primitive array type and executed the primitive arithmetic function with the results.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -20,8 +20,6 @@ pub enum ArrowError {
     /// Error during import or export to/from a format
     ExternalFormat(String),
     DictionaryKeyOverflowError,
-    /// Error produced when unable to downcast an array
-    DowncastError(String),
     Other(String),
 }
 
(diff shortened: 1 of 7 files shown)


src/error.rs
```diff
@@ -20,6 +20,8 @@ pub enum ArrowError {
     /// Error during import or export to/from a format
     ExternalFormat(String),
     DictionaryKeyOverflowError,
+    /// Error produced when unable to downcast an array
+    DowncastError(String),
```
jorgecarleitao, line 24: Note that a downcast error never happens by design: we validate that an array has the right `DataType` on all constructors, so the error will appear before (when the user tries to initialize an array with a DataType that is invalid). This is different from `arrow`, where no such validation is currently made.

My suggestion is that we keep these as panics, as they represent a development error on our side. there is also nothing that the users of this library can do much about...

jorgecarleitao (commented):
Great stuff. Lots of less code! Thanks a lot! :)

Let a comment that IMO we should address. Other than that, ready to :shipit: :)

src/error.rs
elferherrera: I see. It makes sense. Let me change them to panics. Do you think we should keep the DowncastError in the Enum or it should be removed as well?

jorgecarleitao: Let's remove them: this way it is clear to end users that that is not an error that we emit.

PR #33, opened 2021-04-04
Add support to read from parquet

### LINKED ISSUE #21: Add support to read from Parquet
* Create new module `io/parquet`
* add dependency on the `parquet` crate without `arrow` feature
* add conditional feature `parquet` and feature-gate `io/parquet` with it
* re-write the `arrow/` section of parquet's crate for this API

This is a large task ^_^

Note that parquet's crate has non-trivial and potentially unsound signatures of the physical types. E.g. `INT96` is represented as `Option<[u32; 3]>` which has 16, not 12 bytes. We should spend some time investigating if a re-write of that is needed to e.g. pass MIRI checks and be generally safe and sound.

jorgecarleitao:
This PR adds initial support to read from parquet.

This (and the whole [parquet2](https://github.com/jorgecarleitao/parquet2)) does not use `unsafe` and is ~10-20x faster than the official `parquet` crate. It also scales better with increasing size.

It supports all physical types but a very limited number of logical types. V2 pages are also not yet supported. It does support the two most common encodings (dict and RLE-bitpacking hybrid).

This PR also:

* adds Python code to generate parquet files from pyarrow, for the purposes of integration testing and benchmarking.
* adds interoperability tests against `pyarrow==3` for `i64`, `f64`, `bool` and `utf8`.
* adds a bench to read from parquet
* adds Python to benchmark reading a parquet from pyarrow
* (`parquet2`, really) opens the possibility to deserialize parquet files in parallel, as it decouples the reading of parquet pages (IO-intensive) from deserialization (CPU-intensive).

Some benchmarks (reading 10^4 and 10^5 with null density ~40%) generated by pyarrow:

parquet+arrow (master):
```
read u64 10000          time:   [217.26 us 217.41 us 217.58 us]
read u64 100000         time:   [1.6632 ms 1.6761 ms 1.6876 ms]
read utf8 10000         time:   [611.50 us 611.92 us 612.34 us]
read utf8 100000        time:   [5.7691 ms 5.7763 ms 5.7836 ms]
```

This PR:
```
read u64 10000          time:   [81.456 us 81.552 us 81.650 us]
read u64 100000         time:   [138.43 us 138.54 us 138.64 us]
read utf8 10000         time:   [89.185 us 89.281 us 89.380 us]
read utf8 100000        time:   [274.32 us 274.49 us 274.68 us]
```

Using a brute `timeit` on pyarrow yields 

```
read u64 100000     time: 1.540 ms
read utf8 100000    time: 3.532 ms
```

This is not 100% fair as `parquet` also reads the arrow schema and pyarrow has some python overhead, but IMO those do not explain the significant difference.

This purposely does not read row groups to `RecordBatch`es, as that operation can be trivially parallelizable across columns and should thus be handled downstream (DataFusion, Polars, etc.), according to their respective execution model (async, custom scheduler, rayon, etc.).

Closes #21

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -66,6 +66,13 @@ jobs:
           rustup toolchain install stable
           rustup default stable
           rustup component add rustfmt
+      - name: Setup parquet files
+        run: | 
+          apt update && apt install python3-pip python3-venv -y -q
+          python3 -m venv venv
+          venv/bin/pip install pip --upgrade
+          venv/bin/pip install pyarrow==3
+          venv/bin/python parquet_integration/write_parquet.py
       - name: Run
         run: |
           export CARGO_HOME="/github/home/.cargo"
@@ -91,7 +98,8 @@ jobs:
           rustup component add rustfmt
       - name: Run
         shell: bash
-        run: cargo test
+        # no need to run the whole thing. Rust guarantees interoperability
+        run: ARROW2_IGNORE_PARQUET= cargo test --lib
 
   clippy:
     name: Clippy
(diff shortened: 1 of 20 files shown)


jorgecarleitao:
fyi @nevi-me @Dandandan @alamb @andygrove, as this has been a long standing issue in DataFusion.

Dandandan:
Thanks @jorgecarleitao awesome results 😎

How do you think this work could be integrated into DataFusion? Any part that could be reused 1:1 in the parquet crate?

alamb:
> Thanks @jorgecarleitao awesome results 😎

I agree -- nice work @jorgecarleitao  

I would love to see some of this coolness making it back into `parquet` or DataFusion

PR #37, opened 2021-04-10
Adding and subtracting duration to timestamp

### LINKED ISSUE #26: Add support for arithmetics between naive timestamps and durations

elferherrera:
This PR defines functions to add and subtract a duration to a timestamp. The duration is always scaled up/down to have the same time scale as the timestamp.

Closes #26

### DIFF
--- a/src/compute/arithmetics/timestamp.rs
+++ b/src/compute/arithmetics/timestamp.rs
@@ -32,7 +32,7 @@ use crate::{
 };
 
 /// Calculates the scale factor between two TimeUnits. The function returns the
-/// number that should multiply the TimeUnit "b" to have at the same scale as
+/// scale that should multiply the TimeUnit "b" to have the same time scale as
 /// the TimeUnit "a".
 fn timeunit_scale(a: &TimeUnit, b: &TimeUnit) -> f64 {
     match (a, b) {
--- a/src/compute/arithmetics/timestamp.rs
+++ b/src/compute/arithmetics/timestamp.rs
@@ -1,0 +1,351 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! Defines the arithmetic kernels for Timestamp and Duration.
+//! For the purposes of Arrow Implementations, adding this value to a Timestamp
+//! ("t1") naively (i.e. simply summing the two number) is acceptable even
+//! though in some cases the resulting Timestamp (t2) would not account for
+//! leap-seconds during the elapsed time between "t1" and "t2".  Similarly,
+//! representing the difference between two Unix timestamp is acceptable, but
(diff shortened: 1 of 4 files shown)


src/compute/arithmetics/timestamp.rs
```diff
@@ -0,0 +1,351 @@
...
+    array::{Array, PrimitiveArray},
+    compute::arity::unchecked_binary,
+    datatypes::{DataType, TimeUnit},
+    error::{ArrowError, Result},
+};
+
+/// Calculates the scale factor between two TimeUnits. The function returns the
+/// scale that should multiply the TimeUnit "b" to have the same time scale as
+/// the TimeUnit "a".
+fn timeunit_scale(a: &TimeUnit, b: &TimeUnit) -> f64 {
```
jorgecarleitao, line 37: What do you think about moving this to the `temporal_conversions` module? Seems related.

src/compute/arithmetics/timestamp.rs
```diff
@@ -0,0 +1,351 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
jorgecarleitao, line 1: Could we remove this header? This is by default Apache 2.0 as per README; Let's make it ASF if/when this gets donated.

jorgecarleitao (commented):
Great work, really simple and easy to follow. Great coverage also :D

I left some comments.

src/compute/arity.rs
```diff
@@ -117,3 +117,41 @@ where
         validity,
     ))
 }
+
+// The types of the arrays are not checked with this operation. The closure
+// "op" needs to handle the different types in the arrays. The datatype for the
+// resulting array has to be selected by the implementer of the function as
+// an argument for the function.
+#[inline]
+pub fn unchecked_binary<T, F>(
```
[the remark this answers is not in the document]
elferherrera: Yeap... it makes more sense to remove the check from binary. I wasn't sure if you weren't using the check for binary in one test. Btw, should the datatype be inferred by the first array or be given as an argument?

elferherrera:
@jorgecarleitao I changed the binary function to avoid arrays checks. If you agree with this changes, do you mind merging this to main so I can merge these changes into Decimal PR?

src/compute/arithmetics/timestamp.rs
```diff
@@ -0,0 +1,403 @@
...
+/// subtracting seconds with seconds, or milliseconds with milliseconds.
+pub fn timestamp_diff(
+    lhs: &PrimitiveArray<i64>,
+    rhs: &PrimitiveArray<i64>,
+) -> Result<PrimitiveArray<i64>> {
+    // Matching on both data types from both arrays.
+    // Both timestamps have a Timeunit enum in its data type.
+    // This enum is used to adjust the scale between the timestamps.
+    match (lhs.data_type(), rhs.data_type()) {
+        (DataType::Timestamp(timeunit_a, _), DataType::Timestamp(timeunit_b, _)) => {
```
jorgecarleitao, line 79: I am not sure, I think that this is only valid for naive (i.e. `None`) timezones. For time-aware, I think that we need to take into account the summer times and all that complexity here.

elferherrera: mmm... that's right. I think it will need to use crono to consider the timezones. To make this difference time-aware we would have to create a larger PR. 

I created this function thinking it may be useful to have it, but if you think we should wait until we have better management of timezones, I can remove it.

jorgecarleitao: (I like it, I would just use `None` there for now, and handle the timezone problems in another PR)

elferherrera: Cool. I left only the case when both timezones are `None` and that is tested as well. This could be the template function to work with more complex timezones.

src/compute/arity.rs
```diff
@@ -88,7 +92,7 @@ where
 {
     if lhs.data_type() != rhs.data_type() {
```
jorgecarleitao, line 93: should we be consistent here and also not check this one?

elferherrera: I was going to change that one with the Decimals PR, but now it can be done before merging this branch to Decimals

PR #39, opened 2021-04-19
Basic arithmetics kernels

elferherrera:
This PR defines functions for basic arithmetics on PrimitiveArrays. It includes Saturating, Overflowing and Checked functions for the add, sub, mul and div operations. It also includes documentation and examples on these operations. 

Extra documentation was added to the compute and arithmetics modules. These comments adds description for the design principles followed when designing the arithmetics kernels. 

Documentation examples were added to the time and decimal modules to be consistent with the basic module

### DIFF
--- a/src/compute/mod.rs
+++ b/src/compute/mod.rs
@@ -15,6 +15,55 @@
 // specific language governing permissions and limitations
 // under the License.
 
+//! Kernel operations for arrow arrays
+
+//! # Compute module
+//!
+//! The compute module is composed by independent operations common in
+//! analytics
+//!
+//! ## Design principles
+//!
+//! Below are the design principles followed to define the compute kernels:
+//!
+//! * APIs MUST return an error when either:
+//!     * The arguments are incorrect
+//!     * The execution results in a predictable error (e.g. divide by zero)
+//!
+//! * APIs MAY error when an operation overflows (e.g. `i32 + i32`)
+//!
+//! * kernels MUST NOT have side-effects
+//!
+//! * kernels MUST NOT take ownership of any of its arguments (i.e. everything
+//!   must be a reference).
+//!
+//! * APIs SHOULD error when an operation on variable sized containers can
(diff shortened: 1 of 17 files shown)


src/compute/mod.rs
```diff
@@ -15,6 +15,55 @@
...
+//! # Compute module
+//!
+//! The compute module is composed by independent operations common in
+//! analytics
+//!
+//! ## Design principles
+//!
+//! Below are the design principles followed to define the compute kernels:
+//!
+//! * APIs MUST return an error when either:
```
jorgecarleitao, line 29: I think that this is repeated with the README.md, right? IMO we should either place it in the readme or here, but not in both, as they will drift apart when someone will inevitable forgets to change one and not the other.

I am not sure these are useful to end users, as they refer to design principles using when writing new kernels, not so much when using them (which IMO the example and docs specific of the kernel are more useful).

jorgecarleitao (approved):
🤯🤯🤯🤯🤯 Really impressive, @elferherrera . This is ready to go. I have a minor comment, but the implementation is really great, and well tested

src/compute/mod.rs
elferherrera: I know what you mean. I was looking for a way to read MD files when doing cargo doc. I haven't invested much time in doing so. 

For me it makes sense to add this to the `crates.io` documentation. I think the richer that documentation the better and easier to find. When we put it in the README might be a bit harder for people looking into the documentation. Again, Im thinking as if this repo is already in `crates.io`. But I leave it to you, I can later find a way to add the README file via cargo docs

jorgecarleitao: I see your point; I agree that the documentation benefits from some of this information, but IMO needs some curation that is beyond this PR. I suggest we focus this PR on the arithmetics and address this on a separate issue.

elferherrera: No problem. Let me remove that section from the code.

PR #42, opened 2021-04-21
Auto-vectorized sum (17x)

jorgecarleitao:
This adds support for SIMD for `NativeType` and demonstrates its usefulness in vectorizing the non-null sum of a primitive array (N=65536):

Before:
```
add f32                 time:   [103.06 us 103.67 us 104.26 us]
```

After:
```
add f32                 time:   [6.1983 us 6.2450 us 6.2885 us]
                        change: [-94.038% -93.994% -93.951%] (p = 0.00 < 0.05)
```

`arrow` with SIMD (nightly):

```
sum 65536               time:   [10.691 us 10.809 us 10.936 us]
```


jorgecarleitao:
@Dandandan @elferherrera , it seems that we can achieve a lot in stable Rust, by just helping a bit the compiler.

elferherrera:
@jorgecarleitao that looks really good.

I was looking at that in the `from_trusted_len_iter` function that we use in the arithmetics function. I was wondering how to create an array from the iterator and implement the TryInto trait.

jorgecarleitao:
I tried this in the arithmetics, but I got worse (+155%) performance. My hypothesis is that we are already auto-vectorizing the arithmetics (it is faster than `arrow` SIMD atm).

Dandandan:
Really cool! I saw earlier there are also a few crates doing something like this. Would be nice to see what it can do together with `multiversion` (to utilize sse4/abc instructions)!

elferherrera:
@jorgecarleitao so, this vectorisation is useful when doing it with operations that only take one vector. 

How did you implement this on the arithmetics module? did you create another function or did you use the `from_trusted_iter` function?

Dandandan:
@jorgecarleitao 

With `multiversion` I'm getting a further 40% speed up compared to this branch


```
Benchmarking add f32: Collecting 100 samples in estimated 5.0363 s (162k iterati                                                                                add f32                 time:   [31.142 us 31.178 us 31.217 us]
                        change: [-38.585% -38.006% -37.435%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Diff:


```
diff --git a/Cargo.toml b/Cargo.toml
index fd8592e..5af9e1d 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,8 @@ itertools = { version = "^0.10", optional = true }
 
 base64 = { version = "0.13.0", optional = true }
 
+multiversion = "0.6.1"
+
 [dependencies.parquet2]
 git = "https://github.com/jorgecarleitao/parquet2"
 rev = "fb65dcaf7d02c7e0bcf8ec0aa85d3fe0cb0630c6"

diff --git a/src/compute/aggregate/mod.rs b/src/compute/aggregate/mod.rs
index e35f191..a00a2d1 100644
--- a/src/compute/aggregate/mod.rs
+++ b/src/compute/aggregate/mod.rs
@@ -21,6 +21,7 @@ use std::{iter::Sum, ops::AddAssign};
 
 use crate::array::{ord::total_cmp, Array, BooleanArray, Offset, PrimitiveArray, Utf8Array};
 use crate::types::NativeType;
+use multiversion::multiversion;
 
 /// Helper macro to perform min/max of strings
 fn min_max_string<O: Offset, F: Fn(&str, &str) -> bool>(
@@ -176,7 +177,8 @@ pub fn max_boolean(array: &BooleanArray) -> Option<bool> {
         .flatten()
         .or(Some(false))
 }
-
+#[multiversion]
+#[clone(target = "x86_64+avx+avx2")]
 fn nonnull_sum<T: NativeType + AddAssign + Sum>(values: &[T]) -> T {
     let chunks = values.chunks_exact(T::LANES);
     let remainder = chunks.remainder();
```

jorgecarleitao:
@elferherrera , I think that this is related to the [horizonallity of the operation](https://rust-lang.github.io/packed_simd/perf-guide/vert-hor-ops.html), which is often more difficult for the compiler to SIMD it.

jorgecarleitao:
@Dandandan amazing :D we just shaved about 2x vs arrow SIMD, in stable Rust :)

PR #43, opened 2021-04-21
Corrected write parquet benchmark

elferherrera:
There was a missing argument in the write_parquet bench


jorgecarleitao:
Thanks a lot, @elferherrera !

PR #44, opened 2021-04-22
Simd sum null

jorgecarleitao:
Same spirit as #42 . This requires SIMD masks, which we use `uX` to do.

Before:
```
add nulls f32           time:   [324.46 us 325.94 us 327.32 us]
```

after:
```
add nulls f32           time:   [49.267 us 49.499 us 49.718 us]
```

Arrow SIMD:
```
sum nulls 512           time:   [10.863 us 10.914 us 10.966 us]
```

Arrow has an out of bound read on `chunks`, so this comparison is not 100% fair. Still, I was unable to reach the same perf. as the current Arrow implementation by a factor of 5x (yet).


Dandandan:
Impressive! I found the crate with the same idea of using auto vectorization mentioned in the other PR:

https://docs.rs/slipstream/0.2.0/slipstream/

Maybe there is also some cool inspiration to find there.

jorgecarleitao:
> Impressive! I found the crate with the same idea of using auto vectorization mentioned in the other PR:
> 
> https://docs.rs/slipstream/0.2.0/slipstream/
> 
> Maybe there is also some cool inspiration to find there.

Really good finding, @Dandandan ; I agree. I will ping the owner :)

PR #48, opened 2021-04-23
Arithmetic trait

elferherrera:
This PR introduces traits for basic arithmetic operations that gives more functionality to the primitive arrays

### DIFF
--- a/src/compute/arithmetics/basic/add.rs
+++ b/src/compute/arithmetics/basic/add.rs
@@ -9,8 +9,13 @@ use num::{
 use crate::{
     array::{Array, PrimitiveArray},
     bitmap::Bitmap,
-    compute::arity::{
-        binary, binary_checked, binary_with_bitmap, unary, unary_checked, unary_with_bitmap,
+    compute::{
+        arithmetics::{
+            ArrayAdd, ArrayCheckedAdd, ArrayOverflowingAdd, ArraySaturatingAdd, NotI128,
+        },
+        arity::{
+            binary, binary_checked, binary_with_bitmap, unary, unary_checked, unary_with_bitmap,
+        },
     },
     error::{ArrowError, Result},
     types::NativeType,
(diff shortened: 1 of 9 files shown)


jorgecarleitao:
Hi.

Thanks a lot for this PR 👍 

Note that this goes against the strict separation of "data" and "compute" that all arrow libraries have.

The reason arrays do not have these traits is that this library is not a "frontend library"; it is not meant to declare a public API to easily write analytical code. We delegate that to DataFusion, Polars, etc.

Introducing these traits will start to blur this separation, as we are offering an opinionated (albeit very reasonable based on the `std`) view of an API.

I am not against it, because, like I said, this is very aligned with `std`, just interested in your opinion here.

elferherrera:
I see your point and understand this separation from data and compute, and I think defining these types of traits enforce that separation by making clear how to operate on the arrays. It is the same reason as why the computation kernels are defined within Arrow. If Arrow were purely a `data` crate, the kernels should be defined somewhere else.

I thought about these traits during the week thinking that they may allow users to create generic implementations with arrays. Instead of each crate (polars, or even ndarray) defining their own traits for their generic operations, Arrow can start defining these abstractions for the final users. I'm sure some of these traits could be reused and improved by developers using the Arrow implementation, enriching the whole ecosystem. The other alternative is to have a `arrow-num` crate, which could be similar to the `num` crate.

jorgecarleitao:
So, my reasoning is the following: all methods over arrays are `O(1)`; most compute operations are `O(N)`. This is what I meant with the separation of "compute" and "data". In std Rust, compute (e.g. `u64 + u64`) is `O(1)`.

AFAI understand, the main rational for this PR is that it offers a compute frontend to the arrays. IMO that frontend does not need to live in this crate. OTOH, I also think that this frontend is so natural that it will hardly be contested.

Would it be ok if we leave this here for some time, to mature the thoughts, before we commit to a decision?

elferherrera:
I think it will make sense for the project to share these interfaces for people wanting to use arrow. For example, I wouldn't see the point of `polars` and `ndarray` to implement the same trait in different projects. I am looking at this from the perspective of creating a unified API for all projects using arrow. 

We can keep it on hold until this crate has reached a more mature status. It still needs to be accepted by the other implementations.

elferherrera:
Do you mind if this PR is closed?

jorgecarleitao:
I was still planning to merge this :)

jorgecarleitao:
I kind of got convinced, mostly because people will still need to import the trait. I was not fully convinced because in other languages the method gets bounded automatically on every file. But in Rust, only when the person imports the trait it is bound to the object, which IMO is a good safeguard as it does not limit others from implementing a competing method.

I say we rebase and merge

PR #49, opened 2021-04-24
Removed unsafe code

jorgecarleitao:
This PR refactors `BitChunk` and `BitChunkIterator` APIs, removing all `unsafe` on them.

Also added a battery of tests since they were lacking.

No change in performance.

Also made the iterator endianess-independent, which brings the crate closer to support big endianess.


jorgecarleitao:
@vertexclique, this was something that we discussed long ago in the arrow crate; I think that this addresses that old discussion while still allowing the compiler to auto-vectorize.

PR #58, opened 2021-04-25
Moved function and other minor API improvements

jorgecarleitao:


jorgecarleitao:
@elferherrera , I moved the `neg` out of sub, for easier discoverability.

PR #60, opened 2021-04-25
Fix a couple broken links in the guide

abreis:
Fixes two broken links that I noticed, and since the intention seems to be to use `low-level / high-level API` instead of `low-end / high-end API` (which IMO is correct), I also updated a few section headers, in a separate commit.


jorgecarleitao (approved):
Thanks, @abreis !

PR #74, opened 2021-05-09
Take command for ListArray

### LINKED ISSUE #52: Add take for `ListArray`

elferherrera:
Proposal for take command for ListArray. The next items were added:

- Take command for ListArrays
- Test with normal and nested ListArrays.
- Check for nested level when creating a ListArray.
- Add option for List and LargeList in main Take command
- Apply suggestions found by clippy (remove unneeded unsafe and unused imports)

closes #52

### DIFF
--- a/src/array/list/mod.rs
+++ b/src/array/list/mod.rs
@@ -47,7 +47,12 @@ impl<O: Offset> ListArray<O> {
         check_offsets(&offsets, values.len());
 
         // validate data_type
-        let _ = Self::get_child(&data_type);
+        let child_data_type = Self::get_child(&data_type);
+        assert_eq!(
+            child_data_type,
+            values.data_type(),
+            "The child's datatype must match the inner type of the \'data_type\'"
+        );
 
         Self {
             data_type,
@@ -211,5 +216,52 @@ mod tests {
             format!("{}", array),
             "ListArray[\nInt32[1, 2],\nInt32[],\nInt32[3],\nInt32[4, 5]\n]"
         );
+    }
+
+    #[test]
+    #[should_panic(
+        expected = "The child's datatype must match the inner type of the \'data_type\'"
+    )]
+    fn test_nested_panic() {
+        let values = Buffer::from([1, 2, 3, 4, 5]);
+        let values = PrimitiveArray::<i32>::from_data(DataType::Int32, values, None);
+
+        let data_type = ListArray::<i32>::default_datatype(DataType::Int32);
+        let array = ListArray::<i32>::from_data(
+            data_type.clone(),
+            Buffer::from([0, 2, 2, 3, 5]),
+            Arc::new(values),
+            None,
+        );
+
+        // The datatype for the nested array has to be created considering
+        // the nested structure of the child data
+        let _ =
+            ListArray::<i32>::from_data(data_type, Buffer::from([0, 2, 4]), Arc::new(array), None);
+    }
+
+    #[test]
+    fn test_nested_display() {
+        let values = Buffer::from([1, 2, 3, 4, 5, 6, 7, 8, 9, 10]);
+        let values = PrimitiveArray::<i32>::from_data(DataType::Int32, values, None);
+
+        let data_type = ListArray::<i32>::default_datatype(DataType::Int32);
+        let array = ListArray::<i32>::from_data(
+            data_type,
+            Buffer::from([0, 2, 4, 7, 7, 8, 10]),
+            Arc::new(values),
+            None,
+        );
+
+        let data_type = ListArray::<i32>::default_datatype(array.data_type().clone());
+        let nested = ListArray::<i32>::from_data(
+            data_type,
+            Buffer::from([0, 2, 5, 6]),
+            Arc::new(array),
+            None,
+        );
+
+        let expected = "ListArray[\nListArray[\nInt32[1, 2],\nInt32[3, 4]\n],\nListArray[\nInt32[5, 6, 7],\nInt32[],\nInt32[8]\n],\nListArray[\nInt32[9, 10]\n]\n]";
+        assert_eq!(format!("{}", nested), expected);
     }
 }
--- a/src/compute/take/list.rs
+++ b/src/compute/take/list.rs
@@ -1,0 +1,217 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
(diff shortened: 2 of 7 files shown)


src/array/list/mod.rs
```diff
@@ -46,8 +46,13 @@ impl<O: Offset> ListArray<O> {
     ) -> Self {
         check_offsets(&offsets, values.len());
 
-        // validate data_type
-        let _ = Self::get_child(&data_type);
+        let nest_level = Self::get_nested_level(&data_type);
+        let expected_data_type = Self::default_datatype(values.data_type().clone());
+        let expected_nest_level = Self::get_nested_level(&expected_data_type);
```
jorgecarleitao, line 51: I think that this should be more complete:

```rust
        // validate data_type
        let data_type = Self::get_child(&data_type);
        assert_eq!(data_type, values.data_type(), "The child's datatype must match the inner type of the \'data_type\'");
```

as it avoids recursing and also proves that not only the levels are right, but that the type itself is right.

src/compute/take/list.rs
```diff
@@ -0,0 +1,214 @@
...
+/// `take` implementation for ListArrays
+pub fn take<I: Offset, O: Offset>(
+    values: &ListArray<I>,
+    indices: &PrimitiveArray<O>,
+) -> Result<ListArray<I>> {
+    let usize_indices: Result<Vec<usize>> = indices
+        .values()
+        .iter()
+        .map(|i| maybe_usize::<O>(*i))
+        .collect();
```
jorgecarleitao, line 37: I _think_ that we can remove this `.collect` and merge this iteration with the iteration below, thus avoiding an extra allocation?

src/compute/take/list.rs
```diff
@@ -0,0 +1,214 @@
...
+
+    let arrays: Vec<ListArray<I>> = usize_indices?
+        .into_iter()
+        .map(|index| values.slice(index, 1))
+        .collect();
```
jorgecarleitao, line 42: I think that this collect can also be removed, and `collect` the values directly to `let array_ref: Vec<&dyn Array> = ... .collect()`.

src/compute/take/list.rs
```diff
@@ -0,0 +1,214 @@
...
+
+    let mut array_ref: Vec<&dyn Array> = Vec::with_capacity(indices.len());
+    for array in arrays.iter() {
+        array_ref.push(array);
+    }
+
+    let mut growable: GrowableList<I> = GrowableList::new(&array_ref, true, 0);
+
+    for index in 0..indices.len() {
+        if let Some(validity) = indices.validity() {
```
jorgecarleitao, line 52: Maybe branch off before the `for` loop and initialize `growable` with `true` or `false` accordingly?

jorgecarleitao (commented):
Amazing implementation, test coverage and clarity over the code 💯

I left minor comments.

src/compute/take/list.rs
elferherrera: I tried that, but I couldn't unwrap the result from the `maybe_usize`. That's why I had to put it in a different allocation.

src/compute/take/list.rs
elferherrera: I couldn't create a vector of references at the same time the slices are created. The complication is that the `array_ref` is made out of refences to the slices, which need to last longer than the references in `array_ref`. This is another reason why the Growable interface could be improved by passing the references when `extending` the array.

src/compute/take/list.rs
elferherrera: Good point. Let me change that

src/array/list/mod.rs
elferherrera: That is a good solution. Let me change that

elferherrera: I tried this approach and it fails to panic when a ListArray is nested

jorgecarleitao: That is unexpected to me: I would expect the inner `ListArray` to perform the same check when it is initialized via `from_data`.

src/compute/take/list.rs
jorgecarleitao:
```
    let usize_indices = indices.values().iter().map(|i| maybe_usize::<O>(*i));

    let arrays = usize_indices
        .map(|index| index.map(|x| &(values.slice(x, 1)) as &dyn Array))
        .collect::<Result<Vec<_>>>()?;
```

src/compute/take/list.rs
jorgecarleitao: I agree, I there are two problems imo. 1. `GrowableList` should expect a `Vec<&ListArray<O>>`, since:

* it takes ownership of the `vec`
* it downcasts arguments to `ListArray`

The first issue affects all interfaces, the latter issue is only for `GrowableList` and some others; `GrowablePrimitive` for example already expects a `PrimitiveArray`.

Great catch!

src/array/list/mod.rs
elferherrera: I thought the same, but I used your proposal to see if it will fail the nested test and it did pass without the expected panic.

src/compute/take/list.rs
elferherrera: I just pushed a change to the take function. With your suggestions I managed to remove some allocations and I managed to use iterators. However, I didn't manage to remove the loop to create the reference of arrays. As you pointed, this is needed for the actual implementation of Growable.

elferherrera: Btw, If you change all the Growable implentations to take `Vec<&dyn Array>` instead of `&[&dyn Array]` the take implementation can be simplified further.

src/array/list/mod.rs
jorgecarleitao: I can't reproduce your behavior; if I change to 

```
        // validate data_type
        let child_data_type = Self::get_child(&data_type);
        assert_eq!(
            child_data_type,
            values.data_type(),
            "The child's datatype must match the inner type of the \'data_type\'"
        );
```

`test_nested_panic` passes after changing the corresponding the expected to `expected = "The child's datatype must match the inner type of the \'data_type\'"`

src/compute/take/list.rs
jorgecarleitao: Let's keep this PR to the `take` and handle this afterwards 👍

src/array/list/mod.rs
elferherrera: mmm, that is really odd. Let me see what Im doing wrong. I did just what you did and the nested test didn't fail

elferherrera:
I screw up the rebase and ended up repeating the commit history. Sorry. I guess you will squash them together if you merge this PR, right?

PR #79, opened 2021-05-12
Pinned nightly in tests.

jorgecarleitao:
See https://github.com/rust-lang/packed_simd/issues/320


houqp:
@jorgecarleitao I think the nightly pin is not needed anymore in the latest simd crate release ` v0.3.5`.

jorgecarleitao:
Thanks @houqp , I tried it in #147 , but I am hitting an internal compiler error when compiling with miri.

PR #81, opened 2021-05-13
Removed need of unsafe call.

jorgecarleitao:
This removes the need for using `unsafe` to extend bitmaps from a trusted len iter.


jorgecarleitao:
Errors are unrelated to this PR.

PR #86, opened 2021-05-14
Fixed error about valid time32 and time64.

jorgecarleitao:
`Time32` and `Time64` should not accept any `TimeUnit`. This fixes it.


elferherrera (approved):
Good simplification with the duration case, there is no need for a generic case when using the TimeUnit that way.

PR #88, opened 2021-05-14
Correct add duration for arithmetics

elferherrera:
Complements the correction done to subtract duration


jorgecarleitao:
Thanks, @elferherrera 👍

PR #91, opened 2021-05-15
Discoverability functions for kernels

### LINKED ISSUE #57: Add discoverability of supported types to each operation
`cast` has `can_cast`, IMO we should add `can_*` to every kernel, so that consumers can "poke" the operation, thereby allowing physical coercion rules to be applied before accessing the data.

elferherrera:
Added discoverability functions for the next kernels:

- basic arithmetic
- comparison
- hash
- length
- sort
- substring
- take
- temporal

Closes #57

### DIFF
--- a/src/compute/arithmetics/mod.rs
+++ b/src/compute/arithmetics/mod.rs
@@ -93,9 +93,9 @@ macro_rules! primitive_arithmetic_match {
     }};
 }
 
-/// Function to execute an arithmetic operation with two arrays
-/// It uses the enum Operator to select the type of operation that is going to
-/// be performed with the two arrays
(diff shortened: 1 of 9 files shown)


jorgecarleitao (commented):
Looks great, thanks a lot, @elferherrera !

What do you think about adding a consistency test like [this](https://github.com/jorgecarleitao/arrow2/blob/main/src/compute/cast/mod.rs#L893), just to make sure we report both sides correctly.

elferherrera:
@jorgecarleitao yeap... that test looks good. Let me add it

jorgecarleitao:
Amazing. Thanks a lot, @elferherrera !

PR #99, opened 2021-05-16
Improved arithmetics kernel

jorgecarleitao:
Thereby allowing more logical type variations to be used (mostly dates, times and durations)

### DIFF
--- a/src/compute/arithmetics/mod.rs
+++ b/src/compute/arithmetics/mod.rs
@@ -63,7 +63,7 @@ use std::ops::{Add, Div, Mul, Neg, Sub};
 
 use num::Zero;
 
-use crate::datatypes::DataType;
+use crate::datatypes::{DataType, TimeUnit};
 use crate::error::{ArrowError, Result};
 use crate::types::NativeType;
 use crate::{array::*, bitmap::Bitmap};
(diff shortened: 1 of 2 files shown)


src/compute/arithmetics/mod.rs
```diff
@@ -75,19 +75,13 @@ use super::arity::unary;
...
-            .unwrap();
-
-        arithmetic_primitive(res_lhs, $op, res_rhs)
+macro_rules! primitive {
+    ($lhs: expr, $rhs: expr, $op: expr, $array_type: ty) => {{
+        let res_lhs = $lhs.as_any().downcast_ref().unwrap();
+
+        let res_rhs = $rhs.as_any().downcast_ref().unwrap();
+
+        arithmetic_primitive::<$array_type>(res_lhs, $op, res_rhs)
```
elferherrera, line 84: You are always finding these little simplifications with dynamic allocators. Nice 😄

src/compute/arithmetics/mod.rs
```diff
@@ -97,43 +91,66 @@ macro_rules! primitive_arithmetic_match {
...
+                Add => decimal::add::add(lhs, rhs),
+                Subtract => decimal::sub::sub(lhs, rhs),
+                Multiply => decimal::mul::mul(lhs, rhs),
+                Divide => decimal::div::div(lhs, rhs),
             };
 
-            res.map(Box::new).map(|x| x as Box<dyn Array>)
+            res.map(|x| Box::new(x) as Box<dyn Array>)
+        }
+        (Time32(_), Add, Duration(_)) | (Date32, Add, Duration(_)) => {
```
elferherrera, line 122: I think this is only valid for `TimeUnit::Millisecond` and `TimeUnit::Second`. The same goes for the subtraction

src/compute/arithmetics/mod.rs
```diff
@@ -97,43 +91,66 @@ macro_rules! primitive_arithmetic_match {
...
+            let lhs = lhs.as_any().downcast_ref().unwrap();
+            let rhs = rhs.as_any().downcast_ref().unwrap();
+            time::add_duration::<i32>(lhs, rhs).map(|x| Box::new(x) as Box<dyn Array>)
+        }
+        (Time32(_), Subtract, Duration(_)) | (Date32, Subtract, Duration(_)) => {
+            let lhs = lhs.as_any().downcast_ref().unwrap();
+            let rhs = rhs.as_any().downcast_ref().unwrap();
+            time::subtract_duration::<i32>(lhs, rhs).map(|x| Box::new(x) as Box<dyn Array>)
+        }
+        (Time64(_), Add, Duration(_))
```
elferherrera, line 132: I think this is only valid for `TimeUnit::Microsecond` and `TimeUnit::Nanosecond`. The same goes for the subtraction

elferherrera (commented):
It is great that you added mixed types to the `can_arithmetic` and simplified the function. I most confess that I prefer to have the full path of the enum (Datatype). It makes it easier to read when you are not familiar with the code. But I guess that is just a matter of preferences. 

I left some comments for the Time options

src/compute/arithmetics/mod.rs
jorgecarleitao: great points. Addressed :)

PR #103, opened 2021-05-18
Made CSV read parallelizable.

jorgecarleitao:
This PR is a small change to the CSV's API to allow multi-threading processing of a CSV. This is achieved by dividing the IO-intensive task from CPU-intensive task, thereby allowing a read head to live on a thread that is not performing the CPU-intensive task of de-serializing the rows.

Together with using `ByteRecords`, this makes the CSV reading fully capable of distributing work across threads.

This follows the same pattern as the parquet reader.

This PR also moves the example in the guide to an example, so that people can run it.

### DIFF
--- a/examples/csv_read.rs
+++ b/examples/csv_read.rs
@@ -1,0 +1,37 @@
+use arrow2::error::Result;
+use arrow2::io::csv::read;
+use arrow2::record_batch::RecordBatch;
+
+fn read_path(path: &str, projection: Option<&[usize]>) -> Result<RecordBatch> {
+    // Create a CSV reader. This is typically created on the thread that reads the file and
+    // thus owns the read head.
+    let mut reader = read::ReaderBuilder::new().from_path(path)?;
+
+    // Infers the schema using the default inferer. The inferer is just a function that maps a string
+    // to a `DataType`.
+    let schema = read::infer_schema(&mut reader, None, true, &read::infer)?;
+
+    // skip 0 (excluding the header) and read up to 100 rows.
+    // this is IO-intensive and performs minimal CPU work. In particular,
+    // no deserialization is performed.
+    let rows = read::read_rows(&mut reader, 0, 100)?;
+
+    // initializes a parser. A parser is a stateless trait to parse rows.
+    // Its logic must be consistent with `read::infer` above.
+    let parser = read::DefaultParser::default();
+
+    // parse the batches into a `RecordBatch`. This is CPU-intensive, has no IO,
+    // and can be performed on a different thread by passing `rows` through a channel.
+    read::parse(&rows, schema.fields(), projection, 0, &parser)
+}
+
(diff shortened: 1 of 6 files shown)


examples/csv_read.rs
```diff
@@ -0,0 +1,36 @@
...
+    let parser = read::DefaultParser::default();
+
+    // initializes a default parser. A parser is stateless; it is just an API to parse rows.
+    // it must match the corresponding "inferer", that is used to infer the schema (if applicable).
+    let schema = read::infer_schema(&mut reader, None, true, &read::infer)?;
+
+    // skip 0 (excluding the header) and read up to 100 entries
+    // this is IO-intensive and performs minimal CPU work. In particular,
+    // no deserialization is performed
+    let rows = read::read_rows(&mut reader, 0, 100)?;
```
elferherrera, line 20: is this function the one that can be run in multiple threads? 
could it be run with async?

examples/csv_read.rs
```diff
@@ -0,0 +1,36 @@
+use arrow2::error::Result;
+use arrow2::io::csv::read;
+use arrow2::record_batch::RecordBatch;
+
+fn read_path(path: &str, projection: Option<&[usize]>) -> Result<RecordBatch> {
```
elferherrera, line 5: Silly question, What is projection?

guide/src/io/csv_reader.md
```diff
@@ -16,28 +16,18 @@ This crate relies on [the crate `csv`](https://crates.io/crates/csv) to scan and
...
-    let mut reader = read::ReaderBuilder::new().from_path(path)?;
-    let parser = read::DefaultParser::default();
-
-    let schema = read::infer_schema(&mut reader, None, true, &infer)?;
-
-    // 0: start from
-    // 100: up to (max batch size)
-    let batch = read::read_batch(&mut reader, &parser, 0, 100, schema, None)?;
-}
+{{#include ../../../examples/csv_read.rs}}
```
elferherrera, line 19: didn't know we can embed examples in the docs now. Very cool

elferherrera (approved):
Nice example for reading a CSV file. I just left a couple of questions for my own understanding of the API

examples/csv_read.rs
jorgecarleitao: when we want to skip certain columns. The indexes represent the columns that we want. Should be documented ^_^

examples/csv_read.rs
jorgecarleitao: I pushed a new example that uses multi-threading to describe the idea. 👍 

The CSV crate does not support async which means that we also do not :(

guide/src/io/csv_reader.md
jorgecarleitao: Yeah, makes it very easy because people can run the example via cargo, but also understand its context on the guide.

PR #110, opened 2021-05-21
Removed miri failure.

jorgecarleitao:


jorgecarleitao:
ccing @roee88 , which actually found the solution to this in https://github.com/apache/arrow-rs/pull/323

PR #141, opened 2021-06-09
Add comparison for boolean array

jorgecarleitao:


elferherrera (commented):
Would it be a good idea to add a `any` function that returns true is at least one value is true. Or `all` if all are true?

jorgecarleitao:
> Would it be a good idea to add a `any` function that returns true is at least one value is true. Or `all` if all are true?

I agree. I do think that that is an aggregation (aka horizontal) operation, right? It is something like (`f(BooleanArray) -> Option<bool>`)?

elferherrera:
> I agree. I do think that that is an aggregation (aka horizontal) operation, right? It is something like (f(BooleanArray) -> Option<bool>)?

That's precisely what Im thinking. Im just not sure it needs to be an Option<bool>. In what case would it be None?

jorgecarleitao:
Yeap, it does not need option. ^_^

PR #142, opened 2021-06-11
Intro page for arrow description

elferherrera:
Adding intro page from the arrow guide that I wrote long time ago. Im working on adapting the remaining text from the old guide since it is based on the first arrow and the buffer is very different now different.


jorgecarleitao:
Thanks a lot, @elferherrera ; great intro! 👍

PR #148, opened 2021-06-13
add unsafe from_data_unchecked

ritchie46:
This proposes an unsafe `from_data_unchecked` for `ListArray<O>` and `Utf8Array<O>`.

The rationale for this change is that I cast the inner type of the list data and create a new `ListArray`, where the offsets are unchanged (only cloned).

```rust
let child = list.values();
let offsets = list.offsets();
let child = cast::cast(child, &child_type)?;
let list = ListArray::from_data(child_type, offsets.clone(), Arc::new(*child), list.validity.clone());
```

There was a name collision on the `Utf8Array` so maybe another function name is better.

### DIFF
--- a/src/array/list/mod.rs
+++ b/src/array/list/mod.rs
@@ -45,7 +45,21 @@ impl<O: Offset> ListArray<O> {
         validity: Option<Bitmap>,
     ) -> Self {
         check_offsets(&offsets, values.len());
+        unsafe { Self::from_data_unchecked(data_type, offsets, values, validity) }
+    }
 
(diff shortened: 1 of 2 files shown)


jorgecarleitao:
Did we observe a performance difference when e.g. casting?

ritchie46:
I should have read the code. I thought that yesterday I read somewhere that `check_offsets` checks if all offsets are monotonously increasing. So that would be `O(n)`. But it only does a few sensible asserts. 

I am closing this one, as that would be needless optimization.

PR #153, opened 2021-06-16
Generalized "take" for unsigned.

jorgecarleitao:


jorgecarleitao:
cc @ritchie46

elferherrera (commented):
I'm lost on this one. What advantages does it have over the previous take?

jorgecarleitao:
Sorry, @elferherrera , was not very clear: it allows taking indexes as `u32`, `i32`, `i64` and `u64`, like the arrow crate does.

elferherrera:
@jorgecarleitao I see. That makes it more versatile. Nicely done.

PR #155, opened 2021-06-18
Improved Primitive API for easiness of use

jorgecarleitao:
The primitive/builder API was difficult to use for nested types. This PR refactors it to make it easier.

The main limitations were:

1. `Builder` could not be used as a trait object
2. `.push(None)` for nested types was not easy (it required a type annotation of an iterator type `T`)
3. it was difficult to control the capacity of the values in nested types

This PR addresses all of the above.

It also renames all `XPrimitive` to `XBuilder` to be consistent with the arrow crate and to disambiguate them from the `Primitive` in `PrimitiveArray`.


jorgecarleitao:
Closing in favor of #156

PR #156, opened 2021-06-20
Generalized builder API into MutableArray API

jorgecarleitao:
This PR removes the Builder trait and all `*Primitive`, replacing them by a more generic and user-friendly `MutableArray` API.

It also greatly simplifies how to create `PrimitiveArrays`, by moving the `fn to(DataType)` from `Primitive` to `PrimitiveArray`.

The `MutableArray` is the mutable counter-part of `Array`. Contrarily to `Array`, `MutableArray`'s buffers are not under an `Arc` and can be modified in place. The tradeoff is that they are not clonable nor slicable, as that requires ref counts. Also, as opposed to arrays, they are strongly typed, which means that they do not currently support dynamic nesting, only static-typed nesting.

All types have their own `MutableArray` so far, but their APIs are experimental and further work is needed: so far the only valid operation is to build them out of a (potentially nested) iterator, and `set` for Primitive and boolean arrays. `Utf8` arrays are not suitable to be mutated, as it incurs a a full re-allocation. Thus, this functionality is mostly relevant for numerics.

Example of the API being used to build a dictionary:

```rust
let data = vec![Some("a"), Some("b"), Some("a")];

let mut a = MutableDictionaryArray::<i32, MutableUtf8Array<i32>>::new();
a.try_extend(data)?;

assert_eq!(a.len(), 3);
assert_eq!(a.values().len(), 2);

let array: DictionaryArray = a.into();

assert_eq!(array.len(), 3);
assert_eq!(array.values().len(), 2);
```

`try_extend` may fail, when the keys of the dict do not fit.


jorgecarleitao:
@Dandandan , you mentioned some time ago in DataFusion that we could benefit from having a mutable array when performing multiple operations (e.g. `a + b * 2 - 1`).

I just benchmarked running `arr1 + arr2` (`PrimitiveArray<f32>`) vs `arr1 += arr2` (`MutablePrimitiveArray<f32>` in this PR, replaces `PrimitiveArrayBuilder` in arrow):

```
add 2^10 f32            time:   [447.36 ns 450.47 ns 454.06 ns]
add mut 2^10 f32        time:   [149.78 ns 151.13 ns 152.76 ns]

add 2^14 f32            time:   [5.4052 us 5.4263 us 5.4472 us]
add mut 2^14 f32        time:   [4.2864 us 4.3044 us 4.3221 us]

add 2^20 f32            time:   [558.88 us 559.33 us 559.80 us]
add mut 2^20 f32        time:   [377.47 us 378.44 us 379.74 us]
```

i.e. about 3x for small arrays, 33% gains for large (it is a non-trivial combination of allocation cost vs operation cost).

ritchie46:
> The MutableArray is the mutable counter-part of Array. Contrarily to Array, MutableArray's buffers are not under an Arc and can be modified in place. The tradeoff is that they are not clonable nor slicable, as that requires ref counts. Also, as opposed to arrays, they are strongly typed, which means that they do not currently support dynamic nesting, only static-typed nesting.

This is great. So you can already utilize the "builders" (doing arithmetic, etc.) before you freeze into a `dyn  Array`?

jorgecarleitao:
@ritchie46 , almost there: this PR does not yet expose methods to mutate the inner of the `MutablePrimitiveArray`, but it is an easy exercise:

```rust
impl<T: NativeType> MutablePrimitiveArray<T> {
    /// Mutable values
    pub fn values_mut_slice(&mut self) -> &mut [T] {
        self.values.as_mut_slice()
    }
}
```

and 

```rust
fn bench_mut_op(lhs: &mut MutablePrimitiveArray<f32>, rhs: &PrimitiveArray<f32>) {
    let len = lhs.len();
    lhs.values_mut_slice()
        .iter()
        .zip(rhs.values().iter())
        .for_each(|(l, r)| *l += *r);
}
```

Dandandan:
> @Dandandan , you mentioned some time ago in DataFusion that we could benefit from having a mutable array when performing multiple operations (e.g. `a + b * 2 - 1`).
> 
> I just benchmarked running `arr1 + arr2` (`PrimitiveArray<f32>`) vs `arr1 += arr2` (`MutablePrimitiveArray<f32>` in this PR, replaces `PrimitiveArrayBuilder` in arrow):
> 
> ```
> add 2^10 f32            time:   [447.36 ns 450.47 ns 454.06 ns]
> add mut 2^10 f32        time:   [149.78 ns 151.13 ns 152.76 ns]
> 
> add 2^14 f32            time:   [5.4052 us 5.4263 us 5.4472 us]
> add mut 2^14 f32        time:   [4.2864 us 4.3044 us 4.3221 us]
> 
> add 2^20 f32            time:   [558.88 us 559.33 us 559.80 us]
> add mut 2^20 f32        time:   [377.47 us 378.44 us 379.74 us]
> ```
> 
> i.e. about 3x for small arrays, 33% gains for large (it is a non-trivial combination of allocation cost vs operation cost).

Yes, looks really cool!

There are also some other places usefulness mutable arrays come to mind, mainly around aggregation and window functions, they might have an even bigger impact there.

Really cool @jorgecarleitao

PR #157, opened 2021-06-21
Removed ToFfi requirement of Array trait.

jorgecarleitao:
The trait can live without it.


jorgecarleitao:
cc @ritchie46 , since Polars uses the trait for the Chunks or series. With this change, it will not need to implement to `ToFfi`.

PR #168, opened 2021-06-24
Added support for IPC 2.0

### LINKED ISSUE #163: Does arrow2 support codecs when reading IPC files?
Does arrow2 support codecs when reading IPC files (Feather v2)?

Feather v2 files created by pandas will use lz4 compression by default. arrow-rs crashes on this kind of files (for now).
Does arrow2 support reading IPC files with a compression codes (lz4, zstd).

https://github.com/apache/arrow-rs/issues/286

jorgecarleitao:
Closes #163

This adds support for IPC format 2.0, which includes compression. The implementation is not the fastest, because it currently requires an extra allocation per buffer, but the roundtrip over the golden files passes, so we can live with it for now. ^_^

Note that this adds a new feature `io_ipc_compression`, to activate the corresponding dependencies. These overlap with `io_parquet`, but it is still good to be aware of.

### DIFF
--- a/src/io/ipc/read/deserialize.rs
+++ b/src/io/ipc/read/deserialize.rs
@@ -22,14 +22,15 @@
 
 use std::{collections::VecDeque, convert::TryInto};
 use std::{
-    io::{Read, Result, Seek, SeekFrom},
+    io::{Read, Seek, SeekFrom},
     sync::Arc,
 };
 
 use crate::buffer::Buffer;
 use crate::datatypes::{DataType, IntervalUnit};
 use crate::endianess::is_native_little_endian;
-use crate::error::Result as ArrowResult;
+use crate::error::{ArrowError, Result};
+use crate::io::ipc::gen::Message::{BodyCompression, CompressionType};
 use crate::{
     array::*,
     bitmap::Bitmap,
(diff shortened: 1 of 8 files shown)


src/io/ipc/read/deserialize.rs
```diff
@@ -80,51 +105,162 @@ fn read_buffer<T: NativeType, R: Read + Seek>(
...
+    match compression.codec() {
+        CompressionType::LZ4_FRAME => {
+            // fast case where we can just copy the contents as is
+            unsafe {
+                // transmute T to bytes.
+                let out_slice = std::slice::from_raw_parts_mut(
+                    buffer.as_mut_ptr() as *mut u8,
+                    length * std::mem::size_of::<T>(),
+                );
+                compression::decompress_lz4(&slice[8..], out_slice)?
```
nevi-me, line 144: If a writer tries to compress a buffer, and then it finds that the compressed buffer ratio isn't good enough, it can still write uncompressed buffers. In that case, we have to check if the first 8 bytes = `-1i64`. It's some of the details that I struggled with

jorgecarleitao: I am following the c++ implementation [here](https://github.com/apache/arrow/blob/master/cpp/src/arrow/ipc/reader.cc#L399), but I can't find that case there :/

nevi-me: I also can't find it now, but I'll look in more detail; as I recall a review comment being raised in the Java implementation for this detal.

jorgecarleitao: Wouldn't it be fine for now to use the same as the c++ implementation does, and improve that behavior over a future PR? None of this results in UB; the worse case is a panic.

nevi-me: I agree, it definitely would :)

PR #187, opened 2021-07-04
expose utils/implementations to public API

ritchie46:
This PR proposes exposing some utils/implementations to the public API. 

`Utf8ValuesIter` was already public, but could not be imported, `alloc` is public in arrow-rs and can be useful in edge cases, and 'trusted_len_unzip` is a great utility to have in a public

### DIFF
--- a/src/array/binary/from.rs
+++ b/src/array/binary/from.rs
@@ -47,9 +47,7 @@ impl<O: Offset, P: AsRef<[u8]>> FromIterator<Option<P>> for BinaryArray<O> {
 /// # Safety
 /// The caller must ensure that `iterator` is `TrustedLen`.
 #[inline]
-pub(crate) unsafe fn trusted_len_unzip<O, I, P>(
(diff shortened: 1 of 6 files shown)


jorgecarleitao:
Hey!

Thanks for this.

`Utf8ValuesIter` is exposed via `Utf8Array::values_iter()`. The equivalent for `BinaryArray` is `BinaryArray::values_iter()`. Would that this sufficient? Others I agree.

ritchie46:
> `Utf8ValuesIter` is exposed via `Utf8Array::values_iter()`. The equivalent for `BinaryArray` is `BinaryArray::values_iter()`. Would that this sufficient? Others I agree.

I wanted to implement a custom trait for the iterators used in Polars, that includes the `Utf8ValuesIter`. This needs the struct to be importable to be able to do so. The methods can be private IMO.

jorgecarleitao (approved):

PR #195, opened 2021-07-06
Fixed (horizontal) equality 

jorgecarleitao:
This PR fixes different issues that exist in equating sliced arrays by removing the previous code and instead relying on arrays' iterators to check for equality.

This is more expensive than the previous code, but given the importance of equality, it is safer to use an item by item approach that results in less code complexity and overall more faithful results.

There are optimizations based on the existence of the validity bitmap that can be applied in the future, but given the importance of a correct equality operator, this PR favors correctness over performance.

Note that array equality is mostly used for testing; query engines usually perform the vertical (element-wise) equality, which this PR does not touch.


jorgecarleitao:
fyi, @nevi-me, since this has been a source of pain.

PR #198, opened 2021-07-07
Read parquet lists plain

jorgecarleitao:
This PR adds support to read one-level `ListArray`, non-dict encoded from parquet:

* [x] `List<X>` nullable, required, and required items
* [x] `List<Primitive>`
* [x] `List<Binary>`
* [x] `List<Boolean>`

Both v1 and v2, compressed and not compressed.

In particular, the following are tested against pyarrow:

```
[
    pa.field("list_int64", pa.list_(pa.int64())),
    pa.field("list_int64_required", pa.list_(pa.field("item", pa.int64(), False))),
    pa.field(
        "list_int64_required_required",
        pa.list_(pa.field("item", pa.int64(), False)),
        False,
    ),
    pa.field("list_int16", pa.list_(pa.int16())), # a non-native parquet type
    pa.field("list_bool", pa.list_(pa.bool_())),
    pa.field("list_utf8", pa.list_(pa.utf8())),
]
```


jorgecarleitao:
I am merging this. A relevant note: the design used here makes this reading almost as fast as reading a non-nested type. The major difference is that the reading of the validity cannot be memcopied directly because the bitpacked of more than one def level does not have the same representation as arrow's validity (bitmaps with the first slot being represented by the LSB).

PR #209, opened 2021-07-20
Update README.md

alamb:
I found some of the statements in this README misleading and I would like to propose some small corrections:

MIRI checks are running on the official crate (e.g): https://github.com/apache/arrow-rs/runs/3116901146

The official crate also has several feature flags, documented at  https://github.com/apache/arrow-rs/tree/master/arrow#features , to control dependencies

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ venv/bin/python parquet_integration/write_parquet.py
 * Uses Rust's compiler whenever possible to prove that memory reads are sound
 * Reading parquet is 10-20x faster (single core) and deserialization is parallelizable
 * Writing parquet is 3-10x faster (single core) and serialization is parallelizable
-\* parquet IO has no `unsafe`
+* parquet IO has no `unsafe`
 * IPC supports big endian
 * More predictable JSON reader
 * `MutableArray` API to work with arrays in-place.
--- a/README.md
+++ b/README.md
@@ -51,13 +51,11 @@ venv/bin/python parquet_integration/write_parquet.py
 * Uses Rust's compiler whenever possible to prove that memory reads are sound
 * Reading parquet is 10-20x faster (single core) and deserialization is parallelizable
 * Writing parquet is 3-10x faster (single core) and serialization is parallelizable
-* MIRI checks on non-IO components (MIRI and file systems are a bit funny atm)
-* parquet IO has no `unsafe`
+\* parquet IO has no `unsafe`
 * IPC supports big endian
 * More predictable JSON reader
 * `MutableArray` API to work with arrays in-place.
 * Generalized parsing of CSV based on logical data types
-* conditional compilation based on cargo `features` to reduce dependencies and size
 * faster IPC reader (different design that avoids an extra copy of all data)
 * IPC supports 2.0 (compression)
 


alamb (commented):
@jorgecarleitao  I also wonder if we should list `arrow-flight` support as something the official crate has that this one does not?

Perhaps also `parquet_derive`?

README.md
```diff
@@ -51,13 +51,11 @@ venv/bin/python parquet_integration/write_parquet.py
 * Uses Rust's compiler whenever possible to prove that memory reads are sound
 * Reading parquet is 10-20x faster (single core) and deserialization is parallelizable
 * Writing parquet is 3-10x faster (single core) and serialization is parallelizable
-* MIRI checks on non-IO components (MIRI and file systems are a bit funny atm)
```
jorgecarleitao, line 54: Aren't we ignoring miri checks in the official crate over some of the tests because they are not passing? E.g. https://github.com/apache/arrow-rs/blob/master/arrow/src/compute/kernels/cast.rs#L3525 and https://github.com/apache/arrow-rs/blob/master/arrow/src/array/raw_pointer.rs#L60 ?

README.md
```diff
@@ -51,13 +51,11 @@ venv/bin/python parquet_integration/write_parquet.py
...
-* parquet IO has no `unsafe`
+\* parquet IO has no `unsafe`
```
jorgecarleitao, line 54: typo?

jorgecarleitao (commented):
I agree with features; I forgot to update here after the new releases. 👍 

parquet-derive: good point.

Arrow-flight is also here (the migration was trivial, though).

README.md
alamb: It is enabled: https://github.com/apache/arrow-rs/pull/421 thanks to the great work of @roee88  and @jhorstmann

README.md
alamb:
```suggestion
* parquet IO has no `unsafe`
```

yes, indeed -- sorry about that

alamb:
re arrow-flight -- I was confused -- I didn't see it on crates.io, but I see it in the repo 👍  https://docs.rs/arrow2/0.1.0/arrow2/?search=flight and https://crates.io/search?q=arrow-flight

README.md
jorgecarleitao: My point is that if a crate has MIRI checks on the CI but ignores certain tests because they fail the check, then that does not really qualify as passing the check, right? It is a bit like commenting tests to make the CI green.

Maybe re-word to

> all tests pass MIRI checks

instead?

alamb: The only checks that are ignored are as follows. Most because they consume too much memory or take too long when they are run under MIRI. Full list below

Implying that the arrow-rs crate doesn't run MIRI seems inaccurate to me. Perhaps you can word this section of the readme a bit more positively and focus on on the complete MIRI coverage of all the tests so far or something?

```
/Users/alamb/Software/arrow-rs/arrow/src/array/raw_pointer.rs:60:    #[cfg_attr(miri, ignore)] // sometimes does not panic as expected
/Users/alamb/Software/arrow-rs/arrow/src/util/integration_util.rs:725:    #[cfg_attr(miri, ignore)] // running forever
/Users/alamb/Software/arrow-rs/arrow/src/compute/kernels/cast.rs:3525:    #[cfg_attr(miri, ignore)] // running forever
/Users/alamb/Software/arrow-rs/arrow/src/compute/kernels/cast_utils.rs:223:    #[cfg_attr(miri, ignore)] // unsupported operation: can't call foreign function: mktime
/Users/alamb/Software/arrow-rs/arrow/src/compute/kernels/length.rs:157:    #[cfg_attr(miri, ignore)] // running forever
/Users/alamb/Software/arrow-rs/arrow/src/compute/kernels/length.rs:174:    #[cfg_attr(miri, ignore)] // running forever
/Users/alamb/Software/arrow-rs/arrow/src/compute/kernels/length.rs:284:    #[cfg_attr(miri, ignore)] // error: this test uses too much memory to run on CI
/Users/alamb/Software/arrow-rs/arrow/src/compute/kernels/length.rs:301:    #[cfg_attr(miri, ignore)] // error: this test uses too much memory to run on CI
 ```

alamb: Actually, now that I double checked it turns out that the results of MIRI were still being ignored 🤦  -- we'll fix that: https://github.com/apache/arrow-rs/pull/578

jorgecarleitao: fwiw, those are not ignored because they take a long time to run; they take a long time to run because they constitute UB. The same tests pass on arrow2.

alamb: can't wait to get all this arrow2 goodness into arrow-rs and share it with the world 👍

PR #212, opened 2021-07-22
Fixed nested FFI.

jorgecarleitao:
This PR fixes a bug in reading FFI children of child (nested types).

### DIFF
--- a/src/ffi/ffi.rs
+++ b/src/ffi/ffi.rs
@@ -643,10 +643,6 @@ impl<'a> ArrowArrayRef for ArrowArrayChild<'a> {
     fn schema(&self) -> &Ffi_ArrowSchema {
         self.schema
     }
-
-    fn child(&self, _: usize) -> Result<ArrowArrayChild> {
-        todo!()
-    }
 }
 
 impl<'a> ArrowArrayChild<'a> {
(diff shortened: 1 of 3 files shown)


src/ffi/ffi.rs
```diff
@@ -644,8 +644,8 @@ impl<'a> ArrowArrayRef for ArrowArrayChild<'a> {
         self.schema
     }
 
-    fn child(&self, _: usize) -> Result<ArrowArrayChild> {
-        todo!()
+    fn child(&self, index: usize) -> Result<ArrowArrayChild> {
+        create_child(self.array, self.schema, self.parent.clone(), index)
```
roee88, line 648: Since the implementation is the same I think that you can keep just the one in ArrowArrayRef (this is how it's done in arrow-rs)

jorgecarleitao: Sorry, I am not sure I am following: what do you mean with "you can keep just the one in ArrowArrayRef"?

roee88: There is a default implementation in the trait in https://github.com/jorgecarleitao/arrow2/blob/a5035bd016ad89121bc528d547bfbc9e74880799/src/ffi/ffi.rs#L568

In `impl ArrowArrayRef for Arc<ArrowArray>` you rely on it and don't reimplement child. Can't you eo the same here?

jorgecarleitao: you are right, even better. Thanks!

PR #218, opened 2021-07-23
Added support for limited sort

### LINKED ISSUE #215: Introduce limit option to sort kernel like arrow
We are considering switch to arrow2 in datafuse, yet there are some features to implement like sort limit.

jorgecarleitao:
Closes #215

This:

* adds support to sort with a limit.
* adds bench to sort `utf8` with nulls
* improves performance of (unlimited) sort of `utf8` for small arrays by 10-15%
* makes all sorts unstable (most were already)

This change is backward incompatible: all public functions in `compute::sort` now expect an `Option<usize>` denoting the limit of the sort. The migration is to add a new parameter `None` to them.

Bench results (utf8 with `10%` of nulls)

```
sort utf8 null 2^10     time:   [130.62 us 130.75 us 130.86 us]                                
                        change: [-14.887% -14.549% -14.269%] (p = 0.00 < 0.05)
sort utf8 null 2^12     time:   [651.78 us 652.41 us 653.03 us]                                
                        change: [-12.832% -12.593% -12.372%] (p = 0.00 < 0.05)
                        Performance has improved.
sort utf8 null 2^14     time:   [3.0201 ms 3.0244 ms 3.0286 ms]                                 
                        change: [-11.194% -11.032% -10.871%] (p = 0.00 < 0.05)
```

### DIFF
--- a/src/compute/sort/lex_sort.rs
+++ b/src/compute/sort/lex_sort.rs
@@ -49,7 +49,7 @@ pub struct SortColumn<'a> {
 ///             nulls_first: false,
 ///         }),
 ///     },
-/// ]).unwrap();
+/// ], None).unwrap();
 ///
 /// let sorted = sorted_columns[0].as_any().downcast_ref::<Int64Array>().unwrap();
 /// assert_eq!(sorted.value(1), -64);
--- a/src/compute/sort/lex_sort.rs
+++ b/src/compute/sort/lex_sort.rs
@@ -1,20 +1,3 @@
-// Licensed to the Apache Software Foundation (ASF) under one
-// or more contributor license agreements.  See the NOTICE file
-// distributed with this work for additional information
-// regarding copyright ownership.  The ASF licenses this file
-// to you under the Apache License, Version 2.0 (the
-// "License"); you may not use this file except in compliance
-// with the License.  You may obtain a copy of the License at
-//
-//   http://www.apache.org/licenses/LICENSE-2.0
-//
-// Unless required by applicable law or agreed to in writing,
-// software distributed under the License is distributed on an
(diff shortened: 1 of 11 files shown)


src/compute/sort/lex_sort.rs
```diff
@@ -135,9 +118,12 @@ pub(crate) fn build_compare(array: &dyn Array, sort_option: SortOptions) -> Resu
     })
 }
 
-/// Sort elements lexicographically from a list of `ArrayRef` into an unsigned integer
-/// [`Int32Array`] of indices.
-pub fn lexsort_to_indices(columns: &[SortColumn]) -> Result<PrimitiveArray<i32>> {
+/// Sorts a list of [`SortColumn`] into a non-nullable [`PrimitiveArray<i32>`]
+/// representing the indices that would sort the columns.
```
sundy-li, line 122: Why not return u32 instead of i32, [0,65536) is very small range for indices in bigdata area.

sundy-li: The official arrow used UInt32Array.

jorgecarleitao: I agree. This was due to a limitation that I meanwhile lifted. Follow up PR I will make this sort generic over index, which means that I plan to support `u32, i32, u64, i64` at the users' discretion.

jorgecarleitao: Done here: https://github.com/jorgecarleitao/arrow2/pull/220 Thanks for the input :)

sundy-li:
Great job, and I think it will be better to add limit option to merge sort.

https://github.com/datafuselabs/datafuse/blob/master/common/datavalues/src/columns/common.rs#L225-L329

PR #222, opened 2021-07-25
Add support to merge sort with a limit

### LINKED ISSUE #221: Merge sort support limit option
After #218, I think it will be better to add limit option in merge_sort. Best wishes.

https://github.com/datafuselabs/datafuse/blob/master/common/datavalues/src/columns/common.rs#L225-L329

sundy-li:
Closes #221 

The functions `compute::merge_sort::take_arrays` and `compute::merge_sort::merge_sort` now expects a third argument, `Option<usize>`, denoting an optional limit. To migrate, add `None` to their calls.

### DIFF
--- a/src/compute/merge_sort/mod.rs
+++ b/src/compute/merge_sort/mod.rs
@@ -38,7 +38,7 @@
 //! let slices2 = merge_sort_slices(a2, a3);
 //! let slices = merge_sort_slices(slices1, slices2);
 //!
-//! let array = take_arrays(&[a0, a1, a2, a3], slices);
+//! let array = take_arrays(&[a0, a1, a2, a3], slices, None);
 //! ```
 //!
 //! A common operation in query engines is to merge multiple fields based on the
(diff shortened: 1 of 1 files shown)


src/compute/merge_sort/mod.rs
```diff
@@ -89,13 +89,22 @@ type MergeSlice = (usize, usize, usize);
...
     let slices = slices.into_iter();
     let len = arrays.iter().map(|array| array.len()).sum();
-    let mut growable = make_growable(arrays, false, len);
+    let limit = limit.unwrap_or(len);
+    let mut growable = make_growable(arrays, false, limit);
 
+    let mut current_len = 0;
     for (index, start, len) in slices {
-        growable.extend(index, start, len)
+        if len + current_len >= limit {
```
Dandandan, line 101: I think this probably makes the code without a limit (limit=None) slower.

Dandandan: This can be avoided by having two different implementations/bodies.

sundy-li: done

jorgecarleitao: I would prefer to not expose a new function just for this; wouldn't it make sense to move this `if` to outside of the loop and keep the function with the `limit` argument?

sundy-li: I'm afraid the *if* can't move outside of the loop because the slice is an iterator.

Dandandan: If you do `slices.into_iter()` in two code blocks instead of reusing it, it shouldn't be a compile error?

sundy-li: What I mean is that`slices` is lazy iterated, we don't know the item of `len` unless we call next(), so the `if` check must be inside the loop when limit is some value.

jorgecarleitao: I was thinking in something like this:

```
...
if let Some(limit) = limit {
    iterate with the limit condition
} else {
   iterate without the limit condition
}
...
growable.to_box()
```

sundy-li: Sounds great.

PR #227, opened 2021-07-25
Fix generic parameter in lexsort usage in benches

sundy-li:


ritchie46:
This is a fix for the benches that don't run on main.

sundy-li:
> This is a fix for the benches that don't run on main.

Thanks for the advice, but better to fix this if we need run some bench tests.

Dandandan (approved):

Dandandan:
Went ahead and merged this. Thanks @sundy-li

PR #229, opened 2021-07-26
take list: do not collect array of pointers

ritchie46:
This PR proposes a new constructor method on `GrowableList`. This constructor is generic over `&[AsRef[ListArray<O>>` and can be constructed with an array over `ListArray`. 

This removes a redundant vector allocation in the take kernel, where a `Vec<ListArray>` was converted to a `Vec<&dyn Array>` so that it could be passed to the constructor.

### DIFF
--- a/src/array/growable/list.rs
+++ b/src/array/growable/list.rs
@@ -67,6 +67,57 @@ pub struct GrowableList<'a, O: Offset> {
 }
 
 impl<'a, O: Offset> GrowableList<'a, O> {
+    fn finish_new(
+        arrays: Vec<&'a ListArray<O>>,
+        inner: Vec<&'a dyn Array>,
+        extend_null_bits: Vec<ExtendNullBits<'a>>,
+        capacity: usize,
+        use_validity: bool,
+    ) -> Self {
+        let values = make_growable(&inner, use_validity, 0);
+
+        let mut offsets = MutableBuffer::with_capacity(capacity + 1);
+        let length = O::default();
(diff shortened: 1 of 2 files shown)


src/array/growable/list.rs
```diff
@@ -67,6 +67,57 @@ pub struct GrowableList<'a, O: Offset> {
...
+        let inner = arrays
+            .iter()
+            .map(|array| array.values().as_ref())
+            .collect::<Vec<_>>();
+        Self::finish_new(arrays, inner, extend_null_bits, capacity, use_validity)
+    }
+
     /// # Panics
     /// This function panics if any of the `arrays` is not downcastable to `PrimitiveArray<T>`.
     pub fn new(arrays: &[&'a dyn Array], mut use_validity: bool, capacity: usize) -> Self {
```
jorgecarleitao, line 123: I think that we should make this expect a `ListArray` instead of `&dyn Array`. No need for two functions just because of a downcast, and this also makes it more explicit that there is an expectation.

fwiw, this signature is historical; e.g. GrowablePrimitive uses PrimitiveArray already.

jorgecarleitao:
I will go ahead and change all signatures in growable to make this consistent across and avoid the extra allocation altogether.

jorgecarleitao:
Proposal in #238

jorgecarleitao:
I am closing this one since it was generalized by #238. Thanks a lot for the proposal, @ritchie46 !

PR #235, opened 2021-07-29
Special-case filter single array record batch in filter_record_batch

### LINKED ISSUE #234: Avoid materialization of indices in filter_record_batch for single arrays
Some optimization possible in the kernel seems to avoid materializing `(start, end)` and pre-building the filter  when there is only a single input array. This probably will be a time/memory usage win.

This is a common situation in DataFusion / SQL (`select a from x where [...]`).

Dandandan:
Gives a speed up for single-column batches.

Closes #234

```
filter single record batch                                                                            
                        time:   [229.93 us 231.52 us 233.29 us]
                        change: [-54.913% -54.501% -54.106%] (p = 0.00 < 0.05)
                        Performance has improved.
  4 (4.00%) high mild
  3 (3.00%) high severe
```

### DIFF
--- a/src/compute/filter.rs
+++ b/src/compute/filter.rs
@@ -256,12 +256,21 @@ pub fn filter_record_batch(
     record_batch: &RecordBatch,
     filter: &BooleanArray,
 ) -> Result<RecordBatch> {
-    let filter = build_filter(filter)?;
-    let filtered_arrays = record_batch
-        .columns()
-        .iter()
-        .map(|a| filter(a.as_ref()).into())
-        .collect();
+    let num_colums = record_batch.columns().len();
+
+    let filtered_arrays = match num_colums {
+        1 => {
+            vec![super::filter::filter(record_batch.columns()[0].as_ref(), filter)?.into()]
+        }
+        _ => {
+            let filter = build_filter(filter)?;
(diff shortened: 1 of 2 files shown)


src/compute/filter.rs
```diff
@@ -256,12 +256,21 @@ pub fn filter_record_batch(
...
-    let filtered_arrays = record_batch
-        .columns()
-        .iter()
-        .map(|a| filter(a.as_ref()).into())
-        .collect();
+    let num_colums = record_batch.columns().len();
+
+    let filtered_arrays = match num_colums {
+        1 => {
+            vec![super::filter::filter(record_batch.columns()[0].as_ref(), filter)?.into()]
```
jorgecarleitao, line 263:
```suggestion
            vec![filter(record_batch.columns()[0].as_ref(), filter)?.into()]
```

jorgecarleitao (approved):
Looks great 👍

src/compute/filter.rs
Dandandan: Not possible, but I'll rename the parameter instead (also micro-bench incoming).

Dandandan:
Nice! About a 2x improvement for this case

PR #238, opened 2021-07-30
Simplified `new` signature of growable API

jorgecarleitao:
It also removes an extra allocation that was being done internally and an extra check.

# Backward incompatible changes:

`Growable*::new` now expects a vector (not slice) of references of their corresponding types and no longer panic (the type is by definition correct). To migrate, downcast the vector of arrays prior to passing them to the growable.

```rust
let array = Utf8Array::<i32>::from_iter(vec![Some("a"), Some("bc"), None, Some("defh")]);
let mut a = GrowableUtf8::new(vec![&array], false, 0);
```

I.e. signature change is `new(&[&'a dyn Array], ...) -> new(Vec<&'a A>, ...)` where `A` is the corresponding array type.

### DIFF
--- a/examples/growable.rs
+++ b/examples/growable.rs
@@ -1,0 +1,38 @@
+use arrow2::array::growable::{Growable, GrowablePrimitive};
+use arrow2::array::{Array, PrimitiveArray};
+
+fn main() {
+    // say we have two sorted arrays
+    let array0 = PrimitiveArray::<i64>::from_slice(&[1, 2, 5]);
+    let array1 = PrimitiveArray::<i64>::from_slice(&[3, 4, 6]);
+
+    // and we found a way to compute the slices that sort them:
+    // (array_index, start of the slice, length of the slice)
+    let slices = &[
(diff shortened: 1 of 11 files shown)


ritchie46:
I was thinking about a generic constructor to merge the two into one (in #229), but I do think this better. The caller can be responsible for creating the needed data structure.

As there is already a lifetime in `Growable..`. How do you feel about the following?

```rust
pub struct GrowableBinary<'a, 'b, O: Offset> {
    arrays: &'b[&'a BinaryArray<O>>],
    ...
}
```

Or using the same lifetime `'a` for the slice/ `Cow` (in case of ownership problems).

jorgecarleitao:
I agree that we can further simplify this. Would it be ok to leave it to a next iteration? I think I need to think about this a bit more: it seems to me that we do not even need the `Vec` in all cases for as long as we internally allocate references to the supporting buffers and validities. It needs a small internal refactor, but it might work.

PR #239, opened 2021-07-30
Added support to parquet IO on wasm32

jorgecarleitao:
This PR adds a new cargo feature, `io_parquet_compression`, that activates the compressions in parquet.

The corollary is that this allows compiling this crate against wasm32 with parquet io. AFAIK this effectively makes arrow2 to be usable to interact with parquet directly on a browser.

As an example, this seems to allow building a client-side web application that reads or writes (uncompressed) parquet files.

For compression support, [some work is needed on the `zstd`](https://github.com/gyscos/zstd-rs/issues/48).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -271,7 +271,7 @@ jobs:
           export CARGO_HOME="/github/home/.cargo"
           export CARGO_TARGET_DIR="/github/home/target"
           # no need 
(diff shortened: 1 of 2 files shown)


jorgecarleitao:
cc @domoritz since if I remember, you were interested in the wasm aspect.

domoritz:
Yes, this change is awesome.

I have to rewrite my arrow-wasm repo against arrow2 but ran into some issue where some types were not exported. I'll try again sometime and consolidate the things I find in an issue.

Issue #231, opened 2021-07-26
Support loading Feather v2 (IPC) files with more than 1 million tables


ghuls:
As can be seen in https://github.com/pola-rs/polars/issues/1023, loading of Feather v2 (IPC) files with more than 1 million tables does not work.

(py)arrow had the same bug: https://issues.apache.org/jira/projects/ARROW/issues/ARROW-10056

It boils down to the flatbuffer verification code, which has max_tables=1_000_000 by default.
Increasing this limit solves the problem.

In (py)arrow the max table value is determined per dataset based on the footer size, to prevent specially crafted IPC files to take an extraordinary amount of time to verify a very small input IPC file:

https://github.com/apache/arrow/pull/9447

```python
In [33]: %time df2 = pl.read_ipc(''test.v2.feather'', use_pyarrow=False)
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<timed exec> in <module>

/software/polars/py-polars/polars/io.py in read_ipc(file, use_pyarrow, storage_options)
    415             tbl = pa.feather.read_table(data)
    416             return pl.DataFrame.from_arrow(tbl)
--> 417         return pl.DataFrame.read_ipc(data)
    418 
    419 

/software/polars/py-polars/polars/eager/frame.py in read_ipc(file)
    606         """
    607         self = DataFrame.__new__(DataFrame)
--> 608         self._df = PyDataFrame.read_ipc(file)
    609         return self
    610 

RuntimeError: Any(ArrowError(Ipc("Unable to get root as footer: TooManyTables")))
```

I think if your you replace gen::File::root_as_footer with gen::File::root_as_footer_with_opts, you can set the max_table option: At https://github.com/jorgecarleitao/arrow2/blob/16c089e0c4c2b079da10b6fb8fbec19cdabaeab3/src/io/ipc/read/reader.rs#L98

https://docs.rs/flatbuffers/2.0.0/src/flatbuffers/get_root.rs.html#39-49

PR #240, opened 2021-07-30
Added support for reading IPC with more than 1 million columns

ghuls:
Set Flatbuffer verification parameters to the same settings as the
C++ arrow implementation (ARROW-11559). This change allows reading
IPC data with more than 1 milion columns.

Closes: #231


jorgecarleitao (approved):
Awesome; thanks!

PR #243, opened 2021-08-01
Make merge_sort_slices MergeSortSlices public

sundy-li:
In OLAP database system, sort/merge_sort is not just for one column.

The unit of sort compute is `block`, we should make `merge_sort_slices` public, then we can take the indices from the block to build new blocks.


sundy-li:
We also need `materialize_merge_indices` to reuse the iterator to sort other arrays.

Issue #245, opened 2021-08-02
Poor performance in `sort::sort_to_indices`  with limit option in arrow2


sundy-li:
Hello, I did a benchmark vs arrow1.

It shows that arrow2 has great performance improved in `min`, `sum` and `max` etc.

But the `sort::sort_to_indices` function with limit 100  has a performance drop.

Benchmark results:
```
arrow2-sort 2^13 f32    time:   [67.831 us 67.978 us 68.144 us]   

arrow1-sort 2^13 f32    time:   [34.306 us 34.405 us 34.521 us] 
```

Codes: https://github.com/sundy-li/learn/tree/master/arrow-vs-arrow2/benches

Scripts:
`cargo bench -- "arrow1-sort 2\^13 f32"    `
`cargo bench -- "arrow2-sort 2\^13 f32"    `

sundy-li:
![image](https://user-images.githubusercontent.com/3325189/127879968-d4e0afad-a843-4528-a0d4-2db054666e17.png)

`as_slice` is the bottleneck.


I think we can use `values: &[T]` to replace `get: G`.

Improved in commit:  https://github.com/datafuse-extras/arrow2/pull/1

Now:

```
arrow2-sort-limit 2^13 f32                                                                             
  time:   [46.601 us 46.665 us 46.728 us]
  change: [-24.847% -24.656% -24.483%] (p = 0.00 < 0.05)
  Performance has improved.
```

Another bottleneck is `from_usize().unwrap()` in the loop.

jorgecarleitao:
Great finding; thanks for sharing!

Would you like me to backport it here?

sundy-li:
> Great finding; thanks for sharing!
> 
> Would you like me to backport it here?

Sure, after I finish the migration, I'll create pr to arrow2.

PR #246, opened 2021-08-03
Sort improve

sundy-li:
Related #245 

Changes:

- Add `to_vec` function to `MergeSortSlices`, this makes iterator reuseable.
- Add `buffer_from_range ` to Index trait
- Fix some bad codes in benches.
- Change `cmp(a,b).reverse()` to `cmp(b,a)`

### DIFF
--- a/src/buffer/mutable.rs
+++ b/src/buffer/mutable.rs
@@ -93,6 +93,19 @@ impl<T: NativeType> MutableBuffer<T> {
         }
     }
 
+    /// Allocates a new [MutableBuffer] with `len` and capacity to be at least `len` where
+    /// all bytes are not initialized
+    #[inline]
+    pub fn from_len(len: usize) -> Self {
+        let new_capacity = capacity_multiple_of_64::<T>(len);
+        let ptr = alloc::allocate_aligned(new_capacity);
+        Self {
+            ptr,
+            len,
+            capacity: new_capacity,
+        }
+    }
+
     /// Ensures that this buffer has at least `self.len + additional` bytes. This re-allocates iff
     /// `self.len + additional > capacity`.
     /// # Example
--- a/src/buffer/mutable.rs
+++ b/src/buffer/mutable.rs
@@ -96,14 +96,10 @@ impl<T: NativeType> MutableBuffer<T> {
     /// Allocates a new [MutableBuffer] with `len` and capacity to be at least `len` where
     /// all bytes are not initialized
     #[inline]
-    pub fn from_len(len: usize) -> Self {
-        let new_capacity = capacity_multiple_of_64::<T>(len);
-        let ptr = alloc::allocate_aligned(new_capacity);
-        Self {
-            ptr,
-            len,
-            capacity: new_capacity,
-        }
+    pub unsafe fn from_len(len: usize) -> Self {
+        let mut buffer = MutableBuffer::with_capacity(len);
+        buffer.set_len(len);
+        buffer
     }
 
     /// Ensures that this buffer has at least `self.len + additional` bytes. This re-allocates iff
--- a/src/compute/merge_sort/mod.rs
+++ b/src/compute/merge_sort/mod.rs
@@ -291,6 +291,29 @@ where
             None => self.right = None,
         }
     }
+
+    /// Collect the MergeSortSlices to be a vec for reusing
+    #[warn(dead_code)]
+    pub fn to_vec(self, limit: Option<usize>) -> Vec<MergeSlice> {
+        match limit {
+            Some(limit) => {
+                let mut v = Vec::with_capacity(limit);
+                let mut current_len = 0;
+                for (index, start, len) in self {
+                    v.push((index, start, len));
+
+                    if len + current_len >= limit {
+                        break;
+                    } else {
+                        current_len += len;
+                    }
+                }
+
+                v
+            }
+            None => self.into_iter().collect(),
+        }
+    }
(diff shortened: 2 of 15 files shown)


sundy-li:
Seems `#![feature(step_trait)]` is not acceptable.

src/compute/sort/utf8.rs
```diff
@@ -9,9 +9,13 @@ pub(super) fn indices_sorted_unstable_by<I: Index, O: Offset>(
     options: &SortOptions,
     limit: Option<usize>,
 ) -> PrimitiveArray<I> {
-    let get = |idx| unsafe { array.value_unchecked(idx as usize) };
+    let values = unsafe {
+        (0..array.len())
+            .map(|idx| array.value_unchecked(idx as usize))
+            .collect::<Vec<_>>()
```
jorgecarleitao, line 15: won't this cause a large allocation?

src/buffer/mutable.rs
```diff
@@ -93,6 +93,19 @@ impl<T: NativeType> MutableBuffer<T> {
         }
     }
 
+    /// Allocates a new [MutableBuffer] with `len` and capacity to be at least `len` where
+    /// all bytes are not initialized
+    #[inline]
+    pub fn from_len(len: usize) -> Self {
```
jorgecarleitao, line 99: This is `unsafe` as it allows consumers to read un-initialized memory. The way to achieve this behavior is to use `with_capacity()` followed by `set_len()` (followed by an operation that is guaranteed to not read any entry and write to all entries up to len).

src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![feature(step_trait)]
```
jorgecarleitao, line 1: this is nightly only, right?

src/compute/sort/primitive/indices.rs
```diff
@@ -18,16 +18,14 @@ where
     T: NativeType,
     F: Fn(&T, &T) -> std::cmp::Ordering,
 {
-    unsafe {
-        common::indices_sorted_unstable_by(
-            array.validity(),
-            |x: usize| *array.values().as_slice().get_unchecked(x),
```
jorgecarleitao, line 24: if the `as_slice` is the slow path, wouldn't it make sense to move it to outside the function here instead?

it seems to me that the current design requires allocating 2x the memory for dictionaries, strings, etc.

jorgecarleitao (changes requested):
Thanks! Left some comments around.

src/compute/sort/utf8.rs
sundy-li: Yes, the quick-sort time complex os `O(n * log(n) )`,  means it'll take at least `O(n * log(n) )` times get from the buffer. I think it's worth to initialized it in a large vector.

I need to do some benches to verify Utf8 sort  compare with master.

src/compute/sort/primitive/indices.rs
sundy-li: Got it.

sundy-li:
The only way currently I can find to reduce hot path `from_usize().unwrap()`  is to add a method `buffer_from_range`  in trait `Index`.

src/compute/merge_sort/mod.rs
```diff
@@ -291,6 +291,29 @@ where
...
+    #[warn(dead_code)]
+    pub fn to_vec(self, limit: Option<usize>) -> Vec<MergeSlice> {
+        match limit {
+            Some(limit) => {
+                let mut v = Vec::with_capacity(limit);
+                let mut current_len = 0;
+                for (index, start, len) in self {
+                    v.push((index, start, len));
+
+                    if len + current_len >= limit {
```
jorgecarleitao, line 305: I think that this misses the last entries: if the last slice is past `limit`, imo it should be modified to be equal to `limit - current_len`, so that we do not lose the last items within the slice

src/compute/sort/common.rs
```diff
@@ -134,12 +123,12 @@ where
                 // Soundness:
                 // all indices in `indices` are by construction `< array.len() == values.len()`
                 // limit is by construction < indices.len()
-                let limit = limit - validity.null_count();
+                let limit = limit.saturating_sub(validity.null_count());
```
jorgecarleitao, line 126: This is a bug fix I believe. could we add a test for it?

sundy-li: We already checked `limit  > validity.null_count()`, so it may not be a bug, but better to add this line.

jorgecarleitao:
I can see that there is some limitations with stable Rust wrt to the ranges and indices. I will take a stab at addressing this here, to see if we can get the remaining performance out of this without the `from_usize` around.

jorgecarleitao:
I think that https://github.com/jorgecarleitao/arrow2/pull/247 addresses the limitations; could you take a look and see what do you think?

jorgecarleitao:
I have also created https://users.rust-lang.org/t/idiomatic-way-to-populate-vector-from-branched-loop/63152 to see if we get good ideas.

sundy-li:
Update:

Add   `shrink_to_fit()` to fix #249

sundy-li:
> I think that #247 addresses the limitations; could you take a look and see what do you think?

Ok, I'll take look at that.

jorgecarleitao:
I will go ahead and merge this, and clear up the API in a follow-up PR. Thanks a lot, @sundy-li !

PR #251, opened 2021-08-05
Change to comfy-table

PsiACE:
Change to comfy-table from prettytable. In addition, comfy-table is still actively maintained.

Signed-off-by: Chojan Shang <psiace@outlook.com>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ serde_json = { version = "^1.0", features = ["preserve_order"], optional = true 
 indexmap = { version = "^1.6", optional = true }
 
 # used to print columns in a nice columnar format
-prettytable-rs = { version = "^0.8", optional = true }
+comfy-table = { version = "4.0", optional = true }
 
 flatbuffers = { version = "=2.0.0", optional = true }
 hex = { version = "^0.4", optional = true }
@@ -96,7 +96,7 @@ io_parquet_compression = [
     "parquet2/brotli",
 ]
 io_json_integration = ["io_json", "hex"]
-io_print = ["prettytable-rs"]
+io_print = ["comfy-table"]
 # the compute kernels. Disabling this significantly reduces compile time.
 compute = []
 # base64 + io_ipc because arrow schemas are stored as base64-encoded ipc format.
--- a/src/io/print.rs
+++ b/src/io/print.rs
@@ -17,8 +17,7 @@
 
 use crate::{array::*, error::Result, record_batch::RecordBatch};
 
-use prettytable::format;
-use prettytable::{Cell, Row, Table};
+use comfy_table::{Cell, Table};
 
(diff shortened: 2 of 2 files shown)


jorgecarleitao (approved):
Awesome! Thanks a lot, @PsiACE !

jorgecarleitao:
fwiw, I did go through the code for `unsafe`; https://github.com/Nukesor/comfy-table/issues/42 (not to its dependencies yet)

Nukesor:
`strum` and `strum_macros` don't have unsafe in their code-base.
`crossterm` has a few `unsafe` blocks, which are necessary for terminal control.

`comfy_table` actually does use one of those functions to check, if the current program is on a TTY or not.
It should be this block in crossterm:
```
src/tty.rs
33:        unsafe { libc::isatty(fd) == 1 }
```

Right now, this function is called if the table width isn't explicitely specified.
It isn't in your case, so this is relevant.

In the current code, it get's **always** called if there is no explicit width set via `Table::set_table_width()`.

I think this could be refactored to only be called if either:
- Percental Constrains exist. as we need to know the current width for percental calculations.
- The user uses `Dynamic` mode, as this mode takes all content and formats it in a way that it fits into the given space.

Dynamic content arrangement in tty's is **the** core feature in comfy-table, which makes this libc call a necessary evil.
As this isn't a thing in Prettytable, it seems to be safe in this regard.

Nukesor:
Turns out, one more call is used. Crossterm's `size` function, which is used to determine the terminal size.
This function is only called in conjunction with `is_tty` and if a tty has been detected.
```
    if wrap_with_result(unsafe { ioctl(fd, TIOCGWINSZ.into(), &mut size) }).is_ok() {
```
This seems to be another libc call with, which his used to communicate with `/dev/tty` via a raw filedescriptor.
http://rosettacode.org/wiki/Terminal_control/Dimensions#Library:_BSD_libc

All unsafe functions are called from comfy-table's `Table::get_table_width`:
```
    pub fn get_table_width(&self) -> Option<u16> {
        if let Some(width) = self.table_width {
            Some(width)
        } else if self.is_tty() {
            if let Ok((table_width, _)) = size() {
                Some(table_width)
            } else {
                None
            }
        } else {
            None
        }
    }
   ```

Nukesor:
Turns out, I already had this issue once and you can explicitly disable the whole `tty` logic.
Just call `Table::force_no_tty` and you're good to go.

I'm also open to a PR, which hides this behind a feature flag. (With tty should still stay the default though, as this is still comfy-tables core feature)

jorgecarleitao:
I agree that feature flagging this upstream would be useful to this use-case since we do not care about terminals (that is e.g. [nutshell's business](https://github.com/nushell/nushell), cc @elferherrera :)

Regardless, being maintained is imo more relevant than these :)

PR #254, opened 2021-08-05
Added wrapping_cast to cast kernels

### LINKED ISSUE #252: Improve the performance in cast kernel using AsPrimitive trait in generic dispatch
Beat arrow2 cast performance less than 10 LOC, about 7 times faster in casting 2^20 size u32array to u64array.

Benchmark codes:

https://github.com/sundy-li/learn/blob/master/arrow-vs-arrow2/benches/cast_kernels.rs

```
cast u32 to u64 1048576 time:   [4.3537 ms 4.3615 ms 4.3702 ms]                                     
                        change: [-0.0612% +0.1772% +0.4521%] (p = 0.17 > 0.05)
                        No change in performance detected.


cast u32 to u64 v2 1048576                                                                            
                        time:   [630.99 us 645.11 us 664.46 us]
                        change: [-11.727% -4.8932% +1.8462%] (p = 0.20 > 0.05)
                        No change in performance detected.
```

sundy-li:
fixes #252

### DIFF
--- a/src/compute/cast/mod.rs
+++ b/src/compute/cast/mod.rs
@@ -40,7 +40,8 @@ pub use utf8_to::*;
 #[derive(Clone, Copy, Debug)]
 pub struct CastOptions {
     /// default to false
-    /// if set to true, the cast will use AsPrimitive to convert the primitive value, this could be more efficient
+    /// whether an overflowing cast should be converted to `None` (default), or be wrapped (i.e. `256i16 as u8 = 0` vectorized).
+    /// Settings this to `true` is 5-6x faster for numeric types.
     pub ignore_overflow: bool,
 }
 
--- a/src/compute/cast/mod.rs
+++ b/src/compute/cast/mod.rs
@@ -333,7 +333,7 @@ pub fn cast(array: &dyn Array, to_type: &DataType) -> Result<Box<dyn Array>> {
     cast_with_options(array, to_type, CastOptions { wrapped: false })
 }
 
-/// Similar to cast, but overflowing cast will be wrapped
+/// Similar to [`cast`], but overflowing cast is wrapped
 /// Behavior:
 /// * PrimitiveArray to PrimitiveArray: overflowing cast will be wrapped (i.e. `256i16 as u8 = 0` vectorized).
 pub fn wrapped_cast(array: &dyn Array, to_type: &DataType) -> Result<Box<dyn Array>> {
--- a/src/compute/cast/mod.rs
+++ b/src/compute/cast/mod.rs
@@ -741,6 +741,41 @@ mod tests {
     }
 
     #[test]
+    fn test_cast_i32_as_f64() {
+        let array = Int32Array::from_slice(&[5, 6, 7, 8, 9]);
+        let b = cast(
+            &array,
+            &DataType::Float64,
+            CastOptions {
+                ignore_overflow: true,
+            },
+        )
+        .unwrap();
+        let c = b.as_any().downcast_ref::<Float64Array>().unwrap();
+        assert!((5.0 - c.value(0)).abs() < f64::EPSILON);
+        assert!((6.0 - c.value(1)).abs() < f64::EPSILON);
+        assert!((7.0 - c.value(2)).abs() < f64::EPSILON);
+        assert!((8.0 - c.value(3)).abs() < f64::EPSILON);
+        assert!((9.0 - c.value(4)).abs() < f64::EPSILON);
+    }
+
+    #[test]
+    fn test_cast_u16_as_u8() {
+        let array = UInt16Array::from_slice(&[255, 256, 257, 258, 259]);
+        let b = cast(
+            &array,
+            &DataType::UInt8,
+            CastOptions {
+                ignore_overflow: true,
+            },
+        )
+        .unwrap();
+        let c = b.as_any().downcast_ref::<UInt8Array>().unwrap();
+        let values = c.values().as_slice();
+        assert_eq!(values, &[255, 0, 1, 2, 3])
+    }
+
+    #[test]
     fn test_cast_i32_to_u8() {
         let array = Int32Array::from_slice(&[-5, 6, -7, 8, 100000000]);
         let b = cast(&array, &DataType::UInt8, CastOptions::default()).unwrap();
--- a/src/compute/cast/mod.rs
+++ b/src/compute/cast/mod.rs
@@ -573,8 +573,8 @@ pub fn cast(array: &dyn Array, to_type: &DataType, options: CastOptions) -> Resu
         (UInt32, Int16) => primitive_to_primitive_dyn::<u32, i16>(array, to_type, options),
         (UInt32, Int32) => primitive_to_primitive_dyn::<u32, i32>(array, to_type, options),
         (UInt32, Int64) => primitive_to_primitive_dyn::<u32, i64>(array, to_type, options),
-        (UInt32, Float32) => primitive_to_primitive_dyn::<u32, f32>(array, to_type, options),
-        (UInt32, Float64) => primitive_to_primitive_dyn::<u32, f64>(array, to_type, options),
+        (UInt32, Float32) => primitive_to_primitive_dyn::<u32, f32>(array, to_type, as_options),
+        (UInt32, Float64) => primitive_to_primitive_dyn::<u32, f64>(array, to_type, as_options),
 
         (UInt64, UInt8) => primitive_to_primitive_dyn::<u64, u8>(array, to_type, options),
         (UInt64, UInt16) => primitive_to_primitive_dyn::<u64, u16>(array, to_type, options),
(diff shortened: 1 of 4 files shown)


sundy-li:
It's backwards compatible, but it follows the ways as the modern number cast did.

src/compute/cast/mod.rs
```diff
@@ -36,6 +36,20 @@ pub use primitive_to::*;
 pub use timestamps::*;
 pub use utf8_to::*;
 
+/// options that define how Cast kernels should behave
+#[derive(Clone, Copy, Debug)]
+pub struct CastOptions {
+    pub ignore_overflow: bool,
```
jorgecarleitao, line 42: this needs a documentation `/// ...`.

src/compute/cast/mod.rs
```diff
@@ -36,6 +36,20 @@ pub use primitive_to::*;
 pub use timestamps::*;
 pub use utf8_to::*;
 
+/// options that define how Cast kernels should behave
```
jorgecarleitao, line 39:
```suggestion
/// options defining how Cast kernels behave
```

src/compute/cast/primitive_to.rs
```diff
@@ -72,6 +79,22 @@ where
     PrimitiveArray::<O>::from_trusted_len_iter(iter).to(to_type.clone())
 }
 
+/// Cast [`PrimitiveArray`] as a [`PrimitiveArray`] of another physical type via numeric conversion.
```
jorgecarleitao, line 82: update description to differentiate from above.

jorgecarleitao (commented):
Makes sense. Thanks a lot. 4 observations:

1. could you point me to somewhere describing "the ways as the modern number cast did."? (I feel I should read on that :))
2. could we add a test for other cases (I think we need 4 csaes: (`ignore_overflow = true` and `not overflowing`, `ignore_overflow = true` and overflowing, etc.) 
3. For all cases where we up-cast (e.g. `i32 -> i64`), shouldn't we ignore the option and always hit the fast path since it is guaranteed to not overflow?
4. I wonder if the cast option could somehow follow an equivalent notation as Rust's own notation. If I understand correctly, `ignore_overflow` is equivalent to using `as`, which is equivalent to a saturating the cast

sundy-li:
> could you point me to somewhere describing "the ways as the modern number cast did."

Emm, this is just c-style cast or default cast behavior in MySQL/ClickHouse.

In c, cpp, rust, we use `as` keyword to cast the numbers to ignore overflow.
In MySQL/ClickHouse, it's safe to cast a wider number to the shorter one.

src/compute/cast/mod.rs
```diff
@@ -715,10 +772,156 @@ mod tests {
...
+            &array,
+            &DataType::UInt8,
+            CastOptions {
+                ignore_overflow: true,
+            },
+        )
+        .unwrap();
+        let c = b.as_any().downcast_ref::<UInt8Array>().unwrap();
+        let values = c.values().as_slice();
+        assert_eq!(values, &[255, 0, 1, 2, 3])
```
jorgecarleitao, line 788: awesome! What are your thoughts about calling this `wrapping`? This seems the same behavior as `255u8.wrapping_add(1)` vs `255u8.checked_add(1)`

src/compute/cast/mod.rs
```diff
@@ -36,6 +36,30 @@ pub use primitive_to::*;
 pub use timestamps::*;
 pub use utf8_to::*;
 
+/// options defining how Cast kernels behave
+#[derive(Clone, Copy, Debug)]
+pub struct CastOptions {
+    /// default to false
+    /// if set to true, the cast will use AsPrimitive to convert the primitive value, this could be more efficient
```
jorgecarleitao, line 43:
```suggestion
    /// whether an overflowing cast should be converted to `None` (default), or be wrapped (i.e. `256i16 as u8 = 0` vectorized).
    /// Settings this to `true` is 5-6x faster for numeric types.
```

src/compute/cast/mod.rs
sundy-li: Hmm, I did not anything about `wrapping_add ` before.

sundy-li:
```
pub const fn wrapping_add(self, rhs: Self) -> Self {
            intrinsics::wrapping_add(self, rhs)
        }
```

Seems the arg rhs is `Self`, `255u8.wrapping_add(1u16)` will not work.

jorgecarleitao: We have the vectorized versions in `compute/arithmetics` (e.g. `arrow2::compute::arithmetics::basic::add::saturating_add`, since some engines use it to avoid panics (e.g. default behavior of `numpy`). 

Rust `overflowing` is usually used for an operation that returns `Option<T>`, where `None` is when the operation overflows.

sundy-li: OK, I got that. But seems lots of duplicate codes should be added.

BTW, would it be possible to introduce a version of `sum`which takes care of the overflow problems? After #56  is done.

```
sum(&array) -> Scalar
```

If the array is U8Array~U64Array, the scalar will be u64;
If the array is I8Array~I64Array, the scalar will be i64;
If the array is floatArray, the scalar will be float64.

In datafuse we have to cast `U8Array` to `U64Array`, then apply the sum with simd.

jorgecarleitao: Yes, we can surely do that. Not sure whether we will hit SIMD so easily because there is a cast now, but it is certainly useful. New issue to discuss design?

src/compute/cast/mod.rs
```diff
@@ -291,31 +324,50 @@ fn cast_large_to_list(array: &ListArray<i64>, to_type: &DataType) -> ListArray<i
...
 /// Unsupported Casts
 /// * To or from `StructArray`
 /// * List to primitive
 /// * Utf8 to boolean
 /// * Interval and duration
 pub fn cast(array: &dyn Array, to_type: &DataType) -> Result<Box<dyn Array>> {
+    cast_with_options(array, to_type, CastOptions { wrapped: false })
+}
+
+/// Similar to cast, but overflowing cast will be wrapped
```
jorgecarleitao, line 336:
```suggestion
/// Similar to [`cast`], but overflowing cast is wrapped
```

src/compute/cast/mod.rs
```diff
@@ -36,6 +36,29 @@ pub use primitive_to::*;
 pub use timestamps::*;
 pub use utf8_to::*;
 
+/// options defining how Cast kernels behave
+#[derive(Clone, Copy, Debug)]
+pub struct CastOptions {
```
jorgecarleitao, line 41: given that we now have the `cast` and `wrapped_cast`, we can keep this private and only keep the others as the public surface?

src/compute/cast/mod.rs
```diff
@@ -291,31 +324,50 @@ fn cast_large_to_list(array: &ListArray<i64>, to_type: &DataType) -> ListArray<i
...
+
+/// Similar to cast, but overflowing cast will be wrapped
+/// Behavior:
+/// * PrimitiveArray to PrimitiveArray: overflowing cast will be wrapped (i.e. `256i16 as u8 = 0` vectorized).
+pub fn wrapped_cast(array: &dyn Array, to_type: &DataType) -> Result<Box<dyn Array>> {
+    cast_with_options(array, to_type, CastOptions { wrapped: true })
+}
+
+#[inline]
+pub fn cast_with_options(
```
jorgecarleitao, line 344: we can keep this private since `wrapped_cast` and `cast` cover all cases.

src/compute/cast/mod.rs
```diff
@@ -291,31 +324,50 @@ fn cast_large_to_list(array: &ListArray<i64>, to_type: &DataType) -> ListArray<i
...
 /// * Utf8 to boolean
 /// * Interval and duration
 pub fn cast(array: &dyn Array, to_type: &DataType) -> Result<Box<dyn Array>> {
+    cast_with_options(array, to_type, CastOptions { wrapped: false })
+}
+
+/// Similar to cast, but overflowing cast will be wrapped
+/// Behavior:
+/// * PrimitiveArray to PrimitiveArray: overflowing cast will be wrapped (i.e. `256i16 as u8 = 0` vectorized).
+pub fn wrapped_cast(array: &dyn Array, to_type: &DataType) -> Result<Box<dyn Array>> {
```
jorgecarleitao, line 339:
```suggestion
pub fn wrapping_cast(array: &dyn Array, to_type: &DataType) -> Result<Box<dyn Array>> {
```

to be aligned with `wrapping_add`,`wrapping_sub`, etc. :)

jorgecarleitao (approved):
This last change was an awesome idea; it keeps it very clear (and even backward incompatible) 👍

Left some final minor comments, but the design is there.

Thanks a lot, @sundy-li , great work here!

src/compute/cast/mod.rs
sundy-li: But `dictionary_to_dictionary_values` is pub, which one do you prefer

1. Create `wrapping_dictionary_to_dictionary_values` and `wrapping_dictionary_to_dictionary_values`
2. Keep the CastOption public

jorgecarleitao: uhm, good point. I would make the two cases for now, just for consistency with Rust's std, but I agree that if we start having a lot of cases we may need to switch.

Is it a big pain for DataFuse to not have the option?

sundy-li: Ok, both are good to me, I keep the CastOption as pub, it will not confuse users if we provide more apis.

If you'd like it to be private,  just add another version of `dictionary_to_dictionary_values ` and `dictionary_to_dictionary_keys`, this could be done in another pr.

sundy-li: > Is it a big pain for DataFuse to not have the option?
Not yet, we just depend on the performance of `cast`. And I don't know if we will have more options in cast kernel.
As `SortOptions ` did.

sundy-li:
> Is it a big pain for DataFuse to not have the option?

Not yet, we just depend on the performance of `cast`.  And I don't know if we will have more options in cast kernel.
As `SortOptions ` did, such like `cast_null_as_default: bool,  overflow_as_default: bool`

jorgecarleitao:
pulled locally and PRed minor changes. Sorry, with 2 I meant the two variations of the functions, not the 2 APIs, to keep it aligned with the other functions :)

PR #256, opened 2021-08-05
Added support to filter parquet pages.

jorgecarleitao:
This is the corresponding of https://github.com/jorgecarleitao/parquet2/pull/34, adding support to skip parquet pages based on its statistics.

# Backwards incompatible changes:

* `RecordReader` now requires an optional argument, `page_filter`. Pass `None` to recover previous behavior
* `get_page_iterator` now requires an optional argument, `page_filter`. Pass `None` to recover previous behavior

The page filter's signature is `Arc<dyn Fn(&ColumnDescriptor, &DataPageHeader) -> bool>`: use the first argument to tell which column is being considered and the second argument for information about the page.

In particular, `DataPageHeader::statistics()` contains the page statistics.

### DIFF
--- a/benches/read_parquet.rs
+++ b/benches/read_parquet.rs
@@ -21,7 +21,7 @@ fn read_decompressed_pages(buffer: &[u8], size: usize, column: usize) -> Result<
     let file = Cursor::new(buffer);
 
     let reader =
-        read::RecordReader::try_new(file, Some(vec![column]), None, Arc::new(|_, _| true))?;
+        read::RecordReader::try_new(file, Some(vec![column]), None, Arc::new(|_, _| true), None)?;
 
     for maybe_batch in reader {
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -58,7 +58,8 @@ futures = { version = "0.3", optional = true }
 # for faster hashing
 ahash = { version = "0.7", optional = true }
 
-parquet2 = { version = "0.2", optional = true, default_features = false, features = ["stream"] }
+#parquet2 = { version = "0.2", optional = true, default_features = false, features = ["stream"] }
+parquet2 = { git = "https://github.com/jorgecarleitao/parquet2", branch = "filter_pages", optional = true, default_features = false, features = ["stream"] }
```
jorgecarleitao, line 62: remember to publish before merging this.

PR #264, opened 2021-08-09
Added support for projection pushdown on IPC files

### LINKED ISSUE #261: Allow projection pushdown to IPC files

jorgecarleitao:
Closes #261 and simplifies internal code (to support this).

# Backwards incompatible changes

* `ipc::read::FileReader::new` now accept a new parameter `projection: Option<Vec<usize>>` that can be used to filter columns. 
* `read::ipc::read_record_batch` now accept a new parameter `projection: Option<(&[usize], Arc<Schema>)>` that can be used to filter columns. 

where `usize` is the index of the column in the IPC schema (known from the file metadata).

Columns not selected are neither read nor deserialized (i.e. are skipped). Pass `None` to recover the original behavior.


src/io/ipc/read/reader.rs
```diff
@@ -228,18 +232,41 @@ pub fn read_batch<R: Read + Seek>(
 }
 
 impl<'a, R: Read + Seek> FileReader<'a, R> {
-    /// Creates a new reader
-    pub fn new(reader: &'a mut R, metadata: FileMetadata) -> Self {
+    /// Creates a new [`FileReader`]. Use `projection` to only take certain columns.
+    pub fn new(
+        reader: &'a mut R,
+        metadata: FileMetadata,
+        projection: Option<HashSet<usize>>,
```
Dandandan, line 239: Doesn't a `&[usize]` type make more sense here?
FWIW, contains doesn't have to be used if performance is a concern as the fields can be looked up by index.

jorgecarleitao: 👍 I agree. I think we need more internal changes to not expose this limitations to the end-user.

Basically IPC writes into a specific sequence based on the schema, and the files offsets are stored based off from that sequence. When we skip a column, we can't simply seek the file; we need to ignore some parts of the header, for everything to fall into place. Maybe the solution is to first create `Vec<Array>` with the IPC order, and then re-order according to the projection, thereby allowing the `&[usize]` to its full extent.

ghuls: @Dandandan Not sure if it is an issue here, but in polars there was a big slow down of selecting a huge number of columns out of a table with 2 million rows. Now it uses a hash to lookup column names when there are more than 300 columns.
https://github.com/pola-rs/polars/pull/1028/commits/1e9622669e5524198ede692c87a890b93e3955f4

Dandandan: @ghuls 

Yeah linear search is something to avoid with e.g. millions of columns. where it easily causes this O(n^2) pattern.

Here the `usize` is an index though already, so the creation of the hashset is doing more than strictly necessary.

jorgecarleitao: in IPC format, we always need to loop over all columns, even if we do not consume them. I.e. we basically need to do

```rust
for (i, column) in ipc_schema.columns() {
     if projection.contains(i) {
           columns.push(read_ipc(...))
     } else {
           skip_ipc(...)
     }
}
```

any suggestion of how we run the above with a `Vec` and not `O(n*m)`? I am probably missing something very basic here.

Dandandan: The fields are already ordered, as the code is now using `enumerate`?

I was thinking of something like this:
```rust
let fields = metadata
                .schema()
                .fields();
            
let fields_projected = projection.iter().map(|i| fields[i].clone())
...
```

jorgecarleitao: for projection the schema, that is possible. Where it is not so easy is in reading the buffers and logical nodes [here](https://github.com/jorgecarleitao/arrow2/pull/264/files#diff-49fe4bc4d69cbab72233ac588f34363626902a238cc0fd8b841408b66d5a63d7R78).

jorgecarleitao: Had to write a small iterator adapter, but the API is now with `Option<Vec<usize>>` for the reader and `&[usize]` for the function.

The code panics if the projection is not ordered, as that behavior is not supported by the IPC format in general (and column re-ordering is more like a query engine's responsibility.

Issue #259, opened 2021-08-07
Improve rem_scalar/div_scalar  performance


sundy-li:
Refer to:

Blog: https://lemire.me/blog/2019/02/08/faster-remainders-when-the-divisor-is-a-constant-beating-compilers-and-libdivide/
Paper:  https://arxiv.org/abs/1902.01961
Go: https://github.com/bmkessler/fastdiv
Rust: https://docs.rs/strength_reduce/0.2.3/strength_reduce/

lideen999:
Fuse processing number% 3 is slow. The main reason is that type conversion is made every time. I try to remove type conversion. The performance can be 2G -- > 5G / s

sundy-li:
@lideen999  

#252 shows the main reason cast.

I did some perfs in datafuse, seems the rem is the hot path.

sundy-li:
@jorgecarleitao   

By using `strength_reduce`, it shows great improvement in datafuse.

jorgecarleitao:
This is a valid request: valid use-case, documented benefits, crate with implementation available. 👍

Would you like to work on it, or would you like me to take it?

Also, I went through the crate and is `unsafe` free, so even easier sell.

ritchie46:
The same optimization can be used for division as well right?

sundy-li:
> The same optimization can be used for division as well right?

Yes! Of course.

> Would you like to work on it, or would you like me to take it?

Sorry, currently I may have no time to work on it, I just provide this idea to make arrow2 work better.

jorgecarleitao:
Done in #275

PR #275, opened 2021-08-11
Specialize division and remainder of unsigned using strength_reduce 

ritchie46:
Already make this PR to start the discussion. I implemented the strength reduction algorithm discussed in #259 for division of `u64` arrays and the performance increases are almost 10x!

I benchmarked a divsion by 4 (which is a single bit shift) and a divison by a prime number.

```
divide_scalar 4         time:   [41.871 us 41.883 us 41.899 us]                             
                        change: [-89.393% -89.388% -89.382%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 11 outliers among 100 measurements (11.00%)
  2 (2.00%) high mild
  9 (9.00%) high severe

divide_scalar prime     time:   [71.806 us 71.821 us 71.845 us]                                
                        change: [-82.082% -82.035% -81.993%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low mild
  2 (2.00%) high mild
  5 (5.00%) high severe
```

@jorgecarleitao, I used some `unsafe` to convince the compiler of the proper types. How do you want to go about this. We could also create some extra traits to do the conversion.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,6 +60,9 @@ ahash = { version = "0.7", optional = true }
 
 parquet2 = { version = "0.3", optional = true, default_features = false, features = ["stream"] }
 
+# for division/remainder optimization at runtime
+strength_reduce = { version = "0.2", optional = true }
+
 [dev-dependencies]
 rand = "0.8"
 criterion = "0.3"
@@ -98,7 +101,7 @@ io_parquet_compression = [
 io_json_integration = ["io_json", "hex"]
 io_print = ["comfy-table"]
 # the compute kernels. Disabling this significantly reduces compile time.
-compute = []
+compute = ["strength_reduce"]
 # base64 + io_ipc because arrow schemas are stored as base64-encoded ipc format.
 io_parquet = ["parquet2", "io_ipc", "base64", "futures"]
 benchmarks = ["rand"]
@@ -166,4 +169,8 @@ harness = false
 
 [[bench]]
 name = "write_ipc"
+harness = false
+
+[[bench]]
+name = "arithmetic_kernels"
 harness = false
--- a/benches/arithmetic_kernels.rs
+++ b/benches/arithmetic_kernels.rs
@@ -1,0 +1,52 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#[macro_use]
+extern crate criterion;
+use criterion::Criterion;
+
+use arrow2::array::*;
+use arrow2::util::bench_util::*;
+use arrow2::{
+    compute::arithmetics::basic::div::div_scalar, datatypes::DataType, types::NativeType,
+};
+use num::NumCast;
+use std::ops::Div;
+
+fn bench_div_scalar<T>(lhs: &PrimitiveArray<T>, rhs: &T)
+where
+    T: NativeType + Div<Output = T> + NumCast,
+{
+    criterion::black_box(div_scalar(lhs, rhs));
+}
+
+fn add_benchmark(c: &mut Criterion) {
+    let size = 65536;
+    let arr = create_primitive_array_with_seed::<u64>(size, DataType::UInt64, 0.0, 43);
+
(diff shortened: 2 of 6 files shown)


src/compute/arithmetics/basic/div.rs
```diff
@@ -109,10 +111,28 @@ where
...
+    T: NativeType + Div<Output = T> + NumCast,
 {
     let rhs = *rhs;
-    unary(lhs, |a| a / rhs, lhs.data_type().clone())
+    match T::DATA_TYPE {
+        DataType::UInt64 => {
+            //  Safety:
+            // We need to go from generic to concrete type and back to generic.
+            // We know the types are the same, but the compiler needs some convincing
+            unsafe {
```
jorgecarleitao, line 122: remove 1 `unsafe`:

```
           let lhs = (lhs as &dyn Array)
                .as_any()
                .downcast_ref::<PrimitiveArray<u64>>()
                .unwrap();
            let rhs = rhs.to_u64().unwrap();
            let reduced_div = StrengthReducedU64::new(rhs);
            let result = unary(lhs, |a| a / reduced_div, lhs.data_type().clone());
            // Safety: we just proved that `lhs` is `PrimitiveArray<u64>` which means that
            // T = u64
            unsafe { std::mem::transmute::<PrimitiveArray<u64>, PrimitiveArray<T>>(result) }
```

Cargo.toml
```diff
@@ -60,6 +60,9 @@ ahash = { version = "0.7", optional = true }
 
 parquet2 = { version = "0.3", optional = true, default_features = false, features = ["stream"] }
 
+# for division/remainder optimization at runtime
+strength_reduce = "0.2"
```
jorgecarleitao, line 64: optional under feature `compute`?

benches/arithmetic_kernels.rs
```diff
@@ -0,0 +1,53 @@
...
+where
+    T: NativeType + Div<Output = T> + NumCast,
+{
+    criterion::black_box(div_scalar(lhs, rhs));
+}
+
+fn add_benchmark(c: &mut Criterion) {
+    let size = 65536;
+    let arr = create_primitive_array_with_seed::<u64>(size, DataType::UInt64, 0.0, 43);
+    let arr_i64 = create_primitive_array_with_seed::<i64>(size, DataType::Int64, 0.0, 43);
```
jorgecarleitao, line 40: not used.

jorgecarleitao (commented):
Overall looks great!

I think that for now we should live with the transmute. I think that the new scalar API will enable us to write this function as `div_scalar(&dyn Array, &dyn Scalar) -> Box<dyn Array>`, which will allow us to avoid this transmute and specialize directly based on `DataType`.

u16 and u32 is also supported by strength division.

Cargo.toml
ritchie46: :+1:

benches/arithmetic_kernels.rs
ritchie46: Ah.. forgot about that one.

src/compute/arithmetics/basic/div.rs
ritchie46: :+1:

ritchie46:
> I think that for now we should live with the transmute. I think that the new scalar API will enable us to write this function as div_scalar(&dyn Array, &dyn Scalar) -> Box<dyn Array>, which will allow us to avoid this transmute and specialize directly based on DataType.

Great! than I will implement the other types as well.

ritchie46:
@jorgecarleitao I updated the PR. It is now implemented for `rem_scalar` and `div_scalar`. 

There is a bit repetition in the code. Because branches work on different concrete types it seems not possible to write a generic function for that. A macro would be possible but, but given the usage of `unsafe`, I chose for repetition as that's more readable.

src/compute/arithmetics/basic/div.rs
Dandandan: Does this mean the "transmute-free" claim must go?

ritchie46: Transmute-lean should be the goal😁

jorgecarleitao: at least when I wrote "transmute-free", I was thinking about `arrow-rs` treating every buffer as `&[u8]` and converting them to `&[T]` without being able to prove (only trust) that the buffers are indeed created from `&[T]` (and valid for some size N).

We do use transmutes in the crate atm; e.g. the IPC has [one transmute](https://github.com/jorgecarleitao/arrow2/blob/main/src/io/ipc/read/read_basic.rs#L69) when reading bytes to `&mut [T]`, because there is no way of doing this in safe Rust (despite being `safe`).

imo this is another of those cases: the way to achieve this in safe rust is via trait specialization, which is only available in the unstable channel.

The relevant aspect imo is that we are in control of the call and know exactly why it is safe.

jorgecarleitao:
Do we have a test hitting the branch?

ritchie46:
> Do we have a test hitting the branch?

Will write that tomorrow.

src/compute/arithmetics/basic/div.rs
jorgecarleitao: But maybe there is a way to achieve the same without transmutes; I would definitely prefer that!

ritchie46:
@jorgecarleitao all branches are rain in tests now. I think its good to go.

PR #276, opened 2021-08-11
Reduce dependencies from confi-table and enable wasm on io_print feature.

jorgecarleitao:
Hat tip to @roee88 for the ping :)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -271,7 +271,7 @@ jobs:
           export CARGO_HOME="/github/home/.cargo"
           export CARGO_TARGET_DIR="/github/home/target"
           # no need 
-          cargo build --no-default-features --features=merge_sort,io_ipc,io_csv,io_json,io_parquet --target wasm32-unknown-unknown
+          cargo build --no-default-features --features=merge_sort,io_ipc,io_csv,io_print,io_json,io_parquet --target wasm32-unknown-unknown
 
   linux-simd-test:
     name: SIMD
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -271,7 +271,7 @@ jobs:
           export CARGO_HOME="/github/home/.cargo"
           export CARGO_TARGET_DIR="/github/home/target"
           # no need 
-          cargo build --no-default-features --features=merge_sort,io_ipc,io_csv,io_json,io_parquet --target wasm32-unknown-unknown
+          cargo build --no-default-features --features=merge_sort,io_ipc,io_csv,io_print,io_json,io_parquet --target wasm32-unknown-unknown
```
roee88, line 274: I doubt wasm32-unknown-unknown will work. Only wasm32-wasi works with stdio.

jorgecarleitao: it worked :/

roee88: It appears like the approach in rust is to throw unsupported operation errors. See https://github.com/rust-lang/rust/tree/master/library/std/src/sys. 

So it compiles but any code path that actually uses it will fail unless the user does some extra low level work (unlike in wasm32-wasi where it's actually implemented out of the box). 

My personal opinion it that it's perfectly fine to just ensure that it compiles and users would need to ensure that their specific usage is using only valid execution paths. It's just that having this specific feature in the CI targeting wasm32-unknown-unknown might mislead users.

jorgecarleitao: Thanks a lot for the investigation!

What is the specific path that would fail? the usage of `println!`?

roee88: So I wrote a little sample and it all works perfectly fine:
- The `arrow2::io::print::write` function works as expected
- The `arrow2::io::print::print` function doesn't print anything in `wasm32-unknown-unknown` but it doesn't fail nor panics.

We're good here. Thanks for the opportunity to learn something new ;)

PR #285, opened 2021-08-13
Removed order implementation for `days_ms` / `Interval(DayTime)`

jorgecarleitao:
The `days_ms` type cannot be ordered as there is a number of ms `m` over which `(1,m) > (2,0)`, and thus ordering by days and then by milliseconds is not a valid order. In general such intervals have no order relation.

Thanks a lot to @emkornfield [for pointing this here](https://github.com/jorgecarleitao/arrow2/pull/268#discussion_r687968328).

All operations relying on order are no longer supported for the logical type `Interval(DayTime)` (ord, comparison, sort).

### DIFF
--- a/src/array/ord.rs
+++ b/src/array/ord.rs
@@ -4,10 +4,7 @@ use std::cmp::Ordering;
 
 use crate::datatypes::*;
 use crate::error::{ArrowError, Result};
-use crate::{
-    array::*,
-    types::{days_ms, NativeType},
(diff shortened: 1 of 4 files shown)


emkornfield:
I think we landed on officially day milliseconds being unorderable.  There was a comment in the java code at some point that MS component was meant to be less than a day (but I don't think it is enforced any place)

PR #305, opened 2021-08-20
Improved performance of comparison with SIMD feature flag (2x-3.5x)

jorgecarleitao:
This PR adds explicit SIMD implementations to comparison compute.

By operating over groups of 8 and applying the relevant conversions with native packed_simd2 apis, we gain 2-3.5x in performance. There is no difference in performance over the non-simd implementation; it was just written in a more abstract fashion to allow the simd version to be plugged in.

```bash
cargo bench --no-default-features --features benchmarks,compute --bench comparison_kernels
cargo bench --no-default-features --features benchmarks,compute,simd --bench comparison_kernels
```

```bash
eq Float32              time:   [20.319 us 20.394 us 20.492 us]                        
                        change: [-70.496% -70.317% -70.165%] (p = 0.00 < 0.05)

eq scalar Float32       time:   [16.452 us 16.512 us 16.585 us]                               
                        change: [-54.292% -53.941% -53.602%] (p = 0.00 < 0.05)

lt Float32              time:   [19.904 us 19.972 us 20.063 us]                        
                        change: [-67.563% -67.329% -67.084%] (p = 0.00 < 0.05)

lt scalar Float32       time:   [16.410 us 16.467 us 16.541 us]                               
                        change: [-54.324% -54.088% -53.814%] (p = 0.00 < 0.05)
```

### DIFF
--- a/benches/comparison_kernels.rs
+++ b/benches/comparison_kernels.rs
@@ -32,7 +32,7 @@ where
 
 fn bench_op_scalar<T>(arr_a: &PrimitiveArray<T>, value_b: T, op: Operator)
 where
-    T: NativeType + std::cmp::PartialOrd,
+    T: NativeType + Simd8,
 {
     primitive_compare_scalar(
         criterion::black_box(arr_a),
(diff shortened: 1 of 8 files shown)


jorgecarleitao:
There is an extra optimization in using the more natural md representation of the native types, but it requires a bit more work on the `MutableBitmap` to be safely constructed from `u8-u64` chunks

PR #321, opened 2021-08-23
Added API to FFI `Field`

### LINKED ISSUE #328: Align FFI API with arrow's C++ API
Generally, the FFI should support the possibility of an external consumer to allocate the bridge structs (`ArrowArray` and `ArrowSchema`).

This requires that the bridge struct is not allocated by us; we should instead expect an allocated and aligned pointer that we can write the struct to it. Such change is backward-incompatible.

jorgecarleitao:
So that the fields can be imported and exported from the C data interface.

The public API is different since ownership rules changed (#328); see `examples/ffi.rs` with a complete example.

Closes #328.

### DIFF
--- a/.github/workflows/integration-ffi.yml
+++ b/.github/workflows/integration-ffi.yml
@@ -38,6 +38,6 @@ jobs:
           python -m venv venv
           source venv/bin/activate
 
-          pip install maturin==0.10.2 toml==0.10.1 pyarrow==4.0.0
+          pip install maturin==0.10.2 toml==0.10.1 pyarrow==5.0.0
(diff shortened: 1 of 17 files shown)


jorgecarleitao:
fwi @ritchie46 , since this adds some backward incompatible changes to FFI.

OTOH, [we can now share `Field` in the FFI](https://github.com/jorgecarleitao/arrow2/pull/321/files#diff-08e80e2f86379de3b24890655798da5dd08c79cda57b5116413e84e819ddeb16R172)

ritchie46:
Right, will take a look!

PR #325, opened 2021-08-24
add a test case of incorrect merge sort

sundy-li:
Related 

https://github.com/jorgecarleitao/arrow2/commit/fa3b003759d5a98200a98ed93facb8751f3aa6a0

Found in https://github.com/datafuselabs/datafuse/pull/1575

### DIFF
--- a/src/compute/merge_sort/mod.rs
+++ b/src/compute/merge_sort/mod.rs
@@ -532,7 +532,7 @@ pub fn build_comparator<'a>(
 
 #[cfg(test)]
 mod tests {
-    use crate::array::{Int32Array, Utf8Array};
+    use crate::array::{BooleanArray, Int32Array, UInt64Array, Utf8Array};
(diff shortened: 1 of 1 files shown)


jorgecarleitao:
found the root cause, fixing it

jorgecarleitao:
This is fixed by #327

jorgecarleitao:
Thanks a lot for taking the time to write a minimal test case; very useful! (also glad that you are using the merge-sort compute kernel; was not sure if it would ever be used ^_^)

sundy-li:
> Thanks a lot for taking the time to write a minimal test case; very useful! (also glad that you are using the merge-sort compute kernel; was not sure if it would ever be used ^_^)

Thanks for the fix.  The merge-sort kernel is useful, and it also can be optimized by heap-sort when the limit option is very small (holding a priority queue in memory to reduce the compare times). I've created an issue in #329 to track it.

Issue #326, opened 2021-08-24
Proper way to add custom logic types base on arrow2.


sundy-li:
Hello, Datafuse data memory runtime system is based on arrow2 and we really appreciate arrow2 to be such a great library.
Its physical type can meet our requirements, but we found it's hard to add custom logic types base on arrow2.

There are Arrow2's primary goals from README doc, such as the first one:
 `MUST NOT implement any logical type other than the ones defined on the arrow specification`
 
I think it's reasonable, but from the side of a database system, this may be not enough. If we want to have a logic DataType named `Date16`,
 whose physical type is `UInt16`, it represents the days range from `1970 - 2149`, this may cause more work on it.
 

In Datafuse:
  1. There is no mandatory requirement to follow the logic types as the arrow specification.
  2. Did not have a plan to go through other query engines base on arrow.
  3. Currently based on arrow2's array/compute/io and flight services.

We have two main ways to implement custom logic types. @jorgecarleitao  Would you give us some advices about it?

1. Fork another version of arrow2, add the logic datatype into enum DataType directly, and make others(such as: `io/ipc`) work well with it. This may be something
similar to https://github.com/tensorbase/tensorbase/blob/53366bb9bc17271096ca31c7861f36a39305c9a7/crates/arrow/src/datatypes/datatype.rs#L97 , but it breaks the baseline codes of arrow2,
 which makes it hard to contribute back upstream.

2. Based on the upstream codes of arrow2, since arrow2's physical type is enough for us, we just need to introduce logic types. 
But we have to introduce our own `io/ipc` and `io/parquet` and flight frameworks to work with our custom types.

jorgecarleitao:
Thanks for reaching out! 

The Arrow spec supports this; it is called [extension types](https://arrow.apache.org/docs/python/extending_types.html#defining-extension-types-user-defined-types). I haven't had the time to implement them yet, but it is very well within the arrow spec, so we are good to go on that front 👍

jorgecarleitao:
Otherwise, a general approach would be to generalize the types:

```
enum DataType {
    ArrowType(Arrow::DataType),
    CustomType1,
    CustomType2,
}
```

and then:

* implement `Into<Datatype> for Arrow::DataType` that is lossy (for ease of use) that falls into arrow's physical in-memory datatype, used within arrow2 compute and stuff
[...]

PR #338, opened 2021-08-25
Try investigate extension types

sundy-li:
Related to #326

### DIFF
--- a/src/ffi/schema.rs
+++ b/src/ffi/schema.rs
@@ -356,6 +356,8 @@ fn to_format(data_type: &DataType) -> String {
             r
         }
         DataType::Dictionary(index, _) => to_format(index.as_ref()),
+        //TODO: get format from metadata
+        DataType::Extension(name, ty, m) => to_format(ty),
     }
 }
 
--- a/src/ffi/schema.rs
+++ b/src/ffi/schema.rs
@@ -356,7 +356,7 @@ fn to_format(data_type: &DataType) -> String {
             r
         }
         DataType::Dictionary(index, _) => to_format(index.as_ref()),
-        DataType::Extension(ty) => format!("{:?}", ty),
+        DataType::Extension(ty) => to_format(ty.data_type()),
     }
 }
 
--- a/src/ffi/schema.rs
+++ b/src/ffi/schema.rs
@@ -357,7 +357,7 @@ fn to_format(data_type: &DataType) -> String {
         }
         DataType::Dictionary(index, _) => to_format(index.as_ref()),
         //TODO: get format from metadata
-        DataType::Extension(name, ty, m) => to_format(ty),
+        DataType::Extension(_, ty, _) => to_format(ty),
     }
 }
 
--- a/src/ffi/schema.rs
+++ b/src/ffi/schema.rs
@@ -357,7 +357,7 @@ fn to_format(data_type: &DataType) -> String {
         }
         DataType::Dictionary(index, _) => to_format(index.as_ref()),
         //TODO: get format from metadata
-        DataType::Extension(_, ty, _) => to_format(ty),
+        DataType::Extension(ty) => ty.to_format().to_string(),
     }
 }
 
--- a/src/ffi/schema.rs
+++ b/src/ffi/schema.rs
@@ -356,8 +356,7 @@ fn to_format(data_type: &DataType) -> String {
             r
         }
         DataType::Dictionary(index, _) => to_format(index.as_ref()),
-        //TODO: get format from metadata
-        DataType::Extension(ty) => ty.to_format().to_string(),
+        DataType::Extension(ty) => format!("{:?}", ty),
     }
 }
 
--- a/src/io/ipc/convert.rs
+++ b/src/io/ipc/convert.rs
@@ -386,6 +386,7 @@ fn type_to_field_type(data_type: &DataType) -> ipc::Type {
         Union(_, _, _) => ipc::Type::Union,
         Struct(_) => ipc::Type::Struct_,
         Dictionary(_, v) => type_to_field_type(v),
+        Extension(_, ty, _) => type_to_field_type(ty),
     }
 }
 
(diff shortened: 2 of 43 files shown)


sundy-li:
I've added the function named `to_physical_type` to make the DataType dispatch more simple, with this we could reduce the refactor codes of the match.

But what's the proper way to refactor the function `get_value_display`?

https://github.com/jorgecarleitao/arrow2/blob/a52211da6c9431638cff31765e7e1ed57fe60cf5/src/array/display.rs#L34-L62

jorgecarleitao:
Those are good ideas!

Even though they are represented equally, imo there are two types of matches: matches of physical types and matches of logical types.

IMO the logical matches should error on extension types (e.g. `Extension + Int16` should error) except at IO / FFI boundaries. Matches of physical types should be mapped accordingly, exactly like you are doing.

Display is a logical construct (Date32 != Int32). For now, I would map the representation to something like `ExtensionArray[representation equal to the base logical type]`.

There is a case here to change the `DataType::Extension(_,_,_)` to `DataType::Extension(Box<dyn Extension>)` where `Extension` is a trait describing the extension including how individual items should be displayed.

sundy-li:
> DataType::Extension(Box<dyn Extension>) where Extension is a trait 

I agree with that.

So:

1. We should have a factory for users to register custom Extension types and create Extension Type by the metadata in field, is that right?  Or `ipc` deserialize/serialize can handle dynamic extension types.
2. We should make it be  `DataType::Extension(Arc<dyn Extension>) `, because Array is Send + Sync.

jorgecarleitao:
I would go for the latter: registering introduces state, potentially a singleton, etc. I try first a stateless solution with a trait containing what is needed for the IPC to deserialize it. Something like

```
pub trait Extension {
     fn name() -> &str;
     fn data_type() -> &DataType;
     fn metadata() -> &Option<HashMap<String,String>>;

     // optional, fall back to the standard `get_display_value`
     fn get_display_value() -> Box<dyn Fn ...>;
}
```

is enough for now?

sundy-li:
Ok, btw what's the error of object safe trait?  I looked at the docs of rust but did not get it.

error: the trait `extension::Extension` cannot be made into an object\nlabel: `extension::Extension` cannot be made into an object

```
pub trait Extension: std::fmt::Debug + Send + Sync + Hash + Ord {
    fn name(&self) -> &str;
    /// Returns physical_type
    fn data_type(&self) -> &DataType;
    fn metadata(&self) -> &Option<HashMap<String, String>>;

    // fn get_display_value<'a>(&self, array: &'a dyn Array) -> Box<dyn Fn(usize) -> String + 'a>;
}
```

jorgecarleitao:
It cannot be `Hash` or `Eq` or `PartialEq` because they require sized objects, which a trait object is not.

I think that we will need to implement `PartialEq for DataType` ourselves, calling the usual in all variants except the Extension, which we use the trait to derive the `PartialEq`, something like:

```rust
impl PartialEq for dyn Extension + '_ {
    fn eq(&self, other: &Self) -> bool {
        self.name() == other.name() && self.data_type() == other.data_type() && self.metadata() == other.metadata()
    }
}
```

sundy-li:
Two more problems:

1. Should  `Field::new()` copy the `metadata` from Extension types?
2. How to Serialize/Deserialize extension types in IPC ? 
  Seems we need to introduce a factory to r register creators and a type Lexer from String in metadata.

src/ffi/schema.rs
```diff
@@ -356,6 +356,8 @@ fn to_format(data_type: &DataType) -> String {
             r
         }
         DataType::Dictionary(index, _) => to_format(index.as_ref()),
+        //TODO: get format from metadata
+        DataType::Extension(ty) => ty.to_format().to_string(),
```
jorgecarleitao, line 360: The extension type does not have a format for FFI: that information is passed on the fields' metadata in the keys `ARROW::extension::name` and `ARROW::extension::metadata`.

src/io/ipc/convert.rs
```diff
@@ -657,6 +658,9 @@ pub(crate) fn get_fb_field_type<'a>(
                 children: Some(fbb.create_vector(&children)),
             }
         }
+        Extension(ex) => {
+            todo!("extension");
```
jorgecarleitao, line 662: here is where we map the extension to the IPC field. To see how this is done by the rest of the arrow implementations, write 

```rust
    #[test]
    fn read_generated_extension() -> Result<()> {
        test_file("1.0.0-littleendian", "generated_extension")
    }
```
on `tests/it/io/ipc/file.rs` and print the schema in the json format (i..e. `arrow_json` inside the function `read_gzip_json`. You will find a small `ARROW:extension:...` key and value, which is how extensions are written in the metadata.

src/io/ipc/gen/Schema.rs
```diff
@@ -782,9 +782,10 @@ impl Type {
     pub const LargeBinary: Self = Self(19);
     pub const LargeUtf8: Self = Self(20);
     pub const LargeList: Self = Self(21);
+    pub const Extension: Self = Self(22);
```
jorgecarleitao, line 785: we can't change this generated code as other implementations will not match it. extension types are shared in a different way, see comment above.

jorgecarleitao (commented):
Looks great so far!

Couple of points:

* [ ] remove `Ord` from `Field` and `DataType`, (why on earth this should be ordered?); only the json reader uses that it is brittle to say the least (I will PR separately)
* [ ] declare a separate `enum` for physical types so that it is easier to work with then (and separate them from the logical types)?
* [ ] the implementation of `to_bytes` should be made by us, based on the trait's information
* [ ] `to_format` is done by us and does not need to be part of the trait
* [ ] `hash` must be done by us and must be compatible with `PartialEq` and `Eq`
* [ ]The non-triviality here is that every array must now be able to convert itself to an extension type of its own compatible logical type, so that the extension type data is carried over with the array itself, so that e.g. consumers can annotate the arrays with it and use `array.data_type()` to match an extension type.
* [ ] The IPC changes should be reverted: they do not follow the Arrow spec; the spec is way easier when it comes to extension types :) They are just a piece of metadata written to the fields' schema.

PR #350, opened 2021-08-27
Added support for `Extension`

jorgecarleitao:
This PR adds support for Arrow's ["Extension type"](https://arrow.apache.org/docs/format/Columnar.html#extension-types).

This is represented by `DataType::Extension` and `ExtensionArray`, respectively.

For now, extension arrays are only supported to be shared via IPC (i.e. not FFI, since metadata is still not supported in FFI).

This PR adds an example demonstrating how to use it.

All of this is pending passing integration tests, as usual, as well as more tests covering the feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,7 @@ venv/bin/python parquet_integration/write_parquet.py
 * `MutableArray` API to work in-memory in-place.
 * faster IPC reader (different design that avoids an extra copy of all data)
 * IPC supports 2.0 (compression)
+* Extension type supported
 * All implemented arrow types pass FFI integration tests against pyarrow / C++
 * All implemented arrow types pass IPC integration tests against other implementations
 
@@ -83,7 +84,7 @@ venv/bin/python parquet_integration/write_parquet.py
 ## Features in the original not available in this crate
 
 * Parquet read and write of struct and nested lists.
-* Map types
+* Map type
 
 ## Features in this crate not in pyarrow
 
--- a/examples/extension.rs
+++ b/examples/extension.rs
@@ -1,0 +1,51 @@
+use std::io::{Cursor, Seek, Write};
+use std::sync::Arc;
+
+use arrow2::array::*;
+use arrow2::datatypes::*;
+use arrow2::error::Result;
+use arrow2::io::ipc::read;
+use arrow2::io::ipc::write;
+use arrow2::record_batch::RecordBatch;
+
+fn write_ipc<W: Write + Seek>(writer: &mut W, array: impl Array + 'static) -> Result<()> {
+    // create a batch
+    let schema = Schema::new(vec![Field::new("a", array.data_type().clone(), false)]);
+
+    let mut writer = write::FileWriter::try_new(writer, &schema)?;
+
+    let batch = RecordBatch::try_new(Arc::new(schema), vec![Arc::new(array)])?;
+
+    writer.write(&batch)
+}
+
+fn read_ipc(reader: &[u8]) -> Result<RecordBatch> {
+    let mut reader = Cursor::new(reader);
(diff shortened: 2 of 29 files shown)


sundy-li:
> @sundy-li here is a proposal. Would it work for you? You would need to match as such:

Yes, this would work in my case, LGTM.

jorgecarleitao:
`Option<String>` is an optional opaque string. E.g. we can base64-encode anything on it.

sundy-li:
This pr looks good except for one thing:  how does the `compute` work for the extension array.

If `Date16` and `UInt16` share the same physical array (UInt16Array), the compute code will work for it automatically.

jorgecarleitao:
AFIAK the compute cannot make many assumptions about extension type because an extension type may have different semantics.

With that said, there are still a lot of things we can assume and forward the call to the corresponding function (e.g. we should still support filter, take, concat, etc).

Are those the type of compute operations that you are thinking about?

jorgecarleitao:
I have been thinking a lot about this and I am not sure this is the right way of doing this; I am starting to agree with you @sundy-li that we need the physical type idea here. I.e. not use  a registry, but still use the `logical_type -> physical_type -> downcast` together with all array types having a `DataType` instead of creating an `ExtensionArray`.

sundy-li:
Currently, this pr can meet our demand, so I decided to introduce custom logic types in Datafuse base on this pr.
But I'm not going to have `ExtensionArray` inside Datafuse, the `ExtensionArray` can only be seen in arrow's BatchRecord converted by Datafuse's `datablock`.

This may help in sharing some compute operations, like `Date16 sub Date16`, `Date16 sub/add UInt16`, `Date16 cmp UInt16` ... 


So I think this pr can be merged now because it did not break any old behaviors, let's take it as an experimental feature.

I do agree that `logical_type` should not be awarded in `Array`, the `data_type()` of `Array` always returns the physical type.  Let the schema/record batches take control of the `logical_type` may be better.

sundy-li:
Let calculations follow specific actual types, not logical types.

jorgecarleitao:
@ritchie46 , does polars use `Array::data_type` for other things other than to downcast the array to the correct implementation, or are logical datatypes carried over somewhere else (e.g. `RecordBatch` or `Series`). Asking because I think that @sundy-li proposal is quite interesting and may result in a lot of simplifications around data types and array construction.

ritchie46:
> @ritchie46 , does polars use Array::data_type for other things other than to downcast the array to the correct implementation, or are logical datatypes carried over somewhere else

Polars keep track of the data type itself on the `ChunkedArray<T>` This data type is mostly aligned with arrows data type, but does not have to be. For instance a polars `CategoricalType` has the following arrow arrays: `UInt32Array` and for comparisson its downcasted to `UInt32Type`. 

Logical type `Series` are downcasted to their physical type, compute is applied and upcasted again. So for `Date32` its casted to `Int32`, some compute is applied and its again casted back to `Date32`. This saves a lot of compiler bloat.

sundy-li:
Datafuse used polars's series idea, the `SeriesTrait` is good for type dispatch, we can handle lot's of type conversion between logic arrays and physical arrays.

But `ChunkedArray<T>` may not be a good idea since T is a logical type already, and `ChunkedArray<T>` hold `ArrayRef`.

So
 `Series  op  Series` 
 ----> `ChunkedArray<T> op Series`  (SeriesTrait) 
 --->     `ChunkedArray<T> op ChunkedArray<T>` (nomalize types) 
 ---->  `UInt32Array op UInt32Array` (downcast from ArrayRef).

And polars did not use flight and IPC, so currently it's easy to have custom logic data types inside it.

jorgecarleitao:
Closing in favor of #359

PR #357, opened 2021-08-30
Prepare APIs for extension types

jorgecarleitao:
This PR is a refactor to enable (logical) extension types.

It is motivated by the design described by @sundy-li in #350.

Main change: all arrays (e.g. `Utf8Array`) now require `DataType` as an argument of `from_data`, `new_null` and `new_empty`.


jorgecarleitao:
@sundy-li , I kept the `DataType` on all arrays for now since it is a non-trivial change to decouple arrays from `DataType` because the logical information is currently necessary in many compute, io and display kernels.

sundy-li:
Ok, that's reasonable.

jorgecarleitao:
@sundy-li @ritchie46 , I found a beautiful representation of the primitive types that simplifies a lot of code (-400 LOC). I pushed to this PR and updated the user guide accordingly.

Something like

```rust
enum PrimitiveType {
    Int8,
    Int16,
    Int32,
    Int64,
    Int128,
    ...
    DaysMs,
}

enum PhysicalType {
    /// A Null with no allocation.
    Null,
    /// A boolean represented as a single bit.
    Boolean,
    /// An array where each slot has a known compile-time size.
    Primitive(PrimitiveType),
    ...
}
```

this allow us to write things like 

```rust
    match values.data_type().to_physical_type() {
        Primitive(primitive) => with_match_primitive_type!(primitive, |$T| {
            let values = values.as_any().downcast_ref().unwrap();
            Ok(Box::new(primitive::take::<$T, _>(&values, indices)))
        }),
        Utf8 => {
            let values = values.as_any().downcast_ref().unwrap();
            Ok(Box::new(utf8::take::<i32, _>(values, indices)))
        }
        ...
```

which shaved LOC via macros (and will likely do the same for dependents).

Basically, instead of matching `PhysicalType::Int32`, match `PhysicalType::Primitive(PrimitiveType::Int32)`.

sundy-li:
Great, we have similar like `with_match_primitive_type` in  https://github.com/datafuselabs/datafuse/blob/7a3bfc5620e545333ca4e35c71ac849b421b74fc/common/functions/src/aggregates/macros.rs#L15-L28

But your implementation is more graceful because the `__with_ty__` is private macro.

PR #359, opened 2021-08-30
Added support for `Extension` (logical) type

### LINKED ISSUE #361: Added `Extension` to `DataType`
To add dedicated support for (logical) extension types.

jorgecarleitao:
Adds support for Arrow's `Extension` "DataType", enabling users to define and use custom data types from Arrow's ecosystem.

Note that, from now on, array's `data_type` is not sufficient to downcast an array. I have updated the user guide with the new rules. The easiest way to think about them:

* `enum PhysicalType` has a one-to-one relationship to concrete arrays (e.g. `PhysicalType::Utf8 <-> Utf8Array<i32>`)
* `enum DataType` has a many-to-one relationship with a `PhysicalType`, which can be obtained via `.to_physical_type() -> PhysicalType`.

The corollary of this change is that, to correctly downcast arrays, users should now use `match array.data_type().to_physical_type()`. This is only relevant to the `ExtensionType`, which boxes a `DataType` which indicates its own logical type.

Thanks to @sundy-li for the initial implementation at #338 and follow-up discussions at #350.

Closes #361

### DIFF
--- a/src/datatypes/mod.rs
+++ b/src/datatypes/mod.rs
@@ -22,7 +22,7 @@ pub use schema::Schema;
 /// Nested types can themselves be nested within other arrays.
 /// For more information on these types please see
 /// [the physical memory layout of Apache Arrow](https://arrow.apache.org/docs/format/Columnar.html#physical-memory-layout).
-#[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub enum DataType {
     /// Null type, representing an array without values or validity, only a length.
     Null,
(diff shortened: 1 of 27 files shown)


src/types/mod.rs
```diff
@@ -35,7 +35,8 @@ macro_rules! create_relation {
         unsafe impl Relation for $native_ty {
             #[inline]
             fn is_valid(data_type: &DataType) -> bool {
-                matches!(data_type, $($impl_pattern)|+)
+                let physical_type = data_type.to_physical_type();
+                matches!(physical_type, $($impl_pattern)|+)
```
sundy-li, line 39: just use `data_type.to_physical_type() = $impl_pattern` ?

src/datatypes/mod.rs
```diff
@@ -1,8 +1,10 @@
 //! Metadata declarations such as [`DataType`], [`Field`] and [`Schema`].
 mod field;
+mod physical_type;
 mod schema;
 
 pub use field::Field;
+pub(crate) use physical_type::*;
```
jorgecarleitao, line 7: This needs to be public so that others can use it also to downcast accordingly.

PR #368, opened 2021-09-02
Added `Send+Sync` to `MutableBuffer`

jorgecarleitao:


jorgecarleitao:
> Why `MutableBuffer` is Sync? There are lots of methods inside that are not sync.

Implementing `Sync` means that a type guarantees that _references_ to its values can be shared between threads, without risk of a data race in safe Rust.

Values of type `&MutableBuffer<T>` are immutable and are thus are safe to share them.

In essence, it is the same reason [why `Vec<T>` is `Sync` iff `T` is `Sync`](https://doc.rust-lang.org/std/vec/struct.Vec.html#impl-Sync).

PR #369, opened 2021-09-02
Made most features not default

jorgecarleitao:
This is a request for feedback: basically, it makes most features non-default and introduced a new feature, "full", containing most features.

This follows the same pattern of Tokio-rs (which has many features and a "full"). This makes it easier to start without being overwhelmed by all the dependencies we can potentially bring.

To migrate, pass `--features full,benchmarks` (or select the features accordingly ^_^)


jorgecarleitao:
cc @sundy-li and @ritchie46 , as it impacts you.

PR #372, opened 2021-09-03
Added support for `async` parquet write

### LINKED ISSUE #371: Async parquet write
Hi !

First of all thanks for this library !

Would it be possible to add support for writing parquet file asynchronously ?

`parquet2` already implements such function.

GrandChaman:
This PR adds support for `async` write to parquet from a `Stream` of `CompressedPage`s. It is the full `async` counterpart of `arrow2::io::parquet::stream::write_stream`, enabling full `async` writes.

Closes #371


jorgecarleitao (approved):
Thank you for the interest and for the contribution! Looks great!

PR #379, opened 2021-09-04
Simplified enums in `datatypes`

jorgecarleitao:
* Removed `PartialOrd` nor `Ord`, since there is no natural order on them.
* Made `IntervalUnit` and `TimeUnit` `Copy`, since they are simple enums.

### DIFF
--- a/src/compute/arithmetics/time.rs
+++ b/src/compute/arithmetics/time.rs
@@ -36,20 +36,20 @@ fn create_scale(lhs: &DataType, rhs: &DataType) -> Result<f64> {
         | (DataType::Time32(timeunit_a), DataType::Duration(timeunit_b))
         | (DataType::Time64(timeunit_a), DataType::Duration(timeunit_b)) => {
             // The scale is based on the TimeUnit that each of the numbers have.
-            timeunit_scale(timeunit_a, timeunit_b)
+            timeunit_scale(*timeunit_a, *timeunit_b)
         }
         (DataType::Date32, DataType::Duration(timeunit)) => {
             // Date32 represents the time elapsed time since UNIX epoch
             // (1970-01-01) in days (32 bits). The duration value has to be
             // scaled to days to be able to add the value to the Date.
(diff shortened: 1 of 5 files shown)


yjshen:
Hi @jorgecarleitao , arrow-datafusion added `PartialOrd` on Expr, which relays on `DataType` to be also `PartialOrd`:
```rust
#[derive(Clone, PartialEq, PartialOrd)]
pub enum Expr {
....
 Cast {
        /// The expression being cast
        expr: Box<Expr>,
        /// The `DataType` the expression will yield
        data_type: DataType,    // <- need DataType PartialOrd as well.
    },
}
```
Introduced in https://github.com/apache/arrow-datafusion/issues/1014 to provide the capability of sorting expressions, for common-subexpression-elimination and further expression transformations. any thoughts on this?

jorgecarleitao:
Ufff 😢

What is the semantics that the partial ord is trying to achieve in DataFusion?

I am a bit surprised that it is relying on the derived `PartialOrd` - semantically it seems to me that the order of the enum is meaningless. Is it to have a stable something?

yjshen:
Yes, I think so, for stabilizing tests (https://github.com/influxdata/influxdb_iox/pull/2564#discussion_r710289316)  and expression algebraic transformations, as the issue brought this in explained https://github.com/apache/arrow-datafusion/issues/1014

houqp:
yeah, i don't think the ordering semantic matters here, just need a consistent ordering defined.

houqp:
Pushed a workaround to implement partial ord using hash: https://github.com/houqp/arrow-datafusion/commit/99ee159fc8069c23829570126eadd63df0362678.

jorgecarleitao:
wow, quite neat idea! I had already given up and was dragging my feet to revert part of this PR and re-introduce partialOrd on 
* IntervalUnit
* TimeUnit
* Field
* DataType

I am so glad that you found a way there!

PR #386, opened 2021-09-07
Made GroupFilter optional in  parquet's`RecordReader` and added method to set it.

jorgecarleitao:
This is more expressive since filtering is optional.

The function `RecordReader::try_new` now expects `Option<GroupFilter>` instead of `GroupFilter`. If you were not using filtering, migrate to this by passing `None`. If you were filtering, pass `Some(...)` instead.

### DIFF
--- a/benches/read_parquet.rs
+++ b/benches/read_parquet.rs
@@ -1,5 +1,4 @@
 use std::io::Read;
-use std::sync::Arc;
 use std::{fs, io::Cursor, path::PathBuf};
 
 use criterion::{criterion_group, criterion_main, Criterion};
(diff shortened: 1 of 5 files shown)


houqp (approved):
Could we also expose `RecordReader.metadata` so it can be used to build the row group filter?

jorgecarleitao:
> Could we also expose `RecordReader.metadata` so it can be used to build the row group filter?

done 👍

houqp (approved):
LGTM

jorgecarleitao:
This has been released under [`v0.5.0`](https://github.com/jorgecarleitao/arrow2/releases/tag/v0.5.0), @houqp

houqp:
I am using git commit in my local arrow2 migration branch at the moment, but good to know we could release at a much faster cadence :D

PR #398, opened 2021-09-11
WIP: Added `Ord` for `days_ms` and `PartialOrd` for `months_days_ns`

houqp:
Follow up from https://github.com/jorgecarleitao/arrow2/pull/395.

### DIFF
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -6,6 +6,7 @@
 //! represent chunks of bits (e.g. `u8`, `u16`), and [`BitChunkIter`], that can be used to
 //! iterate over bitmaps in [`BitChunk`]s.
 //! Finally, this module also contains traits used to compile code optimized for SIMD instructions at [`mod@simd`].
+use std::cmp::Ordering;
 use std::convert::TryFrom;
 
 mod bit_chunk;
@@ -143,7 +144,7 @@ create_relation!(f32, PhysicalType::Primitive(PrimitiveType::Float32));
 create_relation!(f64, PhysicalType::Primitive(PrimitiveType::Float64));
 
 /// The in-memory representation of the DayMillisecond variant of arrow's "Interval" logical type.
-#[derive(Debug, Copy, Clone, Default, PartialEq, Eq, Hash)]
+#[derive(Debug, Copy, Clone, Default, Eq, Hash)]
 #[allow(non_camel_case_types)]
 pub struct days_ms([i32; 2]);
 
@@ -225,8 +226,36 @@ impl days_ms {
     }
 }
 
+const MS_IN_DAY: i32 = 86_400_000;
+
+impl Ord for days_ms {
+    fn cmp(&self, other: &Self) -> Ordering {
+        let self_days = self.days() + self.milliseconds() / MS_IN_DAY;
+        let other_days = other.days() + other.milliseconds() / MS_IN_DAY;
+        let ord = self_days.cmp(&other_days);
+        match ord {
+            Ordering::Greater | Ordering::Less => ord,
+            Ordering::Equal => {
+                (self.milliseconds() % MS_IN_DAY).cmp(&(other.milliseconds() % MS_IN_DAY))
+            }
+        }
+    }
+}
+
+impl PartialOrd for days_ms {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl PartialEq for days_ms {
+    fn eq(&self, other: &Self) -> bool {
+        self.cmp(other) == Ordering::Equal
+    }
+}
+
 /// The in-memory representation of the MonthDayNano variant of the "Interval" logical type.
-#[derive(Debug, Copy, Clone, Default, PartialEq, Eq, Hash)]
+#[derive(Debug, Copy, Clone, Default, Eq, Hash)]
 #[allow(non_camel_case_types)]
 #[repr(C)]
 pub struct months_days_ns(i32, i32, i64);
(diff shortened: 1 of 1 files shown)


src/types/mod.rs
```diff
@@ -144,7 +145,7 @@ create_relation!(f32, PhysicalType::Primitive(PrimitiveType::Float32));
 create_relation!(f64, PhysicalType::Primitive(PrimitiveType::Float64));
 
 /// The in-memory representation of the DayMillisecond variant of arrow's "Interval" logical type.
-#[derive(Debug, Copy, Clone, Default, PartialEq, Eq, Hash)]
+#[derive(Debug, Copy, Clone, Default, Eq, Hash)]
```
houqp, line 148: The derived PartialEq is actually not correct because it's not taking ms overflow into account.

src/types/mod.rs
```diff
@@ -226,8 +227,36 @@ impl days_ms {
...
+
+impl PartialEq for days_ms {
+    fn eq(&self, other: &Self) -> bool {
+        self.cmp(other) == Ordering::Equal
+    }
+}
+
 /// The in-memory representation of the MonthDayNano variant of the "Interval" logical type.
-#[derive(Debug, Copy, Clone, Default, PartialEq, Eq, Hash)]
+#[derive(Debug, Copy, Clone, Default, Eq, Hash)]
```
houqp, line 259: Same here, PartialEq is not implemented using PartialOrd.

houqp:
also cc @emkornfield

jorgecarleitao:
Interesting indeed.

I am trying to get my head around the `Eq` and `PartialEq` here. The condition for `Eq`:

```
* reflexive: a == a
* symmetric: a == b implies b == a
* transitive: a == b and b == c implies a == c
```

and `PartialEq` are

```
* symmetric: a == b implies b == a
* transitive: a == b and b == c implies a == c.
```

Don't we have reflexivity in both types?

I.e. isn't `[1, MS_IN_DAY] == [2, 0]` and `[2, 0] == [1, MS_IN_DAY]` in this PR?


With that said, I am not sure whether we can assume that `[1, MS_IN_DAY] == [2, 0]`: doesn't this assume that there are no leap seconds? I understood the goal of `days_ms` was to preserve leap seconds (the say way `months_days_ns` was used to preserve leap seconds and leap days (I am no expert here, though). Otherwise, isn't `Interval(DaysMs)` logically equivalent to `Duration(Milliseconds)`?

jorgecarleitao:
I am no longer sure about my last comment. We should in fact consider MS to be full days.

houqp:
> Don't we have reflexivity in ms_days?

Yes, that's why I kept the `Eq` trait for both types.

> I am no longer sure about my last comment. We should in fact consider MS to be full days.

Why so? I think we should take leap seconds into account and not return an order when `days_ms.milliseconds` is greater than `86_400_000`.

houqp:
Actually, after taking a second look at `Schema.fbs`, I am agreeing with you that we shouldn't take leap seconds into account. In the comment, leap seconds are excluded for both Date, TimeUnit and Timestamp types. There is no mention of leap second exclusion for IntervalUnit, but I think it would be strange if we handle leap seconds differently only for Interval types.

On the other hand, I found found this comment:

```
// A "calendar" interval which models types that don't necessarily
// have a precise duration without the context of a base timestamp (e.g.
// days can differ in length during day light savings time transitions).
// All integers in the types below are stored in the endianness indicated
// by the schema.
```

This means we need to take day light savings into account :facepalm: So we still won't be able to define a total order for days_ms. Oh god, I hate day light savings.

houqp:
Sent a PR to `Schema.fbs` to formalize handling of leap days and seconds for interval types. Once that's closed, I will update this PR to account for day light saving change.

src/types/mod.rs
```diff
@@ -226,8 +227,36 @@ impl days_ms {
     }
 }
 
+const MS_IN_DAY: i32 = 86_400_000;
+
+impl Ord for days_ms {
+    fn cmp(&self, other: &Self) -> Ordering {
+        let self_days = self.days() + self.milliseconds() / MS_IN_DAY;
```
jorgecarleitao, line 234: an alternative implementation is to wrap the milliseconds on `new` instead of on equality. I.e. `new(1,MS_IN_DAY)` is stored as `[2,0]` instead of `[1,MS_IN_DAY]`. Not sure which is better. Same applies for NS in `MonthsDaysNs`

src/types/mod.rs
jorgecarleitao: If we remove the `PartialEq` from here, I think we should also remove the `Eq`, since they should agree (and this makes them to disagree).

src/types/mod.rs
```diff
@@ -226,8 +227,36 @@ impl days_ms {
...
+    }
+}
+
+impl PartialOrd for days_ms {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl PartialEq for days_ms {
```
jorgecarleitao, line 252:
```suggestion
impl Eq for days_ms {
```

The easiest way imo is to manually implement `Eq` (instead of `PartialEq`): `PartialEq` is automatically derived when an implementation of `Eq` exists.

I think that this is also valid for the `MonthsDaysNs`: in both cases reflexivity exists and thus we can implement `Eq` (instead of `PartialEq`).

jorgecarleitao (commented):
Thanks a lot, @houqp ; left some comments for consistency sake, but it is ready to go once we have the clarification about `days_ms`.

fwiw, if this blocks you on DataFusion, I would be fine returning always `None` for now on the order. I think we will need to migrate DataFusion for `MonthsDaysNs` to be consistent with postgres, anyway.

Time is a pain :/

houqp:
No worry about it blocking datafusion :) I have switched to patching arrow2 to my local check out so all is good.

Changing this PR back to draft mode while waiting for upstream format reviews.

houqp:
ok, @jorgecarleitao following on this, given the dev list discussion, how do you prefer us proceed from here. Should we leave order and equality traits for these two types outside of arrow2 crate and implement them within datafusion using wrapper types? Or do you still prefer to have a default implementation of these traits in arrow2?

jorgecarleitao:
Let's have them here as we did for the others: the only concern was whether the leap seconds are included or not, but this does not really change anything here since we just need to be aware that they are not: the `partialOrd` in this PR makes a lot of sense. 👍

jorgecarleitao:
Let me know if you would like me to complete this one.

houqp:
Sorry, I had this one in my backlog, will pick it up today or tomorrow. I am not very happy with the current implementation because the semantic is too complex for users to understand. I am going to push another refactor to dumb it down a little bit.

jorgecarleitao:
> I am not very happy with the current implementation because the semantic is too complex for users to understand

❤️

houqp:
There is more discussions in https://issues.apache.org/jira/browse/ARROW-14122 following my clarification dev list thread. We should continue the discussion there and see if we can come up with a behavior that's consistent between CPP an Rust implementation.

houqp:
@jorgecarleitao I think we have collected enough data points from upstream to proceed with this change now :) Do you prefer us to drop the PartitialOrd trait entirely or make it to always return None? I think dropping it entirely might make more sense so the type system will prevent users from ordering it at build time and force them to do the casting or define their own ordering. WDYT?

jorgecarleitao:
I agree with you. Let's drop it to avoid difficult to track semantic bugs downstream.

houqp:
alright, issue resolved, no code is the best code ;)

PR #399, opened 2021-09-12
Add Array::with_validity

ritchie46:
This adds `with_validity(&self, validity: Option<Bitmap>) -> Result<Box<dyn Array>>` to the `dyn Array` trait. This allows creating a new array with the new validity without downcasting.

### DIFF
--- a/src/array/binary/mod.rs
+++ b/src/array/binary/mod.rs
@@ -97,9 +97,9 @@ impl<O: Offset> BinaryArray<O> {
 
     /// Sets the validity bitmap on this [`BinaryArray`].
     /// # Panic
-    /// This function panics iff `validity.len() < self.len()`.
+    /// This function panics iff `validity.len() != self.len()`.
     pub fn with_validity(&self, validity: Option<Bitmap>) -> Self {
-        if matches!(&validity, Some(bitmap) if bitmap.len() < self.len()) {
+        if matches!(&validity, Some(bitmap) if bitmap.len() != self.len()) {
             panic!("validity should be as least as large as the array")
         }
         let mut arr = self.clone();
--- a/src/array/binary/mod.rs
+++ b/src/array/binary/mod.rs
@@ -1,4 +1,9 @@
-use crate::{bitmap::Bitmap, buffer::Buffer, datatypes::DataType};
+use crate::{
+    bitmap::Bitmap,
+    buffer::Buffer,
+    datatypes::DataType,
+    error::{ArrowError, Result},
+};
 
 use super::{
     display_fmt, display_helper, specification::check_offsets, specification::Offset, Array,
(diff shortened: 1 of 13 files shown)


src/array/boolean/mod.rs
```diff
@@ -118,6 +119,16 @@ impl Array for BooleanArray {
     fn slice(&self, offset: usize, length: usize) -> Box<dyn Array> {
         Box::new(self.slice(offset, length))
     }
+    fn with_validity(&self, validity: Option<Bitmap>) -> Result<Box<dyn Array>> {
```
jorgecarleitao, line 122: What do you think about the following changes:

1. panic instead of `result`, to be aligned with the `from_data` where we also panic if we pass arguments with the wrong length (and also aligned with the usual bound checks from `Rust`)
2. Return `BooleanArray` instead of `Box<dyn Array>`, so that people can keep the typing information if they want to use the typed arrays

ritchie46: 1. Sure, I'll change that :+1: .
2. I can't. The `Array` trait is implemented for any array implementation, meaning that `with_validity` will return a different implementation of that trait, therefore I wrapped in that trait object.

jorgecarleitao: ah, wrt to 2, what about how we are doing for other "trait methods" (such as `clone` and `slice`):

```rust
impl BooleanArray {
     pub fn with_validity(validity: Option<Bitmap>) -> BooleanArray {
          todo!();
     }
}
```

and 

```rust
impl Array {
    fn with_validty(validity: Option<Bitmap>) -> Box<dyn Array> {
        match self.data_type().to_physical_type() {
            PhysicalType::Boolean => Box::new(self.as_any().downcast_ref::<BooleanArray>().unwrap().with_validity(validity))
        }
    }
}
```

ritchie46: Oh, right. That makes sense.

src/array/binary/mod.rs
```diff
@@ -94,6 +94,18 @@ impl<O: Offset> BinaryArray<O> {
             offset: self.offset + offset,
         }
     }
+
+    /// Sets the validity bitmap on this [`BinaryArray`].
+    /// # Panic
+    /// This function panics iff `validity.len() < self.len()`.
+    pub fn with_validity(&self, validity: Option<Bitmap>) -> Self {
+        if matches!(&validity, Some(bitmap) if bitmap.len() < self.len()) {
```
jorgecarleitao, line 102:
```suggestion
        if matches!(&validity, Some(bitmap) if bitmap.len() != self.len()) {
```

it must be equal to keep the invariant stronger so that people can use `validity.len()` or `values.len()` when iterating item by item.

jorgecarleitao (commented):
Very close; imo we should keep them equal to keep the invariant useful for people that iterate over one or the other without fear of going out of bounds.

src/array/binary/mod.rs
ritchie46: > Very close; imo we should keep them equal to keep the invariant useful for people that iterate over one or the other without fear of going out of bounds.

I was just thinking if that could lead to problems. Agreed.

jorgecarleitao:
Thanks a lot, @ritchie46 !

PR #404, opened 2021-09-14
Improve `comparison` docs and re-export the array-comparing function

### LINKED ISSUE #349: Some `compare` function are not exported
I want to directly use specific types of comparison functions, such as `boolean::compare`, but I found that they are not exported.
But their scalar version such as `boolean_compare_scalar` are exported.

Should we make the following mods be `pub`?
```rust
mod boolean;
mod primitive;
mod utf8;
```
```rust
pub mod boolean;
pub mod primitive;
pub mod utf8;
```

HagaiHargil:
Closes #349.

This PR slightly improves the `compute::comparison` module:

1. It adds extensive top-level documentation with examples and doc-tests.
2. It adds or expands the documentation of the main comparison functions.
3. It re-exports the array-comparing functions in the statically-typed case.

One *breaking* change which I kept in a separate commit and I'm not quite sure about is to change the scalar-oriented comparison functions from using the `_non_null` variants to the standard variants, the ones which check whether the `rhs` is valid first. If this is unwanted for some reason I'll happily revert it back.

### DIFF
--- a/src/compute/comparison/binary.rs
+++ b/src/compute/comparison/binary.rs
@@ -121,6 +121,13 @@ fn gt_eq_scalar<O: Offset>(lhs: &BinaryArray<O>, rhs: &[u8]) -> BooleanArray {
     compare_op_scalar(lhs, rhs, |a, b| a >= b)
 }
 
+/// Compare two [`BinaryArray`]s using the given [`Operator`].
+///
+/// # Errors
+/// When the two arrays have different lengths.
+///
+/// Check the [crate::compute::comparison](module documentation) for usage
+/// examples.
 pub fn compare<O: Offset>(
     lhs: &BinaryArray<O>,
     rhs: &BinaryArray<O>,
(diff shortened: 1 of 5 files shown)


jorgecarleitao:
Thank you very much for this, @HagaiHargil ! There are some tests failing before we can merge this.

HagaiHargil:
Happy to help! 

Sorry for not fixing the tests, I didn't have time to run the tests locally as I had to attend to other matters. I was also hoping to receive feedback from you regarding the change from the `scalar_non_null` type to the `scalar` type of functions. Do you think I should revert my change?

jorgecarleitao:
The reason for `_non_null` is that I wanted to keep the `_scalar` reserved to be used by the scalar API, but either is fine. See how it feels best for you =)

HagaiHargil:
Sorry for the delay, hope everything looks OK now.

jorgecarleitao:
Beautifully written 🙇

PR #407, opened 2021-09-15
Split `io_csv` feature into `read_csv` and `write_csv`

### LINKED ISSUE #405: Split CSV feature gate
Could feature gate for csv-io be split in `read` and `write`? This would prevent some code compilation in polars that will not be used.

VasanthakumarV:
Closes #405

I completely replaced `io_csv` with `read_csv` and `write_csv`, didn't know if we wanted an umbrella `io_csv` feature that encapsulates the other two.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,7 +78,8 @@ rustdoc-args = ["--cfg", "docsrs"]
 [features]
 default = []
 full = [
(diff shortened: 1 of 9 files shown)


VasanthakumarV:
Closing this in favor of #408

PR #413, opened 2021-09-16
Add `extend`/`extend_unchecked` for `MutableUtf8Array`

### LINKED ISSUE #166: Add `extend` / `extend_unchecked` on builders/ mutablearrays
What do you think of extending the `MutableArray` API's with those methods?

The [MutableBuffer](https://jorgecarleitao.github.io/arrow2/docs/arrow2/buffer/struct.MutableBuffer.html#method.extend_from_trusted_len_iter) already exposes this API. Often it's more user-friendly to work on the builders, as they maintain the overhead of the correct composite of arrays (a bitmap, an offset, a value-array, etc). With an `extend_(unchecked)` we can probably amortize allocation/bound checks.

VasanthakumarV:
This PR adds `TrustedLen`-based `extend` and `extend_unchecked` methods for `MutableUtf8Array`.

### DIFF
--- a/src/array/utf8/mutable.rs
+++ b/src/array/utf8/mutable.rs
@@ -276,9 +276,9 @@ impl<O: Offset> MutableUtf8Array<O> {
         I: Iterator<Item = Option<P>>,
     {
         if self.validity.is_none() {
-            self.validity = Some(MutableBitmap::from_trusted_len_iter(
-                std::iter::repeat(true).take(self.len()),
-            ));
+            let mut validity = MutableBitmap::new();
+            validity.extend_constant(self.len(), true);
+            self.validity = Some(validity);
         }
 
         extend_from_trusted_len_iter(
--- a/src/array/utf8/mutable.rs
+++ b/src/array/utf8/mutable.rs
@@ -224,6 +224,75 @@ impl<O: Offset, P: AsRef<str>> FromIterator<Option<P>> for MutableUtf8Array<O> {
 }
 
 impl<O: Offset> MutableUtf8Array<O> {
+    /// Extends the [`MutableUtf8Array`] from an iterator of values of trusted len.
+    /// This differs from `extended_trusted_len` which accepts iterator of optional values.
+    #[inline]
+    pub fn extend_trusted_len_values<I, P>(&mut self, iterator: I)
+    where
+        P: AsRef<str>,
+        I: TrustedLen<Item = P>,
+    {
+        unsafe { self.extend_trusted_len_values_unchecked(iterator) }
+    }
+
+    /// Extends the [`MutableUtf8Array`] from an iterator of values of trusted len.
+    /// This differs from `extended_trusted_len_unchecked` which accepts iterator of optional
+    /// values.
+    /// # Safety
+    /// The iterator must be trusted len.
+    #[inline]
+    pub unsafe fn extend_trusted_len_values_unchecked<I, P>(&mut self, iterator: I)
+    where
+        P: AsRef<str>,
+        I: Iterator<Item = P>,
+    {
+        let (_, upper) = iterator.size_hint();
+        let additional = upper.expect("extend_trusted_len_values requires an upper limit");
+
+        extend_from_trusted_len_values_iter(&mut self.offsets, &mut self.values, iterator);
+
+        if let Some(validity) = self.validity.as_mut() {
+            validity.extend_constant(additional, true);
+        }
+    }
+
+    /// Extends the [`MutableUtf8Array`] from an iterator of trusted len.
+    #[inline]
+    pub fn extend_trusted_len<I, P>(&mut self, iterator: I)
+    where
+        P: AsRef<str>,
+        I: TrustedLen<Item = Option<P>>,
+    {
+        unsafe { self.extend_trusted_len_unchecked(iterator) }
+    }
+
+    /// Extends [`MutableUtf8Array`] from an iterator of trusted len.
+    /// #Safety
+    /// The iterator must be trusted len.
+    #[inline]
+    pub unsafe fn extend_trusted_len_unchecked<I, P>(&mut self, iterator: I)
+    where
+        P: AsRef<str>,
+        I: Iterator<Item = Option<P>>,
+    {
+        if self.validity.is_none() {
+            self.validity = Some(MutableBitmap::from_trusted_len_iter(
+                std::iter::repeat(true).take(self.len()),
+            ));
+        }
+
+        extend_from_trusted_len_iter(
+            &mut self.offsets,
+            &mut self.values,
+            &mut self.validity.as_mut().unwrap(),
+            iterator,
+        );
+
+        if self.validity.as_mut().unwrap().null_count() == 0 {
+            self.validity = None;
+        }
+    }
+
     /// Creates a [`MutableUtf8Array`] from an iterator of trusted length.
     /// # Safety
     /// The iterator must be [`TrustedLen`](https://doc.rust-lang.org/std/iter/trait.TrustedLen.html).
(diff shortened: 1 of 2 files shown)


src/array/utf8/mutable.rs
```diff
@@ -223,7 +223,145 @@ impl<O: Offset, P: AsRef<str>> FromIterator<Option<P>> for MutableUtf8Array<O> {
...
+        let mut length = self.last_offset();
+
+        for item in iterator {
+            let bytes = item.as_ref().as_bytes();
+
+            length += O::from_usize(bytes.len()).unwrap();
+
+            values.extend_from_slice(bytes);
+            offsets.push(length);
+        }
```
VasanthakumarV, line 305: @jorgecarleitao Should I rewrite this function as a single loop?, to avoid creating `values` and `offset` vecs, this might mean we won't be utilizing MutableBuffer's `extend_from_trusted_len_iter_unchecked`

src/array/utf8/mutable.rs
```diff
@@ -223,7 +223,145 @@ impl<O: Offset, P: AsRef<str>> FromIterator<Option<P>> for MutableUtf8Array<O> {
...
+        P: AsRef<str>,
+        I: Iterator<Item = Option<P>>,
+    {
+        let (_, upper) = iterator.size_hint();
+        let additional = upper.expect("extend_trusted_len_unzip requires an upper limit");
+
+        // TODO How to reserve space for `values` and `validity` and take advantage
+        // of `push_unchecked`
+        self.offsets.reserve(additional);
+
```
VasanthakumarV, line 342: @jorgecarleitao Is it possible to deterministically assign reserves for `values` and `validity`?

`validity` is handled differently here than `MutablePrimitiveArray`, here we avoid allocating any space for it if all our inputs are valid, is this correct?

VasanthakumarV (commented):
@jorgecarleitao I started working on `MutableUtf8Array`, it passes a simple test case.

Will you be able to look at the code, and suggest changes, just to make sure I am on the correct track.

jorgecarleitao:
Thanks for picking this up! This is a great start!

Two notes: if we use a vector, we will need to double-allocate: one for the vector, one for the `self.values`. The goal is to extend the `self.values` directly.

An idea: create a new function (not method) `fn unsafe extend_from_trusted_len_iter_values` and modify `trusted_len_values_iter` to become something like

```rust
fn unsafe trusted_len_values_iter(iterator) {
    /// initialize the state 
    let capacity = iterator.size_hint().1.unwrap();
    let mut offsets = MutableBuffer::with_capacity(capacity + 1);
    offsets.push(O::default());
    let mut values = MutableBuffer::new();
    extend_from_trusted_len_iter_values(&mut offsets, &mut values, iterator);
    (offsets, values)
}
```

then, move the `for` loop logic currently in `trusted_len_values_iter` to `extend_from_trusted_len_iter_values`. This way, both creation and extension use the same code based (in `extend_from_trusted_len_iter_values`).

The same ideas can be applied to the one with the validity.

VasanthakumarV:
@jorgecarleitao Now,
- [`extend_trusted_len_values`](https://github.com/VasanthakumarV/arrow2/blob/extend/src/array/utf8/mutable.rs#L244) and [`from_trusted_len_values_iter`](https://github.com/VasanthakumarV/arrow2/blob/extend/src/array/utf8/mutable.rs#L325) share [`extend_from_trusted_len_values_iter`](https://github.com/VasanthakumarV/arrow2/blob/extend/src/array/utf8/mutable.rs#L545)
- [`extend_trusted_len`](https://github.com/VasanthakumarV/arrow2/blob/extend/src/array/utf8/mutable.rs#L273) and [`from_trusted_len_iter`](https://github.com/VasanthakumarV/arrow2/blob/extend/src/array/utf8/mutable.rs#L301) share [`extend_from_trusted_len_iter`](https://github.com/VasanthakumarV/arrow2/blob/extend/src/array/utf8/mutable.rs#L589)

If you can review once more, and provide feedback, I will feel more confident working on the others independently, and hopefully disturb you less often.

PS: All of the existing tests and new tests passed locally, I will spend more time cleaning up new tests written for this PR.

src/array/utf8/mutable.rs
```diff
@@ -481,33 +526,111 @@ where
     O: Offset,
     P: AsRef<str>,
     I: Iterator<Item = P>,
+{
+    let mut offsets = MutableBuffer::<O>::with_capacity(1);
```
jorgecarleitao, line 530: We can estimate this capacity to be `1 + iterator.size_hint().unwrap()`, which avoids an extra allocation :)

src/array/utf8/mutable.rs
```diff
@@ -481,33 +526,111 @@ where
...
+    values: &mut MutableBuffer<u8>,
+    iterator: I,
+) where
+    O: Offset,
+    P: AsRef<str>,
+    I: Iterator<Item = P>,
 {
     let (_, upper) = iterator.size_hint();
-    let len = upper.expect("trusted_len_unzip requires an upper limit");
+    let len = upper.expect("extend_from_trusted_len_iter_values requires an upper limit");
```
jorgecarleitao, line 555:
```suggestion
    let additional = upper.expect("extend_from_trusted_len_iter_values requires an upper limit");
```

and respective changes to make it clear that this is no longer the `len`, but how many more items are added.

src/array/utf8/mutable.rs
```diff
@@ -481,33 +526,111 @@ where
...
+    values: &mut MutableBuffer<u8>,
+    validity: &mut MutableBitmap,
+    iterator: I,
+) where
+    O: Offset,
+    P: AsRef<str>,
+    I: Iterator<Item = Option<P>>,
+{
+    let (_, upper) = iterator.size_hint();
+    let len = upper.expect("extend_from_trusted_len_values_iter requires an upper limit");
```
jorgecarleitao, line 600: same here.

jorgecarleitao (approved):
Awesome, looks great! 💯

Let two minor comments.

Feel free to move this to non-draft, change the title to be only for the `MutableUtf8Array` and we can merge this one.

src/array/utf8/mutable.rs
VasanthakumarV: This is neat, made the change, thanks

src/array/utf8/mutable.rs
VasanthakumarV: Renamed the variable

src/array/utf8/mutable.rs
VasanthakumarV: Renamed the variable

src/array/utf8/mutable.rs
```diff
@@ -224,6 +224,75 @@ impl<O: Offset, P: AsRef<str>> FromIterator<Option<P>> for MutableUtf8Array<O> {
...
+    #[inline]
+    pub unsafe fn extend_trusted_len_unchecked<I, P>(&mut self, iterator: I)
+    where
+        P: AsRef<str>,
+        I: Iterator<Item = Option<P>>,
+    {
+        if self.validity.is_none() {
+            self.validity = Some(MutableBitmap::from_trusted_len_iter(
+                std::iter::repeat(true).take(self.len()),
+            ));
```
jorgecarleitao, line 281:
```suggestion
            let mut validity = MutableBitmap::new();
            validity.extend_constant(self.len(), true);
            self.validity = Some(validity);
```

Missed this one, sorry! This is at least 5x faster. :)

(the reason it is faster is that an iterator populates bit by bit, while `extend_constant` populates byte by byte).

VasanthakumarV: Thanks for the explanation!, I have made the change

PR #414, opened 2021-09-16
Made `DynComparator` `Send+Sync`

yjshen:
This PR is a port of https://github.com/apache/arrow-rs/pull/542 . As I quote the rationale here:

> The cost of building a comparator (initialising a DynComparator) is often significantly higher than the actual cost of executing the comparator's closure on two row IDs. Therefore it makes sense to build the comparator once, and re-use the returned DynComparator for each row you are comparing the two arrays on.
>
> Due to the explicit lifetime of DynComparator I recently found it problematic to store the DynComparator on an object that was used across threads in an async environment.
>
> By refactoring the build_compare implementations I was able to remove the lifetime and it's much easier to use in the Datafusion operator I'm improving.

### DIFF
--- a/src/array/ord.rs
+++ b/src/array/ord.rs
@@ -7,7 +7,7 @@ use crate::error::{ArrowError, Result};
 use crate::{array::*, types::NativeType};
 
 /// Compare the values at two arbitrary indices in two arrays.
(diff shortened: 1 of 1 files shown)


yjshen:
Thank you @jorgecarleitao for the quick response.  👍

PR #417, opened 2021-09-17
Added support to add an interval to a timestamp

### LINKED ISSUE #23: Investigate how to add support for timezones in timestamp
We currently support timezones at logical level, via `Timestamp(_, tz)`, but have no mechanism to perform operations with them, The goal is to investigate how to add support for them. This will likely require depending on [chrono-tz](https://docs.rs/chrono-tz/0.5.3/chrono_tz/).

A simple concrete goal here is to add support to `cast` from and to `Timestamp(_, str)` and `Timestamp(_, None)`.

A question here is how we represent these things: it seems that in `chrono-tz`, timezones are static types (i.e. struct), while timezones in Arrow are dynamic (i.e. `str`).

jorgecarleitao:
This adds support for adding an interval to a timestamp with timezone, thus taking leap days (Feb 29th) and leap hours (daylight saving) into account.

This brings the last missing piece to fully support timestamps with timezones: we can now convert them from and to strings, and perform basic interval arithmetic taking the timezone into account.

3 examples:

* adding 1 hour to `2020-03-29 00:00:00 WET` results in `2020-03-29 02:00:00 WEST`
* adding 7 months and 1 hour to `2020-03-29 00:00:00 WET` results in `2020-10-29 01:00:00 WET`

the former crosses one summer time shift (and thus adds an extra hour), the latter crosses both (and thus adds no extra hour).

* adding 1d,1h,1m to `1972-02-28 01:00:00 +01:00` results in `1972-02-29 02:01:00 +01:00`

because 1972 has a leap year.

When the timestamp has no timezone information, days are `24 * 60 * 60`.

Closes #23

### DIFF
--- a/src/array/display.rs
+++ b/src/array/display.rs
@@ -66,56 +66,31 @@ pub fn get_value_display<'a>(array: &'a dyn Array) -> Box<dyn Fn(usize) -> Strin
             dyn_primitive!(array, i64, temporal_conversions::time64ns_to_time)
         }
         Time64(_) => unreachable!(), // remaining are not valid
-        Timestamp(TimeUnit::Second, tz) => {
+        Timestamp(time_unit, tz) => {
             if let Some(tz) = tz {
-                let offset = temporal_conversions::parse_offset(tz).unwrap();
-                dyn_primitive!(array, i64, |x| {
-                    chrono::DateTime::<chrono::FixedOffset>::from_utc(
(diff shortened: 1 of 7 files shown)


jorgecarleitao:
cc @sundy-li @ritchie46  since I saw you both working with interval arithmetics recently ^_^

PR #420, opened 2021-09-17
Changed IPC `FileWriter` to own the `writer`.

yjshen:


src/io/ipc/write/writer.rs
```diff
@@ -56,19 +56,20 @@ pub struct FileWriter<'a, W: Write> {
...
     }
 
     /// Try create a new writer with IpcWriteOptions
     pub fn try_new_with_options(
-        writer: &'a mut W,
+        writer: W,
         schema: &Schema,
         write_options: IpcWriteOptions,
     ) -> Result<Self> {
+        let mut writer = BufWriter::new(writer);
```
jorgecarleitao, line 72:
```suggestion
```

imo in general we are not the ones adding buffering, since we can't tell whether `W` is already buffered or not: users of the library are the ones making this decision based on the details of the `W` (e.g. cursor vs file).

src/io/ipc/write/writer.rs
```diff
@@ -91,6 +92,10 @@ impl<'a, W: Write> FileWriter<'a, W> {
         })
     }
 
+    pub fn into_inner(self) -> Result<W> {
+        self.writer.into_inner().map_err(|e| Error::from(e).into())
+    }
+
```
jorgecarleitao, line 98:
```suggestion

    /// Consumes itself returning the underlying writer.
    pub fn into_inner(self) -> Result<W> {
        self.writer
    }
```

src/io/ipc/write/writer.rs
```diff
@@ -153,12 +158,3 @@ impl<'a, W: Write> FileWriter<'a, W> {
         Ok(())
     }
 }
-
-/// Finish the file if it is not 'finished' when it goes out of scope
-impl<'a, W: Write> Drop for FileWriter<'a, W> {
```
jorgecarleitao, line 158: In general we need this or the file is not closed e.g. in the case of an error when writing or something, no?

jorgecarleitao (commented):
Looks great, very clean and very easy to follow 💯 

Left two comments where I am not following the rational and would like some help to understand.

src/io/ipc/write/writer.rs
yjshen: I'm following a similar pattern as the `StreamWriter` sibling to the `FilesWriter`. I suppose since the primary usage of `FileWriter` is to write files as its name implies. It's better to have a Buffer instead, or our users will often be required to remember to add a buffer layer when creating.

I'm willing to change this to which you think is more appropriate.

src/io/ipc/write/writer.rs
yjshen: The initial reason I noticed this `Drop` is by a compilation error :

```
   |         self.writer.into_inner().map_err(|e| Error::from(e).into())
   |         ^^^^^^^^^^^
   |         |
   |         cannot move out of here
   |         move occurs because `self.writer` has type `std::io::BufWriter<W>`, which does not implement the `Copy` trait

```

Then I think we could just remove it since if the `FileWriter` is not finished intentionally as the user requires, we are OK to leave a corrupt written file.
1. for user program crash during writing, the file is not correct no matter we finish it or not.
2. for users who forget to finish before exiting, we should train them to add a `finish` intentionally. instead of finish it `secretly`. 😉

src/io/ipc/write/writer.rs
houqp: I think @jorgecarleitao agrees with you on needs of buffering for writes. If I understand it correctly, he was implying that it would be better to require callers to remember to add that buffer layer instead. I think there are two benefits with that design:

1. users can choose to use different/custom buffering implementation if they prefer depending on their use-case and requirements
2. sometimes users already got a buffered writer to work with, in that case, they would need to remove the buffer layer first before passing the writer it in

jorgecarleitao: ☝️

To offer a reasoning: the need of buffering is usually driven by:
* the call patterns that we will use (many calls vs one call) and
* the internals of the writer `W` (e.g. whether each `.write_*` is expensive or not)

We could argue that our call pattern is a necessary condition for using buffering (we do make many calls), but the trait `Write` does not offer us the sufficient conditions for buffering imo. If our API was `P: Path` which we would open using `std::fs::File`, then we would most certainty add the buffering ourselves (and offer an additional parameter `buffer_size`, as part of the API).

jorgecarleitao: Thinking about this, I am not even sure we make so many calls that justify buffering, at least not without profiling: 

A file is composed by a header, a footer and messages. Each message is composed by a header and buffers, which we currently prepare in a single `Vec`. Thus, we make

* about 4 small `.write_all` calls (header + footer) per file
* 1 call message per message

which seems pretty small to me.

We could try to make this more efficient by not having the extra `vec`, but even this would still be small as it would add about 1 small call per message + 1 large call per buffer inside the message, or something like this.

yjshen: Thanks for the detailed explanation. Get it :D

PR #425, opened 2021-09-18
Added bench with `parquet` crate to read `parquet`

jorgecarleitao:
The results of running the benches are outlined here and summarized [here](https://docs.google.com/spreadsheets/d/12Sj1kjhadT-l0KXirexQDOocsLg-M4Ao1jnqXstCpx0/edit#gid=1919295045)


Dandandan:
I can replicate this benchmark, a bit more than 2x as fast.
Tried to replicate the config from https://docs.rs/parquet/5.4.0/src/parquet/file/properties.rs.html#57 + zstd compression

While it changes performance a bit (smaller page size is beneficial it seems), can not the results to be slower than the arrow2 bench.

Maybe there is something in particular with the output from parquet files generated by `parquet` + datafusion?

Dandandan:
I put the parquet files here for your reference

https://drive.google.com/drive/folders/19kUhg_6o8PuEOwCCLGQq0zh1bAq2TzG0?usp=sharing

jorgecarleitao:
❤️  amazing that you spotted where I was heading here and acted accordingly. Thank you very much!

I will go through that file and check what is happening - my suspicion is the (RLE) encoding, which is an under-determined problem (i.e. each writer can compress then in different ways).

Just to check, you mentioned an 8% difference in size; which one is smaller (parquet or parquet2)?

Dandandan:
even a bit more... `parquet2` generated files are smaller (133.7MB each vs 149.7MB using `parquet`)

Both with default settings + zstd + statistics.

Dandandan:
Looks like reading the latest column (index 15) is at least faster in `parquet` than in `parquet2`.

Dandandan:
One finding is that utf8 validation takes about 20% of the time for loading the full file.
Currently the code panics when the file contains invalid data in `check_offsets_and_utf8`.

I am wondering what parquet spec says about utf8 validation.

Using https://github.com/rusticstuff/simdutf8 might make sense too to speed up validation >= 8bytes

jorgecarleitao:
My understanding is that utf8 validation must be done or the reader must be marked as `unsafe` because user input (the file in this case) may lead to UB.

jorgecarleitao:
I ran these benches against the `part-0.parquet` generated from `parquet` via `cd benchmarks && cargo run --release --bin tpch -- convert --input ./data --output data/tpch-parquet --format parquet` on master.

The difference is ~10% (parquet2 faster) for column 13 (a string). I think that there are two aspects here:

1. all fields in `part-0.parquet` are required. Reading utf8 data should be very similar between `parquet` and `parquet2` since required data is just copying byte slices with no compute involved.
2. `parquet` is not performing `utf8` validation, therefore triggering UB in `safe` code (filed at https://github.com/apache/arrow-rs/issues/786)

arrow2 is 2x faster when we (unsafely) disable utf8 validation on it, so it seems to me that the main issue here is not arrow2 being slow, but that the comparison is unfair.

To sum up:
* we are comparing apples with oranges because `parquet+arrow` results in UB for invalid utf8 data
* arrow2 spends half of the time validating utf8 data, which is expected because that is the most expensive operation here (most ops are essentially memcopies)
* even with utf8 validation, arrow2 is still 10% faster than `parquet+arrow`

Dandandan:
> I ran these benches against the `part-0.parquet` generated from `parquet` via `cd benchmarks && cargo run --release --bin tpch -- convert --input ./data --output data/tpch-parquet --format parquet` on master.
> 
> The difference is ~10% (parquet2 faster) for column 13 (a string). I think that there are two aspects here:
> 
> 1. all fields in `part-0.parquet` are required. Reading utf8 data should be very similar between `parquet` and `parquet2` since required data is just copying byte slices with no compute involved.
> 2. `parquet` is not performing `utf8` validation, therefore triggering UB in `safe` code (filed at [Soundness: reading parquet with invalid utf8 results in UB apache/arrow-rs#786](https://github.com/apache/arrow-rs/issues/786))
> 
> arrow2 is 2x faster when we (unsafely) disable utf8 validation on it, so it seems to me that the main issue here is not arrow2 being slow, but that the comparison is unfair.
> 
> To sum up:
> 
> * we are comparing apples with oranges because `parquet+arrow` results in UB for invalid utf8 data
> * arrow2 spends half of the time validating utf8 data, which is expected because that is the most expensive operation here (most ops are essentially memcopies)
> * even with utf8 validation, arrow2 is still 10% faster than `parquet+arrow`

Thanks for the summary 💯! Seems like a great outcome of this inquiry!

Let's see what the difference is when `parquet` starts validating the output.
 
Interestingly, some non string columns where loading a bit slower for me on the `0.parquet` file - I guess there are still some opportunities for optimizating e.g. primitive / basic types.

jorgecarleitao:
The problem with this PR is that we drag `parquet` and `arrow` to `dev-dependencies`, which makes it more cumbersome to develop. I think that the solution here is to move these benches to a separate unpublished crate containing benches.

jorgecarleitao:
I plan to close this PR: I moved the benches to a separate repository: https://github.com/DataEngineeringLabs/parquet-benchmark, so that we do not populate this one with them.

PR #426, opened 2021-09-19
Improved performance of utf8 validation of large strings via `simdutf8` (-40%)

Dandandan:
This is helpful for strings >= 8 bytes.

### DIFF
--- a/src/array/specification.rs
+++ b/src/array/specification.rs
@@ -79,10 +79,8 @@ pub fn check_offsets_and_utf8<O: Offset>(offsets: &[O], values: &[u8]) -> usize 
         let end = window[1].to_usize();
         assert!(end <= values.len());
         let slice = unsafe { std::slice::from_raw_parts(values.as_ptr().add(start), end - start) };
-        #[cfg(feature = "simdutf8")]
-        simdutf8::basic::from_utf8(slice).expect("A non-utf8 string was passed.");
         #[cfg(not(feature = "simdutf8"))]
-        std::str::from_utf8(slice).expect("A non-utf8 string was passed.");
+        simdutf8::basic::from_utf8(slice).expect("A non-utf8 string was passed.");
     });
     len
 }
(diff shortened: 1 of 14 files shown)


jorgecarleitao:
Looks great!

I am tempted to define `utils::from_utf8` that branches on the feature and use this function everywhere in the crate, so that we cover all cases. I can do it on a new PR, just wanted to check if this makes sense to you.

Dandandan:
My benchmarks indicate it's not an improvement on benchmarks / the TPC-H files.
I misread the documentation of `simdutf8` - it only kicks in from 64 bytes - which makes it more useful for larger text / documents / etc.

I don't think the overhead is very high but it seems less useful beyond larger documents in that case.

Dandandan:
Let me add some >= 64 bytes examples to the parquet benchmark first. I think it makes sense to have proof of being faster first.

Dandandan:
@jorgecarleitao 
I benchmarked this change in 1fbe8e3e1515f729aa5d73ecd2f1b22d9b37a299 - and it gives results of around -40% for a large string of emojis on reading the column from parquet - I think pretty impressive. I couldn't spot clear negative consequences on smaller texts, so looks like it is a useful feature to enable in dependencies / at other places where utf8 validation is done.
The `write_parquet` approach is a bit hard to get memory efficient though, so for now I reverted the addition of the test.

jorgecarleitao:
Looks great. IMO we can make `simdutf8` a non-optional feature and use it through the crate (likely until it is merged into `std`).

jorgecarleitao:
I've updated the title accordingly. Could you paste the summary of the benches on the description so that when someone visits this PR can have a quick glance?

src/array/specification.rs
```diff
@@ -79,7 +79,8 @@ pub fn check_offsets_and_utf8<O: Offset>(offsets: &[O], values: &[u8]) -> usize
         let end = window[1].to_usize();
         assert!(end <= values.len());
         let slice = unsafe { std::slice::from_raw_parts(values.as_ptr().add(start), end - start) };
-        std::str::from_utf8(slice).expect("A non-utf8 string was passed.");
+        #[cfg(not(feature = "simdutf8"))]
```
jorgecarleitao, line 82: I think this introduced a vulnerability as it no longer checks utf8 if `simdutf8` is disabled. Both branches should be here.

Dandandan: Ah - I meant to remove this `cfg` here as it's no longer optional.

jorgecarleitao: you are right since this is not optional (also think thinking the previous xD). Pushed a fix directly to main

PR #428, opened 2021-09-19
Optimize hashing using ahash and multiversioning

Dandandan:
This uses the `T::get_hash`, which gives some speedup over the builder.

Also move ahash to the compute feature and add multiversioning to select/specialize on necessary instructions.

### DIFF
--- a/src/compute/hash.rs
+++ b/src/compute/hash.rs
@@ -1,4 +1,5 @@
 use ahash::{CallHasher, RandomState};
+use multiversion::multiversion;
 use std::hash::Hash;
 
 macro_rules! new_state {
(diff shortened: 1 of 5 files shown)


src/compute/hash.rs
```diff
@@ -29,47 +19,42 @@ use crate::{
...
-    unary(
-        array,
-        |x| {
-            let mut hasher = new_hasher!();
-            x.hash(&mut hasher);
-            hasher.finish()
-        },
-        DataType::UInt64,
-    )
+    let state = new_state!();
```
sundy-li, line 25: The state is initialized once for one array, will it cause differenct hash result in another array?

Dandandan: No, it uses the same seeds each time.

jorgecarleitao:
Awesome, thanks a lot! I updated the title to match the findings to show up in the changelog nicely.

src/compute/hash.rs
sundy-li: Ok, but there is a hash builder inside `get_hash`, this may introduce extra allocate in each time, I did not find a way to improve that.

```
#[inline]
    fn get_hash<H: Hash + ?Sized, B: BuildHasher>(value: &H, build_hasher: &B) -> u64 {
        let mut hasher = build_hasher.build_hasher();
        value.hash(&mut hasher);
        hasher.finish()
    }
```

PR #433, opened 2021-09-21
Extract parts of datetime

### LINKED ISSUE #415: Support to extract other parts of a date
We currently support `hour` and `year`, it would be nice to support other parts as-well.

Bonus exercise: without copy-pasting all the code we already have for this.

VasanthakumarV:
This PR extends the datetime extract methods to support:

* year (previously supported)
* month
* day
* weekday
* iso_week
* hour (previously supported)
* minute
* second
* nanosecond

To `Date32`, `Date64`, `Timestamp(_, None)` and `Timestamp(_, Some(_))`, thereby offering broad support to use times. For timestamps with timezone, these methods take leap days (Feb 29th) and hours (summer time) into account.

Closes #415

### DIFF
--- a/src/compute/temporal.rs
+++ b/src/compute/temporal.rs
@@ -28,84 +28,82 @@ use crate::types::NaturalDataType;
 
 use super::arity::unary;
 
-fn extract_impl<T, A, F>(
-    array: &PrimitiveArray<i64>,
-    time_unit: TimeUnit,
-    timezone: T,
-    extract: F,
-) -> PrimitiveArray<A>
-where
-    T: chrono::TimeZone,
-    A: NativeType + NaturalDataType,
-    F: Fn(chrono::DateTime<T>) -> A,
-{
-    match time_unit {
-        TimeUnit::Second => {
-            let op = |x| {
-                let datetime = timestamp_s_to_datetime(x);
-                let offset = timezone.offset_from_utc_datetime(&datetime);
-                extract(chrono::DateTime::<T>::from_utc(datetime, offset))
-            };
-            unary(array, op, DataType::UInt32)
+/// Extracts the years of a temporal array as [`PrimitiveArray<i32>`].
+/// Use [`can_year`] to check if this operation is supported for the target [`DataType`].
+pub fn year(array: &dyn Array) -> Result<PrimitiveArray<i32>> {
+    match array.data_type() {
+        DataType::Date32 | DataType::Date64 | DataType::Timestamp(_, None) => {
+            date_like(array, DataType::Int32, |x| x.year())
         }
-        TimeUnit::Millisecond => {
-            let op = |x| {
-                let datetime = timestamp_ms_to_datetime(x);
-                let offset = timezone.offset_from_utc_datetime(&datetime);
-                extract(chrono::DateTime::<T>::from_utc(datetime, offset))
-            };
-            unary(array, op, A::DATA_TYPE)
+        DataType::Timestamp(time_unit, Some(timezone_str)) => {
+            let time_unit = *time_unit;
+            let timezone = parse_offset(timezone_str);
+
+            let array = array.as_any().downcast_ref().unwrap();
+
+            if let Ok(timezone) = timezone {
+                Ok(extract_impl(array, time_unit, timezone, |x| x.year()))
+            } else {
+                chrono_tz(array, time_unit, timezone_str, |x| x.year())
+            }
         }
-        TimeUnit::Microsecond => {
-            let op = |x| {
-                let datetime = timestamp_us_to_datetime(x);
-                let offset = timezone.offset_from_utc_datetime(&datetime);
-                extract(chrono::DateTime::<T>::from_utc(datetime, offset))
-            };
-            unary(array, op, A::DATA_TYPE)
-        }
-        TimeUnit::Nanosecond => {
-            let op = |x| {
-                let datetime = timestamp_ns_to_datetime(x);
-                let offset = timezone.offset_from_utc_datetime(&datetime);
-                extract(chrono::DateTime::<T>::from_utc(datetime, offset))
-            };
-            unary(array, op, A::DATA_TYPE)
-        }
+        dt => Err(ArrowError::NotYetImplemented(format!(
+            "\"year\" does not support type {:?}",
+            dt
+        ))),
     }
 }
 
-#[cfg(feature = "chrono-tz")]
-#[cfg_attr(docsrs, doc(cfg(feature = "chrono-tz")))]
-fn chrono_tz<F, O>(
-    array: &PrimitiveArray<i64>,
-    time_unit: TimeUnit,
-    timezone_str: &str,
-    op: F,
-) -> Result<PrimitiveArray<O>>
-where
-    O: NativeType,
-    F: Fn(chrono::DateTime<chrono_tz::Tz>) -> O,
-{
-    let timezone = parse_offset_tz(timezone_str)?;
-    Ok(extract_impl(array, time_unit, timezone, op))
+/// Extracts the months of a temporal array as [`PrimitiveArray<u32>`].
+/// Use [`can_month`] to check if this operation is supported for the target [`DataType`].
+pub fn month(array: &dyn Array) -> Result<PrimitiveArray<u32>> {
+    match array.data_type() {
+        DataType::Date32 | DataType::Date64 | DataType::Timestamp(_, None) => {
+            date_like(array, DataType::UInt32, |x| x.month())
+        }
+        DataType::Timestamp(time_unit, Some(timezone_str)) => {
+            let time_unit = *time_unit;
+            let timezone = parse_offset(timezone_str);
+
+            let array = array.as_any().downcast_ref().unwrap();
+
+            if let Ok(timezone) = timezone {
+                Ok(extract_impl(array, time_unit, timezone, |x| x.month()))
+            } else {
+                chrono_tz(array, time_unit, timezone_str, |x| x.month())
+            }
+        }
+        dt => Err(ArrowError::NotYetImplemented(format!(
+            "\"month\" does not support type {:?}",
+            dt
+        ))),
+    }
 }
 
-#[cfg(not(feature = "chrono-tz"))]
-fn chrono_tz<F, O>(
-    _: &PrimitiveArray<i64>,
-    _: TimeUnit,
-    timezone_str: &str,
-    _: F,
-) -> Result<PrimitiveArray<O>>
-where
-    O: NativeType,
-    F: Fn(chrono::DateTime<chrono_tz::Tz>) -> O,
-{
-    Err(ArrowError::InvalidArgumentError(format!(
-        "timezone \"{}\" cannot be parsed (feature chrono-tz is not active)",
-        timezone_str
-    )))
+/// Extracts the days of a temporal array as [`PrimitiveArray<u32>`].
+/// Use [`can_day`] to check if this operation is supported for the target [`DataType`].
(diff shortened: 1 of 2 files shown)


VasanthakumarV:
@jorgecarleitao I have added only `month` and `day` till now, wanted to get your feedback before I proceed further,

The idea is to combine [DateLike](https://docs.rs/chrono/0.4.19/chrono/trait.Datelike.html) components under one macro_rule, and [TimeLike](https://docs.rs/chrono/0.4.19/chrono/trait.Timelike.html) components on another.

src/compute/temporal.rs
```diff
@@ -246,63 +228,90 @@ pub fn can_hour(data_type: &DataType) -> bool {
     )
 }
 
+macro_rules! date_like {
+    ($component:ident, $array:ident, $data_type:path, $chrono_tz:ident) => {
```
VasanthakumarV, line 232: Having four arguments on a macro_rule, is definitely not very readable, I will try to bring this number down

jorgecarleitao: Would it be possible to convert this to a generic with parameters `A, F: Fn(DateTime) -> A`, and pass `|x| x.year()`, so that we do not rely on macros?

jorgecarleitao (commented):
Looks really good!

I am not sure ab out my comment, but I think it is worth a try to convert the macro to a generic with a function takes components out of the datelike object, but maybe it is not possible.

In general generics are more expressive and easier to read because they pinpoint the exact constraints (Trait) that are required to uphold the functionality.

VasanthakumarV:
I quite wasn't able to reach `A, D: DateLike F: Fn(D) -> A`, I had to settle for `A, F: Fn(NaiveDateTime) -> A`, 
for types that doesn't fit this pattern (`DataType::Timestamp(time_unit, Some(timezone_str))` requires `Fn(DateTime<FixedOffset>)` and `Fn(DateTime<Tz>`) I had to repeat code across `year`, `month` and `day`.

I wasn't able to make trait bounds work on closures arguments, I tried `D: DateLike, fn(D) -> A` once, but failed.

VasanthakumarV:
Let me investigate more and reduce the boilerplate.

VasanthakumarV:
@jorgecarleitao Do you think `Fn(&dyn DateLike) -> O` is a good idea,

Since at compile time we won't know if the input will be `NaiveTime`, `NaiveDateTime`, `DateTime<Tz>` or `DateTime<Offset>`, we might have to resort to dynamic dispatch.

Also, this could be costly because it is applied to each element of the array, but I am not sure.

---

Currently, we create one closure for each type with the help of `match` statements, in each `hour`, `minute`, `year` ... functions.

```rust
pub fn hour(array: &dyn Array) -> Result<PrimitiveArray<u32>> {
    match array.data_type() {
        DataType::Date32 | DataType::Date64 | &DataType::Timestamp(_, None) => {
            date_like(array, DataType::UInt32, |x| x.hour())
        }
        DataType::Time32(_) | DataType::Time64(_) => {
            time_like(array, DataType::UInt32, |x| x.hour())
        }
        DataType::Timestamp(time_unit, Some(timezone_str)) => {
            ...
            if let Ok(timezone) = timezone {
                Ok(extract_impl(array, time_unit, timezone, |x| x.hour()))
            } else {
                chrono_tz(array, time_unit, timezone_str, |x| x.hour())
            }
        }
    }
}
```

jorgecarleitao:
Wow, really good changes.

I think it is worth trying the `dyn`: I would create a small benchmark and see the impact. I suspect if we `#[inline]` the function, it won't have any impact, but I may be wrong ^_^

VasanthakumarV:
@jorgecarleitao Unfortunately `Datelike` and `Timelike` traits are not trait object safe, so `dyn` is not possible 😞.
We could create a new trait with the subset of methods we want, but implementing them on the four types can be too much.

---

I had to go back to `macro_rules`, but this time around it is minimal.

src/compute/temporal.rs
```diff
@@ -28,6 +28,243 @@ use crate::types::NaturalDataType;
...
+        self.iso_week().week()
+    }
+}
+
+impl U32IsoWeek for chrono::NaiveDateTime {}
+impl<T: chrono::TimeZone> U32IsoWeek for chrono::DateTime<T> {}
+
+macro_rules! date_like {
+    ($extract:ident, $array:ident, $data_type:path) => {
+        match $array.data_type() {
```
jorgecarleitao, line 51:
```suggestion
        match $array.data_type().to_logical_type() {
```

It change seems odd, but this is used to "desugar" the `DataType::Extension` type, so that if someone has an extension type built on top of e.g. `Timestamp` type, we also support these methods.

jorgecarleitao (commented):
You clearly enjoy writing simple, readable code in Rust =) I went through it and it is very pleasant to read it. I left minor suggestions, but feel free to iterate on it until you are comfortable.

One thing missing is the tests. We do not need to cover all variations since we are using macros to not duplicate code. IMO having tests for the timestamp with and without timezone is important, in particular

1. extract `hour` from a timestamp on a summer time. See [here](https://github.com/jorgecarleitao/arrow2/blob/main/tests/it/temporal_conversions.rs#L116) for an example where there is a summer time shift.
2. extract `day` of a leap day (29th of Feb). See [here](https://github.com/jorgecarleitao/arrow2/blob/main/tests/it/temporal_conversions.rs#L72) for an example of one. 

A note: formally `second(date64)` should be 0. It is a bit misleading in the arrow specification, but `date64` is a multiple of `86400000` (see [here](https://github.com/apache/arrow/blob/master/format/Schema.fbs#L205)) 🤯. IMO we should enforce it on the array construction, so it is fine to keep it like this here.

tests/it/compute/temporal.rs
```diff
@@ -2,173 +2,317 @@ use arrow2::array::*;
...
+                year: Some(Int32Array::from(&[Some(2021), None, Some(1972)])),
+                month: Some(UInt32Array::from(&[Some(1), None, Some(2)])),
+                day: Some(UInt32Array::from(&[Some(30), None, Some(29)])),
+                weekday: Some(UInt32Array::from(&[Some(6), None, Some(2)])),
+                iso_week: Some(UInt32Array::from(&[Some(4), None, Some(9)])),
+                hour: Some(UInt32Array::from(&[Some(16), None, Some(13)])),
+                minute: Some(UInt32Array::from(&[Some(57), None, Some(9)])),
+                second: Some(UInt32Array::from(&[Some(27), None, Some(30)])),
+                nanosecond: Some(UInt32Array::from(&[Some(0), None, Some(0)])),
+            },
```
VasanthakumarV, line 129: Test for leap year, taken from your example

tests/it/compute/temporal.rs
```diff
@@ -2,173 +2,317 @@ use arrow2::array::*;
...
+            year: Some(Int32Array::from(&[Some(2020), None])),
+            month: Some(UInt32Array::from(&[Some(3), None])),
+            day: Some(UInt32Array::from(&[Some(29), None])),
+            weekday: Some(UInt32Array::from(&[Some(7), None])),
+            iso_week: Some(UInt32Array::from(&[Some(13), None])),
+            hour: Some(UInt32Array::from(&[Some(2), None])),
+            minute: Some(UInt32Array::from(&[Some(0), None])),
+            second: Some(UInt32Array::from(&[Some(0), None])),
+            nanosecond: Some(UInt32Array::from(&[Some(0), None])),
+        },
```
VasanthakumarV, line 262: This test is for the summertime adjustment, taken directly from your example

VasanthakumarV (commented):
Thanks for the review @jorgecarleitao 

`nanosecond` method hasn't been properly tested yet,

Do you think we should add or remove any of the time extraction methods?

src/compute/temporal.rs
VasanthakumarV: Add this method in every place we call `data_type`

src/compute/temporal.rs
```diff
@@ -75,149 +312,72 @@ where
...
-    _: &PrimitiveArray<i64>,
-    _: TimeUnit,
-    timezone_str: &str,
-) -> Result<PrimitiveArray<u32>> {
-    Err(ArrowError::InvalidArgumentError(format!(
-        "timezone \"{}\" cannot be parsed (feature chrono-tz is not active)",
-        timezone_str
-    )))
+/// let data_type = DataType::Int8;
+/// assert_eq!(can_year(&data_type), false);
```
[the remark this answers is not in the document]
VasanthakumarV: Made the change

src/compute/temporal.rs
```diff
@@ -75,149 +312,72 @@ where
...
-}
+/// Checks if an array of type `datatype` can perform year operation
+///
+/// # Examples
+/// ```
+/// use arrow2::compute::temporal::can_year;
+/// use arrow2::datatypes::{DataType};
+///
+/// let data_type = DataType::Date32;
+/// assert_eq!(can_year(&data_type), true);
```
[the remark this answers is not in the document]
VasanthakumarV: Made the change

src/compute/temporal.rs
```diff
@@ -28,6 +28,243 @@ use crate::types::NaturalDataType;
...
+                .downcast_ref::<PrimitiveArray<i64>>()
+                .unwrap();
+            Ok(unary(array, |x| op(time64ns_to_time(x)), data_type))
+        }
+        _ => unreachable!(),
+    }
+}
+
+#[cfg(feature = "chrono-tz")]
+#[cfg_attr(docsrs, doc(cfg(feature = "chrono-tz")))]
```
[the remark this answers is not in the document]
VasanthakumarV: Removed it

src/compute/temporal.rs
```diff
@@ -28,6 +28,243 @@ use crate::types::NaturalDataType;
...
+    time_like!(second, array, DataType::UInt32)
+}
+
+/// Extracts the nanoseconds of a temporal array as [`PrimitiveArray<u32>`].
+/// Use [`can_nanosecond`] to check if this operation is supported for the target [`DataType`].
+pub fn nanosecond(array: &dyn Array) -> Result<PrimitiveArray<u32>> {
+    time_like!(nanosecond, array, DataType::UInt32)
+}
+
+pub fn date_variants<F, O>(
```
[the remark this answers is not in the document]
VasanthakumarV: Removed it

tests/it/compute/temporal.rs
```diff
@@ -2,173 +2,317 @@ use arrow2::array::*;
...
 
-#[test]
-fn time32_second_hour() {
-    let array = Int32Array::from(&[Some(37800), None]).to(DataType::Time32(TimeUnit::Second));
+impl TestData {
+    fn data(data_type: &DataType) -> TestData {
+        match data_type {
+            DataType::Date64 => TestData {
+                input: Box::new(
+                    Int64Array::from(&[Some(1514764800000), None, Some(1550636625000)])
```
VasanthakumarV, line 65: The third entry in this test might be a little misleading given 86400000 constraint on `date64`

jorgecarleitao:
Let me know when you think it is ready (since it is still marked as a draft)

VasanthakumarV:
> Let me know when you think it is ready (since it is still marked as a draft)

@jorgecarleitao Was trying just now to add a test that uses TimeUnit::Nanosecond, I wanted to check for today's day represented by Nanoseconds (163265170202000000000), Int64Array is not enough, Int128Array does not support Datestamp logical type, so wasn't sure what to do

jorgecarleitao:
Thank you again for this amazing work and PR 🙇 . I have modified the title and description to cover the great work done here so that anyone looking through the changelog can track it to this PR.

PR #438, opened 2021-09-22
Fixed panic in division using nulls.

### LINKED ISSUE #437: Integer divsion might panic if null values in array: unchecked division
The null values are often ignore in a division operation. This allows for less branching and thus more performant vectorized execution.

However the values behind the null mask often are zero. If we do integer division, this panics, even though the values are "masked out/invalid". I wanted to add https://doc.rust-lang.org/nightly/std/intrinsics/fn.unchecked_div.html but that is nightly only. :(

I don't know if there is a stable way to do unchecked division.

_edit: 
Another option might be a more expensive division implementation where we take slices of valid values and only impelment the division on those.
_

jorgecarleitao:
Close #437

There may be other more efficient implementations, but for now having the bug fixed is more important :)

### DIFF
--- a/src/compute/arithmetics/basic/div.rs
+++ b/src/compute/arithmetics/basic/div.rs
@@ -25,10 +25,10 @@ use strength_reduce::{
 /// use arrow2::compute::arithmetics::basic::div::div;
 /// use arrow2::array::Int32Array;
 ///
-/// let a = Int32Array::from(&[Some(10), Some(6)]);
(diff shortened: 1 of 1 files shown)


ritchie46:
Maybe we could keep the old implementation as fast path in case  both arrays don't have null values.

Dandandan:
> Maybe we could keep the old implementation as fast path in case both arrays don't have null values.

Great idea. Only needs no nulls on the right side I think (as the panic is for division by zero).

PR #442, opened 2021-09-23
Optimized `null_count`

ritchie46:
Every time we slice an array, we count the null values. This PR does a small optimization so that we only count he null values of the smallest chunk of memory.

### DIFF
--- a/src/bitmap/immutable.rs
+++ b/src/bitmap/immutable.rs
@@ -121,9 +121,16 @@ impl Bitmap {
     /// The caller must ensure that `self.offset + offset + length <= self.len()`
     #[inline]
     pub unsafe fn slice_unchecked(mut self, offset: usize, length: usize) -> Self {
+        // count the smallest chunk
+        if length < self.length / 2 {
+            // count the null values in the slice
+            self.null_count = count_zeros(&self.bytes, offset, length);
+        } else {
+            // subtract the null count of the chunk we slice off
+            self.null_count -= count_zeros(&self.bytes, self.offset, self.offset + offset)
+        }
         self.offset += offset;
         self.length = length;
-        self.null_count = count_zeros(&self.bytes, self.offset, self.length);
         self
     }
 
--- a/src/bitmap/immutable.rs
+++ b/src/bitmap/immutable.rs
@@ -7,6 +7,7 @@ use super::{
     utils::{count_zeros, fmt, get_bit, get_bit_unchecked, BitChunk, BitChunks, BitmapIter},
     MutableBitmap,
 };
+use num_traits::zero;
 
 /// An immutable container whose API is optimized to handle bitmaps. All quantities on this
 /// container's API are measured in bits.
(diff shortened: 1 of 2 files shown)


ritchie46:
Nevermind. This is incorrect. It's only valid if we split the array in two.

jorgecarleitao:
Cool idea!

I think it is worth benchmarking: we now have 2 (smaller) iterations instead of one. I can PR a bench for slicing bitmaps.

ritchie46:
> I think it is worth benchmarking: we now have 2 (smaller) iterations instead of one. I can PR a bench for slicing bitmaps.

Yes, I was also wondering this. Perhaps there is some more instruction level parallelism as they are independent. But you are right. Let's benchmark

ritchie46:
I ran this benchmark:

```rust
        let offset = ((size as f64) * 0.2) as usize;
        let len = ((size as f64) * 0.55) as usize;

        c.bench_function(&format!("bitmap_count_zeros {}", log2_size), |b| {
            b.iter(|| {
                let r = bitmap.clone().slice(offset, len);
                assert!(r.null_count() > 0);
            })
        });

```

So that we have a smaller chunk at the start and at the end of the array. We slice only 55% as this would be an almost worst case scenario (51% would be). So the result would be better if we slice bigger chunks. It seems we make a difference once we don't fit the cache anymore.


```
bitmap_count_zeros 10   time:   [31.303 ns 31.316 ns 31.329 ns]                                   
                        change: [+2.2863% +3.3347% +4.2056%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 5 outliers among 100 measurements (5.00%)
  1 (1.00%) low severe
  2 (2.00%) low mild
  2 (2.00%) high severe

bitmap_count_zeros 12   time:   [46.442 ns 46.496 ns 46.555 ns]                                   
                        change: [+13.943% +14.534% +15.153%] (p = 0.00 < 0.05)
                        Performance has regressed.
  5 (5.00%) high mild
  2 (2.00%) high severe

bitmap_count_zeros 14   time:   [91.293 ns 91.322 ns 91.370 ns]                                  
                        change: [-14.881% -14.622% -14.426%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low mild
  2 (2.00%) high mild
  4 (4.00%) high severe

bitmap_count_zeros 16   time:   [304.45 ns 305.34 ns 306.20 ns]                                  
                        change: [-14.044% -13.760% -13.461%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 21 outliers among 100 measurements (21.00%)
  8 (8.00%) low severe
  1 (1.00%) low mild
  8 (8.00%) high mild
  4 (4.00%) high severe

bitmap_count_zeros 18   time:   [1.1123 us 1.1126 us 1.1130 us]                                   
                        change: [-24.429% -23.034% -21.759%] (p = 0.00 < 0.05)
                        Performance has improved.
  3 (3.00%) high mild
  4 (4.00%) high severe

bitmap_count_zeros 20   time:   [4.3841 us 4.3865 us 4.3888 us]                                   
                        change: [-19.887% -19.629% -19.434%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild

```

src/bitmap/immutable.rs
```diff
@@ -7,6 +7,7 @@ use super::{
     utils::{count_zeros, fmt, get_bit, get_bit_unchecked, BitChunk, BitChunks, BitmapIter},
     MutableBitmap,
 };
+use num_traits::zero;
```
jorgecarleitao, line 10: is this needed? I can't find where we use it.

jorgecarleitao (approved):
Awesome; very good result. 

It may be worth creating a bench that you used, in case someone else would like to use it for future improvements.

src/bitmap/immutable.rs
ritchie46: Hmm.. Don't think so as well. :thinking:

ritchie46:
> It may be worth creating a bench that you used, in case someone else would like to use it for future improvements.

Added :+1:

jorgecarleitao:
Could you resolve the conflict?

ritchie46:
> Could you resolve the conflict?

Good to go.

PR #445, opened 2021-09-24
Improved performance of sum aggregation via aligned loads (-10%)

ritchie46:
This adds aligned load instruction to the SIMD aggregation of arrays without null values. This implementation works on any alignment, and does not require an aligned allocator, and also works on sliced arrays.

Benchmark is a bit mixed on small data sizes. It seems to improving with more data, I want to run one on more data later.

```
sum 2^10 f32            time:   [101.38 ns 101.41 ns 101.45 ns]                         
                        change: [+1.5061% +1.5401% +1.5774%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 9 outliers among 100 measurements (9.00%)
  2 (2.00%) high mild
  7 (7.00%) high severe

sum 2^12 f32            time:   [284.72 ns 285.05 ns 285.44 ns]                         
                        change: [+0.4929% +0.6963% +0.9300%] (p = 0.00 < 0.05)
                        Change within noise threshold.
  4 (4.00%) high mild
  3 (3.00%) high severe

sum 2^14 f32            time:   [1.0246 us 1.0246 us 1.0247 us]                          
                        change: [-4.1159% -3.6418% -3.1975%] (p = 0.00 < 0.05)
                        Performance has improved.
  4 (4.00%) high mild
  3 (3.00%) high severe

sum 2^16 f32            time:   [4.0764 us 4.0773 us 4.0783 us]                          
                        change: [+1.4638% +1.6504% +1.7873%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 9 outliers among 100 measurements (9.00%)
  3 (3.00%) high mild
  6 (6.00%) high severe

sum 2^18 f32            time:   [19.596 us 19.600 us 19.606 us]                          
                        change: [+0.2075% +0.2434% +0.2801%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 9 outliers among 100 measurements (9.00%)
  2 (2.00%) high mild
  7 (7.00%) high severe

sum 2^20 f32            time:   [67.890 us 67.909 us 67.931 us]                         
                        change: [-6.5180% -6.4803% -6.4365%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 12 outliers among 100 measurements (12.00%)
  11 (11.00%) high mild
  1 (1.00%) high severe


```

### DIFF
--- a/src/compute/aggregate/sum.rs
+++ b/src/compute/aggregate/sum.rs
@@ -26,10 +26,7 @@ where
     T: NativeType + Simd + Add<Output = T> + std::iter::Sum<T>,
     T::Simd: Sum<T> + Add<Output = T::Simd>,
 {
-    // Safety:
-    // T::Simd is the vector type T and the alignment is similar to aligning to [T; alignment]
-    // the alignment of T::Simd ensures that it fits T.
-    let (head, simd_vals, tail) = unsafe { values.align_to::<T::Simd>() };
+    let (head, simd_vals, tail) = T::Simd::align(values);
 
     let mut reduced = T::Simd::from_incomplete_chunk(&[], T::default());
     for chunk in simd_vals {
--- a/src/compute/aggregate/sum.rs
+++ b/src/compute/aggregate/sum.rs
@@ -19,43 +19,24 @@ pub trait Sum<T> {
     fn simd_sum(self) -> T;
 }
 
-fn split_by_alignment<U, T>(values: &[T]) -> (&[T], &[T]) {
-    let alignment = std::mem::align_of::<U>();
-
-    let vals_ptr = values.as_ptr();
-    let bytes_offset = vals_ptr.align_offset(alignment);
-    let type_offset = if bytes_offset > 0 {
-        std::mem::align_of::<T>() / bytes_offset
-    } else {
-        0
-    };
-
-    let head = &values[..type_offset];
-    let aligned_values = &values[type_offset..];
-    (head, aligned_values)
-}
-
 #[multiversion]
 #[clone(target = "x86_64+avx")]
-fn nonnull_sum<T>(values: &[T]) -> T
+fn nonnull_sum<'a, T>(values: &'a [T]) -> T
 where
     T: NativeType + Simd + Add<Output = T> + std::iter::Sum<T>,
-    T::Simd: Add<Output = T::Simd> + Sum<T>,
+    T::Simd: Sum<T> + Add<Output = T::Simd>,
 {
-    let (head, aligned_values) = split_by_alignment::<T::Simd, _>(values);
-
-    let mut chunks = aligned_values.chunks_exact(T::Simd::LANES);
-
     // Safety:
-    // we just made sure that we work on a slice af data aligned to T::Simd
-    let sum = chunks.by_ref().fold(T::Simd::default(), |acc, chunk| {
-        acc + unsafe { T::Simd::from_chunk_aligned_unchecked(chunk) }
-    });
+    // T::Simd is the vector type T and the alignment is similar to aligning to [T; alignment]
+    // the alignment of T::Simd ensures that it fits T.
+    let (head, simd_vals, tail) = unsafe { values.align_to::<T::Simd>() };
 
-    let remainder = T::Simd::from_incomplete_chunk(chunks.remainder(), T::default());
-    let reduced = sum + remainder;
+    let mut reduced = T::Simd::from_incomplete_chunk(&[], T::default());
+    for chunk in simd_vals {
+        reduced = reduced + chunk.clone()
+    }
 
-    reduced.simd_sum() + head.iter().copied().sum()
+    reduced.simd_sum() + head.iter().copied().sum() + tail.iter().copied().sum()
 }
 
 /// # Panics
@@ -108,7 +89,7 @@ where
 /// Returns the sum of values in the array.
 ///
 /// Returns `None` if the array is empty or only contains null values.
-pub fn sum_primitive<T>(array: &PrimitiveArray<T>) -> Option<T>
+pub fn sum_primitive<'a, T>(array: &'a PrimitiveArray<T>) -> Option<T>
 where
     T: NativeType + Simd + Add<Output = T> + std::iter::Sum<T>,
     T::Simd: Add<Output = T::Simd> + Sum<T>,
--- a/src/compute/aggregate/sum.rs
+++ b/src/compute/aggregate/sum.rs
@@ -21,7 +21,7 @@ pub trait Sum<T> {
 
 #[multiversion]
 #[clone(target = "x86_64+avx")]
-fn nonnull_sum<'a, T>(values: &'a [T]) -> T
+fn nonnull_sum<T>(values: &[T]) -> T
 where
     T: NativeType + Simd + Add<Output = T> + std::iter::Sum<T>,
     T::Simd: Sum<T> + Add<Output = T::Simd>,
@@ -89,7 +89,7 @@ where
 /// Returns the sum of values in the array.
 ///
 /// Returns `None` if the array is empty or only contains null values.
-pub fn sum_primitive<'a, T>(array: &'a PrimitiveArray<T>) -> Option<T>
+pub fn sum_primitive<T>(array: &PrimitiveArray<T>) -> Option<T>
 where
     T: NativeType + Simd + Add<Output = T> + std::iter::Sum<T>,
     T::Simd: Add<Output = T::Simd> + Sum<T>,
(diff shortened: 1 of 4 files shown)


src/compute/aggregate/sum.rs
```diff
@@ -19,23 +19,43 @@ pub trait Sum<T> {
     fn simd_sum(self) -> T;
 }
 
+fn split_by_alignment<U, T>(values: &[T]) -> (&[T], &[T]) {
```
ritchie46, line 22: I need to add a check for the case where `values.len() < align_of<U>()`

src/compute/aggregate/sum.rs
```diff
@@ -19,23 +19,43 @@ pub trait Sum<T> {
...
-    let mut chunks = values.chunks_exact(T::Simd::LANES);
+    let (head, aligned_values) = split_by_alignment::<T::Simd, _>(values);
+
+    let mut chunks = aligned_values.chunks_exact(T::Simd::LANES);
 
+    // Safety:
+    // we just made sure that we work on a slice af data aligned to T::Simd
     let sum = chunks.by_ref().fold(T::Simd::default(), |acc, chunk| {
-        acc + T::Simd::from_chunk(chunk)
+        acc + unsafe { T::Simd::from_chunk_aligned_unchecked(chunk) }
```
jorgecarleitao, line 52: we may not need the `unchecked` here (the compiler may figure it out; to test xD)

src/types/simd/packed.rs
```diff
@@ -23,6 +23,11 @@ macro_rules! simd {
                 <$name>::from_slice_unaligned(v)
             }
 
+            #[inline]
+            unsafe fn from_chunk_aligned_unchecked(v: &[$type]) -> Self {
```
jorgecarleitao, line 27: it may be worth trying `from_chunk_aligned` first.

src/compute/aggregate/sum.rs
jorgecarleitao, line 22: I would expect this method to have signature `fn split_by_alignment<U, T>(values: &[T]) -> (&[T], &[T], &[T])` to also take care of the tail. An option here is to use [align_to](https://doc.rust-lang.org/std/primitive.slice.html#method.align_to), which may lead to not needing `_unchecked`

jorgecarleitao (commented):
Interesting. Left some ideas.

src/compute/aggregate/sum.rs
ritchie46: I like that one, because then we don't need the extra trait method anymore. As this leaves the middle chunk already of the `Simd` type.

ritchie46:
An update on the benchmarks:

```rust
sum 2^10 f64            time:   [151.72 ns 152.40 ns 153.21 ns]                         
                        change: [-3.5650% -3.1521% -2.7679%] (p = 0.00 < 0.05)
                        Performance has improved.

sum 2^10 i64            time:   [113.88 ns 113.91 ns 113.94 ns]                         
                        change: [+4.6575% +4.7503% +4.8275%] (p = 0.00 < 0.05)
                        Performance has regressed.
  2 (2.00%) high mild
  6 (6.00%) high severe

sum 2^12 f64            time:   [526.64 ns 527.46 ns 528.31 ns]                          
                        change: [-2.1437% -1.9416% -1.7502%] (p = 0.00 < 0.05)
                        Performance has improved.
  5 (5.00%) high mild
  2 (2.00%) high severe

sum 2^12 i64            time:   [348.69 ns 348.94 ns 349.20 ns]                         
                        change: [-22.641% -21.575% -20.855%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild

sum 2^14 f64            time:   [1.9927 us 1.9929 us 1.9930 us]                          
                        change: [-3.1503% -2.7050% -2.3456%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  3 (3.00%) high mild
  2 (2.00%) high severe

sum 2^14 i64            time:   [1.7309 us 1.7317 us 1.7325 us]                          
                        change: [-8.9944% -8.9284% -8.8592%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild

sum 2^16 f64            time:   [8.3187 us 8.3202 us 8.3222 us]                          
                        change: [-5.6838% -5.3678% -5.0664%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 15 outliers among 100 measurements (15.00%)
  6 (6.00%) low severe
  3 (3.00%) high mild
  6 (6.00%) high severe

sum 2^16 i64            time:   [9.2400 us 9.2499 us 9.2608 us]                          
                        change: [-7.9995% -7.8217% -7.6153%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 16 outliers among 100 measurements (16.00%)
  5 (5.00%) low severe
  2 (2.00%) low mild
  3 (3.00%) high mild
  6 (6.00%) high severe

sum 2^18 f64            time:   [35.474 us 35.487 us 35.504 us]                          
                        change: [-9.2037% -9.1425% -9.0746%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  6 (6.00%) high mild
  4 (4.00%) high severe

sum 2^18 i64            time:   [35.166 us 35.174 us 35.182 us]                          
                        change: [-10.221% -10.140% -10.061%] (p = 0.00 < 0.05)
                        Performance has improved.
  7 (7.00%) high mild
  1 (1.00%) high severe

sum 2^20 f64            time:   [188.31 us 188.55 us 188.80 us]                         
                        change: [-2.2460% -1.7823% -1.2743%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  2 (2.00%) low mild
  5 (5.00%) high mild
  3 (3.00%) high severe

sum 2^20 i64            time:   [207.85 us 209.90 us 212.52 us]                         
                        change: [-8.0797% -6.7231% -5.2968%] (p = 0.00 < 0.05)
                        Performance has improved.
```

It seems to be consistent ~5-10% improvement.

src/compute/aggregate/sum.rs
```diff
@@ -23,19 +23,20 @@ pub trait Sum<T> {
...
-    let remainder = T::Simd::from_incomplete_chunk(chunks.remainder(), T::default());
-    let reduced = sum + remainder;
+    // Safety:
+    // T::Simd is the vector type T and the alignment is similar to aligning to [T; alignment]
+    // the alignment of T::Simd ensures that it fits T.
+    let (head, simd_vals, tail) = unsafe { values.align_to::<T::Simd>() };
+
+    let mut reduced = T::Simd::from_incomplete_chunk(&[], T::default());
+    for chunk in simd_vals {
+        reduced = reduced + chunk.clone()
```
jorgecarleitao, line 36: why is this clone necessary? `chunks` should all be `Copy` and thus `*chunk` should be enough here.

jorgecarleitao (commented):
Looks good. Left a small comment that imo further simplifies the code.

src/compute/aggregate/sum.rs
ritchie46: That is indeed the case for the `packed_simd` vector lane, but the core version does not implement copy. :/

jorgecarleitao: We can implement it, right?  =)

```rust
#[derive(Copy, Clone)]
pub struct $name(pub [$type; $lanes]);
```

and

```
pub trait NativeSimd: Default + Copy
```

src/compute/aggregate/sum.rs
```diff
@@ -23,19 +23,20 @@ pub trait Sum<T> {
...
-    let mut chunks = values.chunks_exact(T::Simd::LANES);
-
-    let sum = chunks.by_ref().fold(T::Simd::default(), |acc, chunk| {
-        acc + T::Simd::from_chunk(chunk)
-    });
-
-    let remainder = T::Simd::from_incomplete_chunk(chunks.remainder(), T::default());
-    let reduced = sum + remainder;
+    // Safety:
+    // T::Simd is the vector type T and the alignment is similar to aligning to [T; alignment]
```
jorgecarleitao, line 30: I think that we need to make the trait `NativeSimd` `unsafe` and describe the invariant that it must uphold (that alignments match).

I would also move this to a (`safe`) inlined function in the trait `Simd`, so that the `unsafe` is close to where the contract is defined (as it may be used in other instances)

src/compute/aggregate/sum.rs
ritchie46: Oh.. I did not realize we create the native types. I thought the were from `core::arch`. My bad!

src/compute/aggregate/sum.rs
ritchie46: Good idea.

PR #450, opened 2021-09-25
Added `MutableArray::as_box`

### LINKED ISSUE #447: Add `as_boxed` method to `MutableArray`
`MutableArray` has `as_arc`, but not `as_boxed` (or `into_boxed`? not sure which is more appropriate) right now, although `GrowableArray` does - is this something you would consider adding? I have a use case which requires it to implement `Deserialize` for a `Box<dyn Array>` wrapper 🙂 

Happy to send in a PR if this sounds good. Thanks!

sd2k:
This basically just changes the impls of `as_arc` to `as_box` and uses
the `From<Box<T>> for Arc<T>` impl to provide a default implementation
for `as_arc`.

Closes #447.

### DIFF
--- a/src/array/mod.rs
+++ b/src/array/mod.rs
@@ -113,8 +113,16 @@ pub trait MutableArray: std::fmt::Debug {
     /// The optional validity of the array.
     fn validity(&self) -> &Option<MutableBitmap>;
 
-    /// Convert itself to an (immutable) [`Array`].
-    fn as_arc(&mut self) -> Arc<dyn Array>;
+    /// Convert itself to an (immutable) ['Array'].
+    fn as_box(&mut self) -> Box<dyn Array>;
+
+    /// Convert itself to an (immutable) atomically reference counted [`Array`].
+    // This provided implementation has an extra allocation as it first
+    // boxes `self`, then converts the box into an `Arc`. Implementors may wish
+    // to avoid an allocation by skipping the box completely.
+    fn as_arc(&mut self) -> Arc<dyn Array> {
+        self.as_box().into()
+    }
 
     /// Convert to `Any`, to enable dynamic casting.
     fn as_any(&self) -> &dyn Any;
--- a/src/array/binary/mutable.rs
+++ b/src/array/binary/mutable.rs
@@ -146,6 +146,15 @@ impl<O: Offset> MutableArray for MutableBinaryArray<O> {
         &self.validity
     }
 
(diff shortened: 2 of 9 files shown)


sd2k:
Just wondering, is there a reason that the `as_arc` method wasn't a (consuming) `into_arc` method?

src/array/mod.rs
```diff
@@ -113,8 +113,13 @@ pub trait MutableArray: std::fmt::Debug {
...
     fn validity(&self) -> &Option<MutableBitmap>;
 
-    /// Convert itself to an (immutable) [`Array`].
-    fn as_arc(&mut self) -> Arc<dyn Array>;
+    /// Convert itself to an (immutable) ['Array'].
+    fn as_box(&mut self) -> Box<dyn Array>;
+
+    /// Convert itself to an (immutable) atomically reference counted [`Array`].
+    fn as_arc(&mut self) -> Arc<dyn Array> {
+        self.as_box().into()
```
jorgecarleitao, line 121: Just a note: this has an extra allocation (`Box` and `Arc` are incompatible), I do not know if rust optimizes this if we inline `as_box` or not. imo this is not a concern, but still a good to know.

jorgecarleitao (approved):
Thank you for this PR 🙇  Looks great and ready to go.

> Just wondering, is there a reason that the `as_arc` method wasn't a (consuming) `into_arc` method?

Great question :)

For a trait to be a trait object, it cannot be `Sized`. Any method that consumes `self` (and not `&mut self`) makes the trait automatically `Sized` (because `fn a(self)`  requires knowing the size of `Self` to make it to a registry). So, we either have `into_arc(self)` and not a trait object, or a trait object and `as_arc(&mut self)`. Since we really need the trait object here, we make it a trait object (and use `as_arc`).

But yes, you are right that `into_arc` is what makes sense here.

sd2k:
Ah of course, I'd forgotten we needed object safety 😅 Kind of fundamental hehe. Thanks!

src/array/mod.rs
sd2k: Looks like possibly not: https://rust.godbolt.org/z/4TrEqGjaf. Maybe it's worth quickly going through and adding explicit impls?

jorgecarleitao: As you prefer :)

sd2k: I've added explicit impls for the existing arrays and added a comment to the provided method in the trait.

PR #453, opened 2021-09-25
More granular dtype serialization options for csv writer.

ritchie46:
~This changes csv output for the `Date64` datatype to also include the time information. So that a round trip of `Date64 -> csv -> Date64` would not lose information.~


ritchie46:
To stay compliant with arrow-spec, I kept the default the  same. I just give user more granular control in how to serialize different data types. Which makes sense as uses might want to serialize a time64 with more precision than a time32.

PR #454, opened 2021-09-26
cargo fmt

ritchie46:
A `$ cargo fmt` run.


jorgecarleitao:
Maybe we should check this in the CI?

ritchie46:
> Maybe we should check this in the CI?

Yep. :)

PR #460, opened 2021-09-28
Added property testing

jorgecarleitao:
This PR adds a couple of tests using [`proptest`](https://github.com/altsysrq/proptest), a testing framework to test code invariants. See [here](https://github.com/altsysrq/proptest) or [here](https://hypothesis.works/) for an introduction to the subject.

This framework is [being used](https://github.com/rust-lang/portable-simd/blob/master/crates/core_simd/Cargo.toml#L22) by the portable SIMD group, and in my opinion it makes a lot of sense here also, since we can rigorously declare and test the properties of our operations.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,8 @@ lexical-core = { version = "0.8", optional = true }
 # We need to Hash values before sending them to an hasher. This
 # crate provides HashMap that assumes pre-hashed values.
 hash_hasher = "^2.0.3"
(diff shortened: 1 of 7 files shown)


jorgecarleitao:
Failing MIRI is related to `proptest` using `clib::getcwd`, which is not supported in isolation mode. Question [posed upstream](https://github.com/AltSysrq/proptest/issues/253).

PR #467, opened 2021-09-29
expose `shrink_to_fit` to mutable arrays

ritchie46:
This PR proposes to expose `shrink_to_fit` on the mutable arrays. The `MutableBuffer` already implements this, but its impossible to call with mutable array API as the internal buffers are private.

The same could be done for `MutableBitmap`.

### DIFF
--- a/src/array/binary/mutable.rs
+++ b/src/array/binary/mutable.rs
@@ -138,6 +138,7 @@ impl<O: Offset> MutableBinaryArray<O> {
     /// Shrinks the capacity of the [`MutableBinaryArray`] to fit its current length.
     pub fn shrink_to_fit(&mut self) {
         self.values.shrink_to_fit();
+        self.offsets.shrink_to_fit();
         if let Some(validity) = &mut self.validity {
             validity.shrink_to_fit()
         }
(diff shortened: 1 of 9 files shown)


jorgecarleitao:
Yeap, makes a lot of sense.

jorgecarleitao:
I can take care of the `MutableBitmap` and we can then rebase on top of that.

ritchie46:
> I can take care of the MutableBitmap and we can then rebase on top of that.

Great, once #468 is in main, I will do the rebase and add the `MutableBitmap` shrinking as well.

ritchie46:
Good to go!

src/array/binary/mutable.rs
```diff
@@ -135,6 +135,13 @@ impl<O: Offset> MutableBinaryArray<O> {
         let a: BinaryArray<O> = self.into();
         Arc::new(a)
     }
+    /// Shrinks the capacity of the [`MutableBinaryArray`] to fit its current length.
+    pub fn shrink_to_fit(&mut self) {
+        self.values.shrink_to_fit();
```
jorgecarleitao, line 140:
```suggestion
        self.values.shrink_to_fit();
        self.offsets.shrink_to_fit();
```

jorgecarleitao (commented):
Looks great! Thanks a lot, @ritchie46 ! We are missing one buffer xD

PR #471, opened 2021-09-30
Made writing stream to parquet require a non-static lifetime

GrandChaman:
Hello !

Would it be possible to change the lifetime of the iterators to something a little less restrictive when writing to a `parquet` file ?


jorgecarleitao (approved):
Most definitely! Not sure why they were marked as `'static` in the first place, but it is definitely a bug.

Also, very cool picture =)

Thanks for the patch!

PR #478, opened 2021-09-30
Added `extend`/`extend_unchecked` for `MutableBooleanArray`

VasanthakumarV:
This PR adds `TrustedLen`-based extend and extend_unchecked methods for `MutableBooleanArray`.

I have mostly followed the implementation for [`MutablePrimitiveArray`](https://github.com/jorgecarleitao/arrow2/blob/main/src/array/primitive/mutable.rs).

### DIFF
--- a/src/array/boolean/mutable.rs
+++ b/src/array/boolean/mutable.rs
@@ -111,6 +111,7 @@ impl MutableBooleanArray {
     where
         I: TrustedLen<Item = bool>,
     {
+        // Safety: `I` is `TrustedLen`
         unsafe { self.extend_trusted_len_values_unchecked(iterator) }
     }
 
@@ -141,6 +142,7 @@ impl MutableBooleanArray {
         P: std::borrow::Borrow<bool>,
         I: TrustedLen<Item = Option<P>>,
     {
+        // Safety: `I` is `TrustedLen`
         unsafe { self.extend_trusted_len_unchecked(iterator) }
     }
 
@@ -254,6 +256,7 @@ impl MutableBooleanArray {
         P: std::borrow::Borrow<bool>,
         I: TrustedLen<Item = Option<P>>,
     {
+        // Safety: `I` is `TrustedLen`
         unsafe { Self::from_trusted_len_iter_unchecked(iterator) }
     }
 
@@ -287,6 +290,7 @@ impl MutableBooleanArray {
         P: std::borrow::Borrow<bool>,
         I: TrustedLen<Item = std::result::Result<Option<P>, E>>,
     {
+        // Safety: `I` is `TrustedLen`
         unsafe { Self::try_from_trusted_len_iter_unchecked(iterator) }
     }
 }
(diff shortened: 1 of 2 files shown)


src/array/boolean/mutable.rs
```diff
@@ -104,6 +104,69 @@ impl MutableBooleanArray {
...
     }
 
+    /// Extends the [`MutableBooleanArray`] from an iterator of values of trusted len.
+    /// This differs from `extend_trusted_len` which accepts in iterator of optional values.
+    #[inline]
+    pub fn extend_trusted_len_values<I>(&mut self, iterator: I)
+    where
+        I: TrustedLen<Item = bool>,
+    {
+        unsafe { self.extend_trusted_len_values_unchecked(iterator) }
```
jorgecarleitao, line 114:
```suggestion
        /// Safety: `I` is `TrustedLen`
        unsafe { self.extend_trusted_len_values_unchecked(iterator) }
```

src/array/boolean/mutable.rs
```diff
@@ -240,33 +297,53 @@ impl MutableBooleanArray {
...
         } else {
-            validity.push(false);
-            false
+            validity.push_unchecked(false);
+            bool::default()
         };
-        values.push(item);
+        values.push_unchecked(item);
     }
-    assert_eq!(
```
jorgecarleitao, line 264: Could we keep the `debug_assert_eq!` here?

tests/it/array/boolean/mutable.rs
```diff
@@ -64,3 +64,35 @@ fn reserve() {
...
+    assert_eq!(a.validity(), None);
+    assert_eq!(a.values(), &MutableBitmap::from([true, true, false]));
+
+    let mut a = MutableBooleanArray::new();
+    a.push(None);
+    a.extend_trusted_len_values(vec![true, false].into_iter());
+    assert_eq!(
+        a.validity(),
+        Some(&MutableBitmap::from([false, true, true]))
+    );
```
jorgecarleitao, line 97:
```suggestion
    assert_eq!(
        a.validity(),
        Some(&MutableBitmap::from([false, true, true]))
    );
    assert_eq!(
        a.values(),
        &MutableBitmap::from([false, true, false])
    );
```

jorgecarleitao (approved):
As usual, beautiful. Thank you very much!

Left some suggestions (the `Safety` should be applied to every usage of `unsafe`), but otherwise ready to go.

tests/it/array/boolean/mutable.rs
VasanthakumarV: Thanks for pointing this out, the missing assertion has been added

src/array/boolean/mutable.rs
VasanthakumarV: @jorgecarleitao I have added `debug_assert_eq`, but I had to create a variable to store the length value before expansion and use it later in the assert statement

src/array/boolean/mutable.rs
VasanthakumarV: The Safety comment was added to every place inside the file where we use unsafe blocks, but `///` wasn't possible

PR #482, opened 2021-10-01
Use itertools to process the `min_value` and `max_value` in statistics

GrandChaman:
Hello!

Currently, when processing the statistics, the array is scanned twice to get the `min` and `max` value.

The [Itertools::minmax_by](https://docs.rs/itertools/0.10.1/itertools/trait.Itertools.html#method.minmax_by) can do the same and faster, needing to go through `1.5 * n` elements, rather than `2 * n` elements, for an array of `n` elements.

### DIFF
--- a/src/io/parquet/write/primitive/basic.rs
+++ b/src/io/parquet/write/primitive/basic.rs
@@ -14,6 +14,7 @@ use crate::{
     io::parquet::read::is_type_nullable,
     types::NativeType as ArrowNativeType,
 };
+use itertools::{Itertools, MinMaxResult};
 
 pub(crate) fn encode_plain<T, R>(array: &PrimitiveArray<T>, is_optional: bool, buffer: &mut Vec<u8>)
 where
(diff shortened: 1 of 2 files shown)


ritchie46:
Thanks for your PR @GrandChaman.

I think we could do this optimization without the need of adding a new dependency. We could write a function that tracks both in a single pass for instance.

Dandandan:
Also, are you sure it's faster or has a big impact?

As we are using primitive arrays, wouldn't it make sense here to use the optimized compute kernels here instead?
This has further optimizations / vectorization compared to using `array.iter().flatten()` etc.

src/io/parquet/write/primitive/basic.rs
```diff
@@ -98,26 +99,27 @@ where
     R: NativeType,
     T: num_traits::AsPrimitive<R>,
 {
+    let minmax = array
```
Dandandan, line 102: Why not use the min/max from compute here?

Dandandan: As they are primitive you could use `min_primitive`, `max_primitive` here.
I think the downside is having to enable the compute feature for writing parquet. Is that the reason it's not using the compute kernels @jorgecarleitao ?

GrandChaman: Indeed, maybe using `min_primitve` and `max_primitive` could have some advantages, to leverage SIMD for instance.

ritchie46: > I think the downside is having to enable the compute feature for writing parquet. 

We could use a `[cfg(feature="compute")]` tag to use the kernels if compute is enabled and use iterators/ loops otherwise.

GrandChaman:
@ritchie46 The dependencies was already in `arrow2`, although not with the feature `io_parquet`. I agree it'd be best not to add new dependencies, I just wanted to test quickly without re-implementing what `itertools` had to offer.

@Dandandan  From my testing, it is faster. But I agree maybe the compute kernel are better suited for this task.

src/io/parquet/write/primitive/basic.rs
Dandandan: That seems a good trade off to me.

GrandChaman: Although those compute kernels wouldn't be compatible with type that don't implement the `Simd` trait, `i128` for instance.

jorgecarleitao:
Thanks a lot for this! This is a good point.

@Dandandan , exactly: I did not want to drag that dependency just to read/write parquet. I like @ritchie46 's idea: if we have the kernel already in the features, use it; else, use a simple iteration / our own implementation of minmax.

An alternative that would also be fine is to see if itertools offers features that we can pick via `default-features = false, features = [...]` to reduce the dependency to its minimum (itertools is pretty large).

PR #488, opened 2021-10-02
Improved multiversion of hash.

jorgecarleitao:
I noticed that multiversion recommends one `clone` per arch/instruction set. @Dandandan , do you remember why we used all the sets in a single clone? My feeling is that it will create a single clone iff the whole set is hit, instead of one clone per set, but I may be mistaken here.

### DIFF
--- a/src/compute/hash.rs
+++ b/src/compute/hash.rs
@@ -20,7 +20,8 @@ use crate::{
 use super::arity::unary;
 
 #[multiversion]
-#[clone(target = "x86_64+aes+sse3+ssse3+avx+avx2")]
+#[clone(target = "[x86|x86_64]+avx+avx2")]
+#[clone(target = "x86+sse")]
 /// Element-wise hash of a [`PrimitiveArray`]. Validity is preserved.
 pub fn hash_primitive<T: NativeType + Hash>(array: &PrimitiveArray<T>) -> PrimitiveArray<u64> {
     let state = new_state!();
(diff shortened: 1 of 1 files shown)


Dandandan:
> I noticed that multiversion recommends one `clone` per arch/instruction set. @Dandandan , do you remember why we used all the sets in a single clone? My feeling is that it will create a single clone iff the whole set is hit, instead of one clone per set, but I may be mistaken here.
> 
> 

As far as I understand it, a version of the instruction combinations will be compiled for each clone and checked against at runtime. As such, it doesn't make sense to only compile with a subset of instruction sets, as most modern x64 CPUs contain all sets until e.g. `avx2` (from Haswell, 2014 onwards), but avx512 is quite rare.

Also some instruction setsike `sse`, `sse2` will be compiled in the fallback / default option anyway, we don't need to add an extra clone for them.

I think what we need to do for checking whether it makes sense to add clones:

* Benchmark performance improvement given the combination of instructions.
* Check how often this instruction set combination occurs in CPUs we are targeting

src/compute/hash.rs
```diff
@@ -29,7 +30,8 @@ pub fn hash_primitive<T: NativeType + Hash>(array: &PrimitiveArray<T>) -> Primit
 }
 
 #[multiversion]
-#[clone(target = "x86_64+aes+sse3+ssse3+avx+avx2")]
+#[clone(target = "[x86|x86_64]+avx+avx2")]
+#[clone(target = "x86+sse")]
```
Dandandan, line 34: I'm not sure if it makes sense, x86 is pretty old by now, so not sure it makes sense to add it here if we don't know the performance impact.

src/compute/hash.rs
```diff
@@ -29,7 +30,8 @@ pub fn hash_primitive<T: NativeType + Hash>(array: &PrimitiveArray<T>) -> Primit
 }
 
 #[multiversion]
-#[clone(target = "x86_64+aes+sse3+ssse3+avx+avx2")]
+#[clone(target = "[x86|x86_64]+avx+avx2")]
```
Dandandan, line 33: I also needed the features `ssse3` `sse3` and `aes`  at least (as those are used specifically by `aHash`

jorgecarleitao:
Thanks a lot for your insights 💯

PR #496, opened 2021-10-08
Add Wrapping version ops for primitiveArray

yjhmelody:


src/compute/arithmetics/basic/add.rs
```diff
@@ -42,6 +43,34 @@ where
     binary(lhs, rhs, lhs.data_type().clone(), |a, b| a + b)
 }
 
+/// Wrapping addition of two primitive arrays.
+/// It do nothing if the result overflows.
```
jorgecarleitao, line 47:
```suggestion
/// It wraps around at the boundary of the type if the result overflows.
```

(following [the language used in std](https://doc.rust-lang.org/std/primitive.u32.html#method.wrapping_add)

src/compute/arithmetics/basic/add.rs
```diff
@@ -42,6 +43,34 @@ where
     binary(lhs, rhs, lhs.data_type().clone(), |a, b| a + b)
 }
 
+/// Wrapping addition of two primitive arrays.
```
jorgecarleitao, line 46:
```suggestion
/// Wrapping addition of two `PrimitiveArray`s.
```

src/compute/arithmetics/basic/add.rs
```diff
@@ -138,6 +167,18 @@ where
     }
 }
 
+// Implementation of ArrayWrappingAdd trait for PrimitiveArrays
```
jorgecarleitao, line 170:
```suggestion
```

(it is clear from the line below ^_^)

src/compute/arithmetics/basic/add.rs
```diff
@@ -195,6 +236,26 @@ where
     unary(lhs, |a| a + rhs, lhs.data_type().clone())
 }
 
+/// Wrapping addition of a scalar T to a primitive array of type T.
```
jorgecarleitao, line 239:
```suggestion
/// Wrapping addition of a scalar T to a [`PrimitiveArray`] of type T.
```

src/compute/arithmetics/basic/mul.rs
```diff
@@ -42,6 +43,34 @@ where
     binary(lhs, rhs, lhs.data_type().clone(), |a, b| a * b)
 }
 
+/// Wrapping multiplication of two primitive arrays.
+/// It do nothing if the result overflows.
```
jorgecarleitao, line 47: same here?

Issue #500, opened 2021-10-08
write to json does not take validity into account when writing `StructArray`


jorgecarleitao:
Check `struct_array_to_jsonmap_array`.

jorgecarleitao:
Closed by #511 .

PR #511, opened 2021-10-08
[WIP] Validity taken into account when writing `StructArray` to json

VasanthakumarV:
Linked to #500

### DIFF
--- a/src/io/json/write/serialize.rs
+++ b/src/io/json/write/serialize.rs
@@ -18,6 +18,8 @@
 use serde_json::map::Map;
 use serde_json::{Number, Value};
 
+use crate::bitmap::utils::zip_validity;
+use crate::bitmap::Bitmap;
 use crate::{array::*, datatypes::*, record_batch::RecordBatch, types::NativeType};
 
 trait JsonSerializable {
@@ -117,7 +119,17 @@ fn struct_array_to_jsonmap_array(array: &StructArray, row_count: usize) -> Vec<M
         .take(row_count)
         .collect::<Vec<Map<String, Value>>>();
 
-    // todo: use validity...
+    // Combine the two incoming validity `Bitmaps` using `and` logic
+    let combine_validity = |col_validity: Option<&Bitmap>, array_validity: Option<&Bitmap>| match (
+        col_validity,
+        array_validity,
+    ) {
+        (Some(cv), Some(av)) => Some(cv & av),
+        (Some(cv), None) => Some(cv.clone()),
+        (None, Some(av)) => Some(av.clone()),
+        (None, None) => None,
+    };
+
     array
         .values()
         .iter()
@@ -127,6 +139,8 @@ fn struct_array_to_jsonmap_array(array: &StructArray, row_count: usize) -> Vec<M
                 &mut inner_objs,
                 row_count,
                 struct_col.as_ref(),
+                // For each column we combine its validity with the `StructArray`'s validity
+                combine_validity(struct_col.validity(), array.validity()).as_ref(),
                 fields[j].name(),
             );
         });
@@ -234,6 +248,7 @@ fn set_column_for_json_rows(
     rows: &mut [Map<String, Value>],
     row_count: usize,
     array: &dyn Array,
+    validity: Option<&Bitmap>,
     col_name: &str,
 ) {
     match array.data_type() {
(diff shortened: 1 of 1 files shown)


tests/it/io/json/write.rs
```diff
@@ -38,6 +38,55 @@ fn write_simple_rows() {
...
+    }
+
+    assert_eq!(
+        String::from_utf8(buf).unwrap(),
+        r#"{"c1":{"c11":1,"c12":{"c121":null,"c122":null}},"c2":"a"}
+{"c1":{"c11":null,"c12":{"c121":"f","c122":null}},"c2":"b"}
+{"c1":{"c11":null,"c12":{"c121":null,"c122":null}},"c2":"c"}
+"#
+    );
+}
```
VasanthakumarV, line 88: @jorgecarleitao Let me know if the output of this test case showcases the desired outcome when serializing using validity

tests/it/io/json/write.rs
```diff
@@ -38,6 +38,55 @@ fn write_simple_rows() {
...
+    {
+        let mut writer = LineDelimitedWriter::new(&mut buf);
+        writer.write_batches(&[batch]).unwrap();
+    }
+
+    assert_eq!(
+        String::from_utf8(buf).unwrap(),
+        r#"{"c1":{"c11":1,"c12":{"c121":null,"c122":null}},"c2":"a"}
+{"c1":{"c11":null,"c12":{"c121":"f","c122":null}},"c2":"b"}
+{"c1":{"c11":null,"c12":{"c121":null,"c122":null}},"c2":"c"}
```
jorgecarleitao, line 85: I would expect `{"c1":null,"c2":"c"}` in this line?

jorgecarleitao (commented):
Interesting approach. I would imagine that we do not need to combine the validities for this, and instead just write a `null` to the whole field, but maybe you already went through that exercise and deemed not feasible?

If yes, could you clarify this on the code with a comment? If not, I think it is worth the shot, since it would make the file smaller

src/io/json/write/serialize.rs
```diff
@@ -299,6 +298,17 @@ fn set_column_for_json_rows(
                 .for_each(|(row, obj)| {
```
jorgecarleitao, line 298: this `for_each` will hit the iterator, causing a complete serialization even if the item is `null` from the struct's point of view.

what about something like `if let Some(validity) = array.validity()`, we `zip` the validity in this iteration directly (and if not, we use the code we already have here). I.e. essentially try to "blend" your changes to the existing code so that we avoid running the loop twice.

tests/it/io/json/write.rs
```diff
@@ -231,9 +280,9 @@ fn write_list_of_struct() {
 
     assert_eq!(
         String::from_utf8(buf).unwrap(),
-        r#"{"c1":[{"c11":1,"c12":{"c121":"e"}},{"c11":null,"c12":{"c121":"f"}}],"c2":1}
+        r#"{"c1":[{"c11":1,"c12":null},{"c11":null,"c12":{"c121":"f"}}],"c2":1}
 {"c1":null,"c2":2}
-{"c1":[{"c11":5,"c12":{"c121":"g"}}],"c2":3}
+{"c1":[null],"c2":3}
```
VasanthakumarV, line 285: @jorgecarleitao Let me know if this test-case (list of structs) behavior is what was expected.

src/io/json/write/serialize.rs
```diff
@@ -295,9 +297,15 @@ fn set_column_for_json_rows(
...
-                .zip(inner_objs.into_iter())
+                .zip(zip_validity(
+                    inner_objs.into_iter(),
+                    array.validity().map(|v| v.iter()),
+                ))
                 .for_each(|(row, obj)| {
-                    row.insert(col_name.to_string(), Value::Object(obj));
+                    row.insert(
+                        col_name.to_string(),
+                        obj.map(|o| Value::Object(o)).unwrap_or(Value::Null),
```
jorgecarleitao, line 307:
```suggestion
                        obj.map(Value::Object).unwrap_or(Value::Null),
```

src/io/json/write/serialize.rs
```diff
@@ -202,7 +202,9 @@ fn write_array(array: &dyn Array) -> Value {
                 array.as_any().downcast_ref::<StructArray>().unwrap(),
                 array.len(),
             );
-            jsonmaps.into_iter().map(Value::Object).collect()
+            zip_validity(jsonmaps.into_iter(), array.validity().map(|v| v.iter()))
+                .map(|m| m.map(|o| Value::Object(o)).unwrap_or(Value::Null))
```
jorgecarleitao, line 206:
```suggestion
                .map(|m| m.map(Value::Object).unwrap_or(Value::Null))
```

tests/it/io/json/write.rs
jorgecarleitao: Yap, makes a lot of sense to me: the item is null and thus the whole thing is just `null`.

Issue #497, opened 2021-10-08
Add support for `Codec::Deflate` in avro reading


jorgecarleitao:
So that we can read compressed avro.

PR #512, opened 2021-10-09
Add support to read compressed avro

xudong963:
Close https://github.com/jorgecarleitao/arrow2/issues/497


Igosuki:
Hello peeps, just wanted to say that while not explicit in the avro spec, the actualy default compression codec for avro in the wild is snappy, because it's the default compression format of spark-avro. It is not a new concept to this crate since it's used in the parquet testing.
Here it is in action in avro-rs https://github.com/flavray/avro-rs/blob/941000b1e3c71a89010999d7aa0c4955e3672f5a/src/codec.rs#L45
Let me know if I can provide support !

jorgecarleitao:
Thanks @Igosuki, That is very useful to know!

imo once we crack the "efficient decompression logic" we can easily add snappy, since almost all compressions have the same APIs. I would expect to support them all as-well.

I would just try to target to use the same dependencies as `parquet2` has for compression so that if people activate the parquet and avro feature, just load 1 set of dependencies.

potter420:
Hi, actually the snappy seem much simpler to optimize. In the decompress code from `avro-rs` code above, we know decompress size before hand via
```rust
let decompressed_size = snap::decompress_len(&stream[..stream.len() - 4])
                    .map_err(Error::GetSnappyDecompressLen)?;
```
So when we have the size, we can just simply ```decompress.clear(); decompress.reserve(decompressed_size)```
The deflate don't seem to have an easy way to estimate decompression size from compressed data. 

What do you think?

jorgecarleitao:
The flag is the other way around, then xD. We need to tell whether the buffer was swapped, so that consumers can re-use the `decompress` buffer if they want.

jorgecarleitao:
Very sorry for missing the test 👍 I also understood why we use `libflate`, because `avro` is already pulling it to the dependencies. So, I think we just need to make the dependency optional. 

Again sorry for the noise :(

xudong963:
No worry!  I finished it 😄

PR #515, opened 2021-10-09
add dictionary serialization for csv-writer

ritchie46:
This PR adds csv serializations for dictionary encoded arrays. I've only implemented them for typically rust indexes `usize` atm. It would be trivial to add other indexes, but if no-one (don't know how they are used in the wild) uses them it might not be worth the compilation.


src/io/csv/write/serialize.rs
```diff
@@ -265,6 +269,53 @@ pub fn new_serializer<'a>(
...
+        keys.iter(),
+        move |x, buf| {
+            if let Some(x) = x {
+                let i = Index::to_usize(x);
+                if !values.is_null(i) {
+                    // Safety:
+                    // all keys of a dictionary array should be in bounds
+                    // invariant of the struct
+                    // and we just checked bounds via `is_null`
+                    let val = unsafe { values.value_unchecked(i) };
```
jorgecarleitao, line 314: this is currently not checked; we cannot use `unchecked` here atm.

jorgecarleitao (commented):
Thanks a lot, Ritchie!

Almost there imo:

1. need to add the bound checks
2. some tests :P

PR #519, opened 2021-10-10
cargo fmt

ritchie46:
A `$ cargo fmt` run.


xudong963:
Maybe we should add format check in CI? @jorgecarleitao

PR #522, opened 2021-10-10
enhance fmt and clippy

xudong963:
Close https://github.com/jorgecarleitao/arrow2/issues/520


jorgecarleitao:
We should use ´--full´ feature here: that function is used in `io_ipc` (we could also move it there).

PR #528, opened 2021-10-14
Improved performance of writing decimal to `parquet`

### LINKED ISSUE #502: Improve performance of writing decimal to `parquet`
```rust
let mut values = MutableBuffer::<u8>::new(); // todo: this can be estimated
```

which we can replace by a `with_capacity` call with the expected size of the array (which can be deterministically computed).

koneko096:
Close #502


jorgecarleitao (approved):
Thanks a lot for your contribution. It is flawless!

I noticed that you have a actually fixed the unskip as part of the PR 💯 , but merged my fix first, which caused a conflict here :( :( Could you rebase?

Could you revert that so we can merge this?

koneko096:
Ah, sure. In a moment.

PR #532, opened 2021-10-16
Simplified reading parquet

jorgecarleitao:
This PR simplifies the code to read `parquet`, making it a bit more future proof and opening the doors to improve performance in writing by re-using buffers (improvements upstream).

I do not observe differences in performance (vs main) in the following parquet configurations:

* single page vs multiple pages
* compressed vs uncompressed
* different types

It generates flamegraphs that imo are quite optimized:

Screenshot 2021-10-16 at 06 44 55

This corresponds to

```bash
cargo flamegraph --features io_parquet,io_parquet_compression \
    --example parquet_read fixtures/pyarrow3/v1/multi/snappy/benches_1048576.parquet \
    1 0
```

i.e. reading a f64 column from a single row group with 1M rows with a page size of 1Mb (default in pyarrow).

Screenshot 2021-10-16 at 06 53 44

(same but for a utf8 column (column index 2))

The majority of the time is used deserializing the data to arrow, which means that the main gains to have continue to be on that front.

## Backward incompatible

* The API to read parquet now uses `FallibleStreamingIterator` instead of `StreamingIterator` (of `Result<Page>`). As before, we re-export these APIs in `io::parquet::read`.
* The API to write parquet now expects the user to decompress the pages. This is only relevant when not using `RowGroupIterator` (i.e. in parallelizing). This is now enforced by the type system (`DataPage` vs `CompressedDataPage`), so that we do not get it wrong.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,7 +61,9 @@ futures = { version = "0.3", optional = true }
 # for faster hashing
 ahash = { version = "0.7", optional = true }
 
-parquet2 = { version = "0.5.2", optional = true, default_features = false, features = ["stream"] }
+#parquet2 = { version = "0.5.2", optional = true, default_features = false, features = ["stream"] }
+#parquet2 = { git = "https://github.com/jorgecarleitao/parquet2", branch = "reuse_compress", optional = true, default_features = false, features = ["stream"] }
+parquet2 = { path = "../parquet2", optional = true, default_features = false, features = ["stream"] }
 
 avro-rs = { version = "0.13", optional = true, default_features = false }
 
@@ -100,7 +102,7 @@ full = [
     "io_avro",
     "regex",
     "merge_sort",
-    "compute",
+    #"compute",
     # parses timezones used in timestamp conversions
     "chrono-tz"
 ]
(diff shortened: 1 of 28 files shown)


ritchie46:
Any opportunity for `simdutf8` in validation of utf8?

Do you think we could parallize the pages (I believe it were the pages, correct me if I am wrong) decoding similar to what was done in for writing? Maybe accept a closure that takes the pages and returns them decode ones?

Dandandan:
> Any opportunity for `simdutf8` in validation of utf8?

@jorgecarleitao is the flame graph taken with or without the `simdutf8` feature on? I think it's disabled or individual strings are smaller than the `simdutf8` boundary.

jorgecarleitao:
Sorry, what do you mean? There is no `simdutf8` feature: is always on.

jorgecarleitao:
The answer seems to be no:

```rust
fn main() {
    let a = "π";  // valid utf8
    let a = a.as_bytes(); // [207, 128]
    assert!(std::str::from_utf8(&a[..1]).is_ok());
}
```

PR #534, opened 2021-10-16
Added more IPC documentation

HagaiHargil:


src/io/ipc/mod.rs
```diff
@@ -1,4 +1,80 @@
...
+//! Read and write to a file:
+//! ```
+//! use arrow2::io::ipc::{{read::{FileReader, read_file_metadata}}, {write::FileWriter}};
+//! # use std::fs::File;
+//! # use std::sync::Arc;
+//! # use arrow2::datatypes::{Field, Schema, DataType};
+//! # use arrow2::array::Int32Array;
+//! # use arrow2::record_batch::RecordBatch;
+//! // Setup the writer
+//! let path = "example.dat".to_string();
```
jorgecarleitao, line 40:
```suggestion
//! let path = "example.arrow".to_string();
```

src/io/ipc/mod.rs
```diff
@@ -1,4 +1,80 @@
...
+//! let mut file = File::create(&path).unwrap();
```
jorgecarleitao, line 41:
```suggestion
//! let mut file = File::create(&path)?;
```

and add an `fn main() -> Result<()>`?

imo it is important that we (Rust community) illustrate the correct use of `Result` to newcomers, since `.unwrap` is usually pretty brutal for an application difficult to catch.

src/io/ipc/read/stream.rs
```diff
@@ -91,7 +107,8 @@ impl StreamState {
     }
 }
 
-/// Reads the next item
+/// Reads the next item, yielding `None` if the stream is done,
+/// and a `StreamState` otherwise.
```
jorgecarleitao, line 111:
```suggestion
/// and a [`StreamState`] otherwise.
```

src/io/ipc/read/stream.rs
```diff
@@ -191,7 +208,15 @@ pub fn read_next<R: Read>(
     }
 }
 
-/// Arrow Stream reader
+/// Arrow Stream reader.
+///
+/// Once a stream is created the recommended way to interact with it is by
+/// iterating over its data. Each such iteration yields an [`Option<StreamState>`](StreamState),
+/// signaling that the stream might not currently have any data, even though it's still a "live"
+/// stream.
```
jorgecarleitao, line 216:
```suggestion
/// An [`Iterator`] over an Arrow stream that yields a result of [`StreamState`]s.
/// This is the recommended way to read an arrow stream (by iterating over its data).
```

(no need to repeat what `StreamState` state is; the iterator's items are usually referred to as the thing inside the `Option`)

src/io/ipc/read/stream.rs
```diff
@@ -191,7 +208,15 @@ pub fn read_next<R: Read>(
...
+///
+/// For a full usage examples consult [this](https://github.com/jorgecarleitao/arrow2/tree/main/examples/ipc_pyarrow)
+/// example in the main repository.
```
jorgecarleitao, line 219:
```suggestion
/// For a full usage examples consult [this example](https://github.com/jorgecarleitao/arrow2/tree/main/examples/ipc_pyarrow).
```

(no need discuss git repositories here ^_^)

src/io/ipc/mod.rs
HagaiHargil: You are (of course) completely correct. I simply forgot that you can do that in doctests.

HagaiHargil:
I believe that this should be it. There's obviously much more to write about the subject - i.e. I believe that the whole DictionaryBatch support is currently a bit lacking - but for now that's all I can offer :)

PR #557, opened 2021-10-30
Introduced `UnionMode` enum

simonvandel:
I feel like seeing `UnionMode::Sparse` is easier than `true`


jorgecarleitao (approved):
I agree. Looks great. Thanks!

PR #561, opened 2021-10-31
Compute: add partial option into CastOptions

### LINKED ISSUE #569: Made `cast` accept `CastOptions` parameter
Different casts may have different options. To not have to cover every combination as a different function, we should make `cast` accept `CastOptions`.

This is backward incompatible. Migrate:

```rust
let b = wrapping_cast(&array, &DataType::UInt8).unwrap();
```

to

```rust
    let b = cast(
        &array,
        &DataType::UInt8,
        CastOptions {
            wrapped: true,
            ..Default::default()
        },
    )
```

and add `Default::default()` as an argument to `cast`

sundy-li:
Changes:

- Add partial option into CastOptions, this is trying to be compatible with [MySQL behavior](https://dev.mysql.com/doc/refman/8.0/en/cast-functions.html),  `123 abc` - > `123`
- Try to make CastOptions pub again

closes #569

### DIFF
--- a/src/compute/cast/binary_to.rs
+++ b/src/compute/cast/binary_to.rs
@@ -41,7 +41,7 @@ pub fn binary_to_primitive<O: Offset, T>(
 where
     T: NativeType + lexical_core::FromLexical,
 {
-    let parse_fn = if options.partial {
+    let parse_fn = if !options.partial {
         |x| lexical_core::parse(x).ok()
     } else {
         |x| lexical_core::parse_partial(x).ok().map(|x| x.0)
--- a/src/compute/cast/binary_to.rs
+++ b/src/compute/cast/binary_to.rs
@@ -3,6 +3,8 @@ use std::convert::TryFrom;
 use crate::error::{ArrowError, Result};
 use crate::{array::*, buffer::Buffer, datatypes::DataType, types::NativeType};
 
+use super::CastOptions;
+
 /// Conversion of binary
 pub fn binary_to_large_binary(from: &BinaryArray<i32>, to_data_type: DataType) -> BinaryArray<i64> {
     let values = from.values().clone();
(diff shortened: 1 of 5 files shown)


sundy-li:
@jorgecarleitao Hello, is there any problem with this change?

src/compute/cast/binary_to.rs
```diff
@@ -31,26 +33,35 @@ pub fn binary_large_to_binary(
...
+    to: &DataType,
+    options: CastOptions,
+) -> PrimitiveArray<T>
 where
     T: NativeType + lexical_core::FromLexical,
 {
-    let iter = from
-        .iter()
-        .map(|x| x.and_then::<T, _>(|x| lexical_core::parse(x).ok()));
+    let parse_fn = if !options.partial {
```
jorgecarleitao, line 44: I think we need to put the whole `PrimitiveArray::<T>::from_trusted_len_iter(iter).to(to.clone())` inside the if otherwise the compiler can't inline the function call as it branches depending on the parameter.

src/compute/cast/utf8_to.rs
```diff
@@ -11,29 +11,42 @@ use crate::{
...
-pub fn utf8_to_primitive<O: Offset, T>(from: &Utf8Array<O>, to: &DataType) -> PrimitiveArray<T>
+pub fn utf8_to_primitive<O: Offset, T>(
+    from: &Utf8Array<O>,
+    to: &DataType,
+    options: CastOptions,
+) -> PrimitiveArray<T>
 where
     T: NativeType + lexical_core::FromLexical,
 {
+    let parse_fn = if !options.partial {
```
jorgecarleitao, line 27: same here

src/compute/cast/utf8_to.rs
```diff
@@ -11,29 +11,42 @@ use crate::{
     },
 };
 
+use super::CastOptions;
+
 const RFC3339: &str = "%Y-%m-%dT%H:%M:%S%.f%:z";
 
 /// Casts a [`Utf8Array`] to a [`PrimitiveArray`], making any uncastable value a Null.
-pub fn utf8_to_primitive<O: Offset, T>(from: &Utf8Array<O>, to: &DataType) -> PrimitiveArray<T>
```
jorgecarleitao, line 17: I would suggest a new function name and not pass the options, but maybe there is a reason?

src/compute/cast/mod.rs
```diff
@@ -324,14 +327,35 @@ fn cast_large_to_list(array: &ListArray<i64>, to_type: &DataType) -> ListArray<i
...
+        CastOptions {
+            wrapped: true,
+            partial: false,
+        },
+    )
+}
+
+/// Similar to [`cast`], but parse the utf8/binary into integer at the best-effort.
+/// Behavior:
+/// * PrimitiveArray to PrimitiveArray: overflowing cast will be wrapped (i.e. `256i16 as u8 = 0` vectorized).
```
jorgecarleitao, line 349: this is not true, right? I think we need to pass the `CastOptions` here, since we how have 4 combinations for each value of `wrapped` and `partial`. I would go ahead and change/add `pub fn cast(array, to_type, options)` and remove the other functions. This is a sufficiently important use-case to break the API. :)

jorgecarleitao (commented):
Sorry for the delay. Missed this one. Thanks a lot for the PR!

Was not aware of the partial; cool concept. Left some comments: imo we should break the API, make `CastOptions` first class, and remove the old APIs that were branching on the wrapping, as we now have 4 cases to cater for.

jorgecarleitao:
Looks awesome! Thanks a lot for the PR.

PR #563, opened 2021-10-31
Made bitmap not cache null count

jorgecarleitao:
This PR removes the cached `null_count` on `Bitmap`, making the computation only happen on request.

This makes the struct simpler and removes the need to compute the count even if the count will not be used.

### DIFF
--- a/src/bitmap/immutable.rs
+++ b/src/bitmap/immutable.rs
@@ -20,8 +20,6 @@ pub struct Bitmap {
     // both are measured in bits. They are used to bound the bitmap to a region of Bytes.
     offset: usize,
     length: usize,
-    // this is a cache: it must be computed on initialization
-    null_count: usize,
 }
 
 impl std::fmt::Debug for Bitmap {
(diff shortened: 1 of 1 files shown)


jorgecarleitao:
what do you think @ritchie46 @houqp @Dandandan ?

ritchie46:
What do you think about a lazy cache? Same to `init_validities()`. That we only set the null count at first request. I can imagine that we have long living arrays that are base for many computations. 

For instance in Polars, almost every operation starts with a null_count check to see if we can use a fast path.

jorgecarleitao:
Doesn't lazy initialization require `&mut self`?

> For instance in Polars, almost every operation starts with a null_count check to see if we can use a fast path.

shouldn't it start with `if let Some(bitmap) = array.validity()`? Or is it common to have arrays with only null values?

ritchie46:
> Doesn't lazy initialization require `&mut self`?
> 
> > For instance in Polars, almost every operation starts with a null_count check to see if we can use a fast path.
> 
> shouldn't it start with `if let Some(bitmap) = array.validity()`? Or is it common to have arrays with only null values?

The series has potentially multiple chunks, so we check null_count. But indeed, I could add the branch check for the validity and return 0 if it is missing. Thanks, now it should not matter that much.

ritchie46:
Just for clarity, I think it is a good enhancement. :+1:  :)

jorgecarleitao:
Reverted because it triggered a major regression in polars, so more care is needed here.

PR #567, opened 2021-11-03
Transform like pattern to regex

zhyass:
Add transform like pattern to regex to fixes https://github.com/datafuselabs/databend/issues/2478 .

1. Use backslash to escape the regex special character.
2. Process the backslash escape.

e.g. 
'Hello\\._World%\\%' tranform to '^Hello\\\\\\..World.*%$'.

### DIFF
--- a/src/compute/like.rs
+++ b/src/compute/like.rs
@@ -12,8 +12,8 @@ use crate::{
 };
 
 #[inline]
-fn is_like_pattern(c: char) -> bool {
-    c == '%' || c == '_'
+fn is_like_pattern_escape(c: char) -> bool {
+    c == '%' || c == '_' || c == '\\'
 }
 
 #[inline]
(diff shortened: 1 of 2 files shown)


jorgecarleitao:
Hi, I do not think we should do this in `regex` function: there are too many edge cases in the different flavors of regex, making it impossible to please everyone. 

I have two suggestions:

1. make DataBlend implement its own regex, based on the specific rules that you would like (mysql?).
2. Alternatively, define a new regex here called `regex_mysql`, that implements the semantics mysql's regex

zhyass:
> Hi, I do not think we should do this in `regex` function: there are too many edge cases in the different flavors of regex, making it impossible to please everyone.
> 
> I have two suggestions:
> 
>     1. make DataBlend implement its own regex, based on the specific rules that you would like (mysql?).
> 
>     2. Alternatively, define a new regex here called `regex_mysql`, that implements the semantics mysql's regex

Thanks! I will try to realize the first suggestion. The pr will be closed.

jorgecarleitao:
feel free to ping me there if you would like me to take a look at it 👍

sundy-li:
> make DataBlend implement its own regex, based on the specific rules that you would like (mysql?).

Hello, there is a typo, its name is Databend

PR #573, opened 2021-11-04
Made `null_count` not cached

jorgecarleitao:


ritchie46:
Doing a search for `null_count()` shows that there are still quite a lot of branches in main that compute the count to determine the if there are null values. Maybe we should first rewrite those branches, or maybe check when we do those null counts. If we still do them on almost every array creation, its maybe better to keep them cached?

jorgecarleitao:
I agree @ritchie46 , I do not think we are ready (and unclear whether we ever be ready) for this change.

ritchie46:
>  I do not think we are ready (and unclear whether we ever be ready) for this change.

In that context: I also found that in polars we have quite some fast paths for `array.null_count() == array.len()` which are valuable.

jorgecarleitao:
Closing as not useful. :)

PR #582, opened 2021-11-07
fix growable negative keys

ritchie46:
Pyarrow creates dictionary keys with negative integers, that are masked out by the validity bitmap.

This PR fixes conversion from those negative integers to usize, they are mapped to `0`, and that is ok, because the values are masked.

Related downstream issue: https://github.com/pola-rs/polars/issues/1686

Additionally the scalar API also read masked out values, so was incorrect.

### DIFF
--- a/src/array/dictionary/mod.rs
+++ b/src/array/dictionary/mod.rs
@@ -14,6 +14,7 @@ pub use iterator::*;
 pub use mutable::*;
 
 use super::{new_empty_array, primitive::PrimitiveArray, Array};
+use crate::scalar::NullScalar;
 
 /// Trait denoting [`NativeType`]s that can be used as keys of a dictionary.
 pub trait DictionaryKey:
@@ -127,8 +128,12 @@ impl<K: DictionaryKey> DictionaryArray<K> {
     /// Returns the value of the [`DictionaryArray`] at position `i`.
     #[inline]
     pub fn value(&self, index: usize) -> Box<dyn Scalar> {
-        let index = self.keys.value(index).to_usize().unwrap();
-        new_scalar(self.values.as_ref(), index)
+        if self.keys.is_null(index) {
+            Box::new(NullScalar::new())
+        } else {
+            let index = self.keys.value(index).to_usize().unwrap();
+            new_scalar(self.values.as_ref(), index)
+        }
     }
 }
 
--- a/src/array/growable/dictionary.rs
+++ b/src/array/growable/dictionary.rs
@@ -100,7 +100,7 @@ impl<'a, T: DictionaryKey> Growable<'a> for GrowableDictionary<'a, T> {
         self.key_values.extend(
             values
                 .iter()
-                .map(|x| T::from_usize(offset + x.to_usize().unwrap()).unwrap()),
+                .map(|x| T::from_usize(offset + x.to_usize().unwrap_or(0)).unwrap()),
         );
     }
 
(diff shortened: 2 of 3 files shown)


ghuls:
It indeed can take a while before it is fixed in pyarrow, as pyarrow 6.0.0 is released a few days ago.

jorgecarleitao:
wow, this looks like a vulnerability to me on the c++ side. if someone accesses those keys, they may read out of bounds if they do not check bounds (which some implementations do). I think this should be addressed in the pyarrow implementation: dictionary keys and offset deltas should never be negative. I think that our implementation is correct: it is the user responsibility to ensure that the keys are positive, even if the type is a signed integer (for JVM reasons).

I will follow up on this on the Arrow side.

ritchie46:
What do you think about a temporary patch? ^_^. Before a new pyarrow is shipped to all pandas code that uses this, I guess that will take some months. :see_no_evil:

src/array/dictionary/mod.rs
```diff
@@ -127,8 +128,12 @@ impl<K: DictionaryKey> DictionaryArray<K> {
     /// Returns the value of the [`DictionaryArray`] at position `i`.
     #[inline]
     pub fn value(&self, index: usize) -> Box<dyn Scalar> {
-        let index = self.keys.value(index).to_usize().unwrap();
-        new_scalar(self.values.as_ref(), index)
+        if self.keys.is_null(index) {
+            Box::new(NullScalar::new())
```
jorgecarleitao, line 132: `NullScalar.data_type` yields `DataType::Null`, but the value yields a scalar of `values.data_type()`. This needs to be a Scalar of the same type as inner values (yeah, this API is still not very useful: technically the scalar should be "DictionaryScalar" for its data_type to be equal to the array's DataType :/

src/array/growable/dictionary.rs
```diff
@@ -100,7 +100,7 @@ impl<'a, T: DictionaryKey> Growable<'a> for GrowableDictionary<'a, T> {
         self.key_values.extend(
             values
                 .iter()
-                .map(|x| T::from_usize(offset + x.to_usize().unwrap()).unwrap()),
+                .map(|x| T::from_usize(offset + x.to_usize().unwrap_or(0)).unwrap()),
```
jorgecarleitao, line 103:
```suggestion
                // `.unwrap_or(0)` because this operation does not check for null values, which may contain any key.
                .map(|x| T::from_usize(offset + x.to_usize().unwrap_or(0)).unwrap()),
```

tests/it/array/growable/dictionary.rs
```diff
@@ -29,6 +30,24 @@ fn test_single() -> Result<()> {
...
+
+    let keys = PrimitiveArray::from_data(
+        DataType::Int32,
+        keys.into(),
+        Some(vec![true, true, true, false].into()),
+    );
+
+    let arr = DictionaryArray::from_data(keys, Arc::new(Utf8Array::<i32>::from(vals)));
+    // check that we don't panic with negative keys to usize conversion
+    let out = arrow2::compute::concat::concatenate(&[&arr]).unwrap();
```
jorgecarleitao, line 46: Can we replace this call by the growable API?

`cargo test --tests` (without features) is very useful to run when not iterating over the compute module kernels, and this usage causes it to break.

jorgecarleitao (commented):
Thanks a lot!

I was wrong, this is valid in arrow and it is a bug here: we should accept negative keys if they are nulls. Null values can have any (initialized) value.

Left two minor comments, but otherwise ready to sail.

src/array/dictionary/mod.rs
ritchie46: Is there a fix for this atm? Now the test fail, because it tries to read from a negative (masked out) index.

jorgecarleitao: Meh, let's keep it for now like this; better than nothing. xD

PR #586, opened 2021-11-07
Speedup/simplify bitwise operations (avoid extra allocation)

Dandandan:
Avoids an extra allocation for binary bitwise ops.

### DIFF
--- a/src/bitmap/bitmap_ops.rs
+++ b/src/bitmap/bitmap_ops.rs
@@ -20,6 +20,11 @@ where
     let mut a3_chunks = a3.chunks();
     let mut a4_chunks = a4.chunks();
 
+    let rem_a1 = a1_chunks.remainder();
+    let rem_a2 = a2_chunks.remainder();
+    let rem_a3 = a3_chunks.remainder();
+    let rem_a4 = a4_chunks.remainder();
+
     let chunks = a1_chunks
         .by_ref()
         .zip(a2_chunks.by_ref())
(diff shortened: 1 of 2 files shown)


src/bitmap/bitmap_ops.rs
```diff
@@ -85,18 +83,20 @@ where
...
-    let remainder_bytes = lhs_chunks.remainder_len().saturating_add(7) / 8;
-    let rem = op(lhs_chunks.remainder(), rhs_chunks.remainder());
-    let rem = &rem.to_ne_bytes()[..remainder_bytes];
-    buffer.extend_from_slice(rem);
+    // Soundness: `BitChunks` is a trusted len iterator
+    let buffer = unsafe {
+        MutableBuffer::from_chunk_iter_unchecked(
+            chunks.chain(std::iter::once(op(rem_lhs, rem_rhs))),
+        )
+    };
```
jorgecarleitao, line 99:
```suggestion
    let lhs_chunks = lhs.chunks();
    let rhs_chunks = rhs.chunks();

    let rem = op(lhs_chunks.remainder(), rhs_chunks.remainder());

    let chunks = lhs_chunks
        .zip(rhs_chunks)
        .map(|(left, right)| op(left, right));
    let chunks = chunks.chain(std::iter::once(rem));
    let buffer = MutableBuffer::from_chunk_iter(chunks);
```

so that we do not need the `unsafe`: zip and `chain(once(` is trusted len and we can use the traits to get that one if we do not use `by_ref`

jorgecarleitao:
please ignore the security audit: I created the advisory for flatbuffers but have not finished auditing this and `arrow-format` crate to double check that we are not using any of the unsound but safe APIs incorrectly.

src/bitmap/bitmap_ops.rs
Dandandan: Thanks - great suggestion!

PR #589, opened 2021-11-09
allow null array to be casted to all other types

flaneur2020:
this PR is a port of https://github.com/apache/arrow-rs/pull/884


jorgecarleitao:
Thanks a lot, @flaneur2020 !

PR #596, opened 2021-11-10
Fixed error in computing projection in `io::ipc::read::reader::FileReader`

### LINKED ISSUE #594: arrow2::io::ipc::read::reader::FileReader cannot set 0 as projection
Hi, Thanks for the very excellent project!

I would like to parse IPC files, but I cannot set 0 as projection like `arrow2::io::ipc::read::reader::FileReader::new(&mut reader, metadata, vec![0, 1])`. The cause of this seems to come from https://github.com/jorgecarleitao/arrow2/blob/add3e2a01068435f297197a28df885ab5bbf1653/src/io/ipc/read/reader.rs#L237-L243

If projection is `vec![0, 1]`, both `acc` and `v` equal 0, so error occurred.

It seems that code like this is correct,

```rust
let _ = projection.iter().fold(0, |mut acc, v| {
    if !(acc == 0 && *v == 0) {
        assert!( 
            *v > acc, 
            "The projection on IPC must be ordered and non-overlapping" 
        ); 
    }
    acc = *v;
    acc
});
```

Can I create the PR?

illumination-k:
Allow projection to be set to 0 in io::ipc::read::reader::FileReader::new by using `windows(2)`.
modified it according to [this comment](https://github.com/jorgecarleitao/arrow2/issues/594#issuecomment-965545539)

close #594

### DIFF
--- a/src/io/ipc/read/reader.rs
+++ b/src/io/ipc/read/reader.rs
@@ -236,13 +236,11 @@ impl<R: Read + Seek> FileReader<R> {
     /// Panics iff the projection is not in increasing order (e.g. `[1, 0]` nor `[0, 1, 1]` are valid)
     pub fn new(reader: R, metadata: FileMetadata, projection: Option<Vec<usize>>) -> Self {
         if let Some(projection) = projection.as_ref() {
-            let _ = projection.iter().fold(0, |mut acc, v| {
+            let _ = projection.windows(2).for_each(|x| {
                 assert!(
-                    *v > acc,
+                    x[0] < x[1],
                     "The projection on IPC must be ordered and non-overlapping"
                 );
-                acc = *v;
-                acc
             });
         }
         let projection = projection.map(|projection| {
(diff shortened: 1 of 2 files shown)


src/io/ipc/read/reader.rs
```diff
@@ -236,7 +236,7 @@ impl<R: Read + Seek> FileReader<R> {
     /// Panics iff the projection is not in increasing order (e.g. `[1, 0]` nor `[0, 1, 1]` are valid)
     pub fn new(reader: R, metadata: FileMetadata, projection: Option<Vec<usize>>) -> Self {
         if let Some(projection) = projection.as_ref() {
-            let _ = projection.windows(2).for_each(|x| {
+            projection.windows(2).for_each(|x| {
```
Dandandan, line 239: Looks wonderful! Maybe we can add a test for this to make sure it doesn't panic now?

jorgecarleitao: Pointer for the tests: https://github.com/jorgecarleitao/arrow2/blob/main/tests/it/io/ipc/read/file.rs#L181 (changing from 1 to 0 should be enough :) )

Dandandan (approved):

Issue #527, opened 2021-10-13
Proposal: Remove the code related to the Operator enumeration


yjhmelody:
I recommend removing the Operator enumeration.
If arrow users build their own computing systems, they will generally not use the operator enumeration values defined by arrow, because there are often differences in the number of operators defined, or differences in classification, such as the lack of bitwise operators, and there may be more in the future for arithmetic operator.
Or lack some traits implementation, such as serde.

It is very boilerplate code to distribute various operations based on Operator. If it can be removed, the amount of arrow code can be greatly reduced.

houqp:
I agree it's a good idea to not make them part of the core so downstream won't be forced to pull them in. But I think there is value in keeping the code and convert them into a feature or self-contained crate for the following two reasons:

* To be feature parity with the cpp implementation
* For use-case where downstream just want an out of the box simple compute module to get stuff done without having to write these boilerplate code

jorgecarleitao:
@houqp , note that the kernels themselves are kept. I think the proposal is:

use `add_primitive(lhs, rhs)` instead of `arithmetics_primitive<T: NativeType>(lhs, Operator::Add, rhs)`.

The issue with `arithmetics_primitive<T: NativeType>(lhs, Operator::Add, rhs)` is that it requires `T: Add + Sub + Mul + ...`, which may not be fulfiled by all native types (e.g. `months_days_ns` is probably `Add` and `Sub` but not necessarily `Div`?).

An equivalent problem exists in comparison `compare_primitive<T: NativeType>(lhs, Operator, rhs)` requires `T: PartialEq + PartialOrd` but `months_days_ns` has a similar issue.

houqp:
Yeah, I only took a quick scan on the CPP code and noticed they have these comparison operators defined. But on a second thought, I think doing something just because the CPP code is doing it is not a good technical argument.

Personally, I think writing `add_primitive(lhs, rhs)` is cleaner and more readable than `arithmetics_primitive<T>(lhs, Operator::Add, rhs)`. Similarily, I would prefer `compare_lt(a, b)` over `compare(a, b, Operator::Lt)`.

That said, the current operator related code does more than just operator based function dispatching. It also handles array type matching to provide high level APIs centered around Array trait object like other compute modules. The latter is the what I was suggesting us to keep to make the interface more consistent across compute modules and make the library a little bit easier for new users to adopt for less complicated applications.

In short, I am 100% on board with removing all the operators.

PR #604, opened 2021-11-13
Refactored comparison `Operator`

jorgecarleitao:
This PR removes the `Operator` and replaces it by different function calls as proposed by #527 . This does not close that issue yet because a similar PR is necessary for the arithmetic kernels.

This PR removes the enum `Operator` in `compute::comparison`. Instead, the API is composed by

```
arrow2::compute::comparison::Y;
arrow2::compute::comparison::X::Y;
// where X \in {primitive, boolean, utf8, binary}
// where Y \in {eq, neq, lt, lt_eq, gt, gt_eq, eq_scalar, neq_scalar, lt_scalar, lt_eq_scalar, gt_scalar, gt_eq_scalar}
```

the functions inside each `X` are statically-typed (e.g. expect `Utf8Array`), while the functions `arrow2::compute::comparison::Y` are dynamically-typed (i.e. expect `Array` and `Scalar`).

As usual, each dynamic variant contains a `can_Y(&DataType)` that returns whether the operation is valid / is implemented.

Also, the functions no longer return a `Result` and instead simply panic; this is because:
* there is no obvious recovery when the arrays have a different length; the user should check that before using these functions.
* whether a type is valid for each function is already exposed in `can_Y`, which users can use to check whether the operation will panic or not.

This simplifies the overall API and allows implementing equality to types that have no partial or total order defined.

### DIFF
--- a/src/array/utf8/mod.rs
+++ b/src/array/utf8/mod.rs
@@ -178,7 +178,7 @@ impl<O: Offset> Utf8Array<O> {
 impl<O: Offset> Utf8Array<O> {
     /// Returns the length of this array
     #[inline]
-    fn len(&self) -> usize {
+    pub fn len(&self) -> usize {
         self.offsets.len() - 1
     }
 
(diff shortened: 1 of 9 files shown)


houqp:
great simplification, thanks  @yjhmelody for the suggestion :)

PR #607, opened 2021-11-15
Simplified arithmetics compute

### LINKED ISSUE #527: Proposal: Remove the code related to the Operator enumeration

jorgecarleitao:
Closes #527.

The functions are now offered based on the name instead of `Operator` (which was removed).

Also, the functions no longer error, and instead panic. As usual, `can_*` can be used to check if the operation is supported, and checking the `len` of the arrays is required.

### DIFF
--- a/benches/arithmetic_kernels.rs
+++ b/benches/arithmetic_kernels.rs
@@ -22,7 +22,7 @@ fn bench_add<T>(lhs: &PrimitiveArray<T>, rhs: &PrimitiveArray<T>)
 where
     T: NativeType + Add<Output = T> + NumCast,
 {
-    criterion::black_box(add(lhs, rhs)).unwrap();
+    criterion::black_box(add(lhs, rhs));
 }
 
 fn add_benchmark(c: &mut Criterion) {
--- a/benches/comparison_kernels.rs
+++ b/benches/comparison_kernels.rs
@@ -2,7 +2,7 @@ use criterion::{criterion_group, criterion_main, Criterion};
 
 use arrow2::scalar::*;
 use arrow2::util::bench_util::*;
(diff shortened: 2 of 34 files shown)


yjhmelody (approved):
LGTM, but if you can use expect to prompt the user that the type does not match, it will be better than unwrap directly

jorgecarleitao:
@yjhmelody , thanks! Could you clarify what do you mean with

>  the user that the type does not match, it will be better than unwrap directly

So, the design here is: the statically-typed versions of the operators do not check for the type; the dynamically-typed versions of the operators check for type.

I am uncertain about this choice, though: my hypothesis here:
1. query engines know the logical types associated to the arrays
2. they also know whether an operation is valid for a given operator (via `can_*`)

thus, they can match for a valid datatype via

```rust
if can_add(array1.data_type(), array2.data_type()) {
    Ok(add(array1, array2))
} else {
    // custom add for arrow2-unsupported type or error
}
```

Could you clarify your thoughts on this? Also @houqp @ritchie46 @sundy-li

yjhmelody:
I agree with you. I just suggest not to use the `unwrap` method directly, but to use the `expect` method

jorgecarleitao:
iirc unwrapping a Result and will show the result's error message on the panic.

PR #613, opened 2021-11-17
Implement iterator for StructArray

illumination-k:
I implemented Iterator for StructArray. 
Also, the `struct_.rs` file was changed to a directory with `mod.rs` and the `ffi` part was split into `ffi.rs`.

If you need me to open an issue on this PR, I will do it.

### DIFF
--- a/src/array/struct_/iterator.rs
+++ b/src/array/struct_/iterator.rs
@@ -1,36 +1,100 @@
-use std::slice::Iter;
-use std::sync::Arc;
-
 use crate::{
-    array::Array,
     bitmap::utils::{zip_validity, ZipValidity},
+    scalar::{new_scalar, Scalar},
+    trusted_len::TrustedLen,
 };
 
 use super::StructArray;
 
-impl<'a> IntoIterator for &'a StructArray {
-    type Item = Option<&'a Arc<dyn Array>>;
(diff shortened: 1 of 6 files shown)


jorgecarleitao:
Thanks for the PR! No need for an issue :)

The iterators of arrays are iterators over rows.

An iterator over a StructArray would be something like `Iterator<Item=Option<Vec<Box<dyn Scalar>>>`, since each row is composed by multiple entries (one per field) where the `Option` is the Struct's validity.

illumination-k:
Thanks for your comments! I understood it wrong. Maybe, I think I have corrected it.

src/array/struct_/iterator.rs
```diff
@@ -0,0 +1,100 @@
...
+impl<'a> DoubleEndedIterator for StructValueIter<'a> {
+    #[inline]
+    fn next_back(&mut self) -> Option<Self::Item> {
+        if self.index == self.end {
+            None
+        } else {
+            self.end -= 1;
+
+            let mut item = vec![];
+            for i in 0..self.array.fields().len() {
```
jorgecarleitao, line 64: same idea here.

jorgecarleitao (commented):
Thanks a lot, @illumination-k, Looks great! 

I left a comment to simplify the loop.

The idea in Rust is that when using iterator pattern, rust skips bound checks (it also makes it a bit easier to read for non-C developers)

PR #622, opened 2021-11-22
add to conversion to fixedsizebinary

ritchie46:
This adds `fn to` conversion for `FixedSizeBinary`. This makes converting to another logical type consistent with primitive arrays. 

I could add it to the other arrays as well.

### DIFF
--- a/src/array/fixed_size_binary/mod.rs
+++ b/src/array/fixed_size_binary/mod.rs
@@ -136,6 +136,18 @@ impl FixedSizeBinaryArray {
             .get_unchecked(i * self.size..(i + 1) * self.size)
     }
 
+    /// Returns a new [`FixedSizeBinary`] with a different logical type.
+    /// This is `O(1)`.
+    #[inline]
+    pub fn to(self, data_type: DataType) -> Self {
+        Self {
+            size: self.size,
+            data_type,
+            values: self.values,
+            validity: self.validity,
+        }
+    }
+
     /// Returns the size
     pub fn size(&self) -> usize {
         self.size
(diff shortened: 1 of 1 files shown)


jorgecarleitao:
Thanks! I think we need to check that the `data_type` is a valid one and panic if not, like we do in `from_data`.

ritchie46:
> Thanks! I think we need to check that the `data_type` is a valid one and panic if not, like we do in `from_data`.

I didn't fully understand what we mean by valid in this case? If we use an extension type, do we know what is valid or not?

jorgecarleitao:
The function `get_size` is the one that validates in the `from_data`, [here](https://github.com/jorgecarleitao/arrow2/pull/622/files#diff-39c0771210313d62d8ada8dafc2b633c5a8c2d3c9492589d2c3b669bb7e25944L147).

jorgecarleitao:
(note that the `.to_logical_type` converts any extension type to its internal type, to take extension types into account ^_^)

src/array/fixed_size_binary/mod.rs
```diff
@@ -136,6 +136,29 @@ impl FixedSizeBinaryArray {
             .get_unchecked(i * self.size..(i + 1) * self.size)
     }
 
+    /// Returns a new [`FixedSizeBinary`] with a different logical type.
+    /// This is `O(1)`.
+    /// # Panics
+    /// Panics iff the data_type is not supported for the physical type.
+    #[inline]
+    pub fn to(self, data_type: DataType) -> Self {
+        match (
```
ritchie46, line 145: Something like this? I assume we must check that the physical types are correct.

jorgecarleitao: exactly, this works!

PR #630, opened 2021-11-24
Utf8 add unchecked trusted_len_values_iter

ritchie46:
Similar to what already exists for `Iterator<Item=Option<&str>>`, now expose that for `Iterator<Item=<&str>>`


jorgecarleitao:
Thanks!

PR #633, opened 2021-11-25
Faster `take` with null values (2-3x)

jorgecarleitao:
This was achieved by splitting the "take" of values and validity as two separate loops. It seems to vectorize better.

```bash
# native = skylake-avx512
RUSTFLAGS="-C target-cpu=native" cargo bench --features compute,benchmarks \
    --bench take_kernels -- "take .* values nulls 2\^20"
```

```
take i32 values nulls 2^20                                                                            
                        time:   [6.2657 ms 6.2879 ms 6.3151 ms]
                        change: [-57.959% -57.507% -57.123%] (p = 0.00 < 0.05)
take bool values nulls 2^20                                                                             
                        time:   [4.4977 ms 4.5077 ms 4.5197 ms]
                        change: [-69.940% -69.411% -68.964%] (p = 0.00 < 0.05)
take str values nulls 2^20                                                                            
                        time:   [25.706 ms 25.878 ms 26.063 ms]
                        change: [-20.546% -19.827% -19.031%] (p = 0.00 < 0.05)
```

This was a side observation of https://github.com/DataEngineeringLabs/simd-benches/pull/1 which showed that 

```rust
pub fn naive_take(values: &[f32], indices: &[usize]) -> Vec<f32> {
    indices.iter().map(|i| values[*i]).collect()
}
```

is compiled to the same instructions as using [portable_simd's `gather`](https://rust-lang.github.io/portable-simd/core_simd/simd/struct.Simd.html#method.gather_select), suggesting that the compiler is quite smart in compiling `naive_take` and we can thus leverage that.

### DIFF
--- a/benches/take_kernels.rs
+++ b/benches/take_kernels.rs
@@ -54,6 +54,11 @@ fn add_benchmark(c: &mut Criterion) {
             b.iter(|| bench_take(&values, &indices))
         });
 
+        let values_nulls = create_boolean_array(size, 0.2, 0.5);
+        c.bench_function(&format!("take bool values nulls 2^{}", log2_size), |b| {
+            b.iter(|| bench_take(&values_nulls, &indices))
+        });
+
         c.bench_function(&format!("take bool nulls 2^{}", log2_size), |b| {
             b.iter(|| bench_take(&values, &indices_nulls))
         });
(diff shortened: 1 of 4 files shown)


ritchie46:
Intersting. I also had a similar result with splitting the bound check in two loops.

Maybe that makes sense here as well. Als it is now checked in the validity and in the values.

Dandandan:
Cool results 😎

PR #634, opened 2021-11-25
Split "compute" feature

jorgecarleitao:
The feature "compute" is a bit painful to compile. This PR splits the feature in individual features so that consumers can drag what they want from the `compute` module.

There is a longer-term solution of splitting the crate, but I am still figuring out what is the best approach for that. Meanwhile, with this people can go ahead and only drag the parts of the compute that they need :)

## Backward incompatible changes

* `compute::concat -> compute::concatenate` for consistency
* feature `regex` was replaced by `compute_regex_match` and `compute_like`
* feature `merge_sort` was renamed to `compute_merge_sort`

Other than this, the feature `compute` is now just an alias for the union of all `compute_*` features.


ritchie46:
Very nice!

Issue #635, opened 2021-11-26
Add `lower` and `upper` support for string


Xuanwo:
Hello, do you have a plan to support `compute::lower::lower` and `compute::upper::upper`?

If it's OK, I will start a PR to implement those two functions.

> They are required by https://github.com/datafuselabs/databend/issues/3013 and https://github.com/datafuselabs/databend/issues/3010

jorgecarleitao:
Yeap, in scope indeed. 👍

Hint: there is a cool optimization on them

Xuanwo:
> Hint: there is a cool optimization on them

Wow, sounds interesting. How about giving me some more detailed tips?

PR #641, opened 2021-11-27
compute: Add lower support

Xuanwo:
Signed-off-by: Xuanwo <github@xuanwo.io>

Part of #635


src/compute/lower.rs
```diff
@@ -0,0 +1,79 @@
...
+use crate::array::*;
+use crate::{
+    datatypes::DataType,
+    error::{ArrowError, Result},
+};
+
+fn utf8_lower<O: Offset>(array: &Utf8Array<O>) -> Utf8Array<O> {
+    let iter = array.values_iter().map(str::to_lowercase);
+
+    let new = Utf8Array::<O>::from_trusted_len_values_iter(iter);
```
jorgecarleitao, line 29: originally I thought that an invariant of lower would preserve the number of bytes on the string and thus could be optimized away, but now that I look at this I am not sure it does any more xD

src/compute/lower.rs
```diff
@@ -0,0 +1,79 @@
...
+    new.with_validity(array.validity().cloned())
+}
+
+/// Returns an ArrayRef with lowercase of each of the elements in `array`.
```
jorgecarleitao, line 33:
```suggestion
/// Returns a new `Array` where each of each of the elements is lower-cased.
```

src/compute/lower.rs
```diff
@@ -0,0 +1,79 @@
...
+/// this function errors when the passed array is not a \[Large\]String array.
+pub fn lower(array: &dyn Array) -> Result<Box<dyn Array>> {
+    match array.data_type() {
+        // For binary and large binary, lower is no-op.
+        DataType::Binary | DataType::LargeBinary => unsafe {
```
jorgecarleitao, line 38: Could you clarify why we should support "lower" for Binary? Is there a well-defined concept of "lowercase" for blob/binary?

jorgecarleitao (commented):
Thanks a lot! Left some comments

src/compute/lower.rs
Xuanwo: Utf8 is so complex that I don't know is it possible.

src/compute/lower.rs
Xuanwo: Well, it's a decision that should be made by the maintainer.

There are three statuses:

- utf8 / large utf8: well as expected.
- binary / large binary: no-op, but no error.
- other data types: incorrect, and we should return an error.

Most implementations will allow users to lower a binary (even it is meaningless). If arrow2 doesn't support binary, we will need to check the datatype before calling `lower`.

src/compute/lower.rs
```diff
@@ -0,0 +1,79 @@
...
+            array
+                .as_any()
+                .downcast_ref::<Utf8Array<i64>>()
+                .expect("A large string is expected"),
+        ))),
+        DataType::Utf8 => Ok(Box::new(utf8_lower(
+            array
+                .as_any()
+                .downcast_ref::<Utf8Array<i32>>()
+                .expect("A large string is expected"),
```
Xuanwo, line 52:
```suggestion
                .expect("A string is expected"),
```

src/compute/lower.rs
ritchie46: IMO, an implementation of Arrow should decide if a lower in binary makes sense. I don't think it does and I feel that arrow2 should error/panic.

src/compute/lower.rs
```diff
@@ -0,0 +1,79 @@
...
+
+//! Defines kernel to extract a lower case of a \[Large\]StringArray
+
+use crate::array::*;
+use crate::{
+    datatypes::DataType,
+    error::{ArrowError, Result},
+};
+
+fn utf8_lower<O: Offset>(array: &Utf8Array<O>) -> Utf8Array<O> {
```
ritchie46, line 26: Given that there are many possible functions on a `&str`. 

Maybe we can make a `utf8_apply` that is generic over the function applied to the utf8 array.

This would not incur any compilation cost until implemented. 

We could the  simply document the `lower` and `higher`, or add them under a pre-compiled feature flag.

Xuanwo: I will give it a try

jorgecarleitao: that is a good idea; it makes it very easy for others to add new kernels.

ghuls:
Shouldn't the test include some real UTF8 characters instead of only the ASCII subset?
Like some of the characters listed here: https://www.ibm.com/docs/en/i/7.2?topic=tables-unicode-lowercase-uppercase-conversion-mapping-table

src/compute/lower.rs
Xuanwo: Addressed.

Xuanwo:
> Shouldn't the test include some real UTF8 characters instead of only the ASCII subset?

Addressed.

src/compute/lower.rs
Xuanwo: Addressed, I'm not sure whether this is what you want, PTAL.

jorgecarleitao:
Thanks a lot, really well tested! 💯

PR #642, opened 2021-11-27
Fix compilation with individual features activated

ritchie46:
Could not release polars without activating the whole compute feature. Lexical is needed for casting, this exposes the module when cast is activated.


ritchie46:
More coming up:

`$ cargo hack check --each-feature --no-dev-deps` reveals more.

ritchie46:
I also restored the `ArrowError::Other` variant. Maybe you want something else for that? It was still needed when compiling with avro read io.

What do you think of adding `$ cargo hack check --each-feature --no-dev-deps --exclude-features=full,compute,simd` to CI? Dealing with different feature sets becomes hard to maintain if we don't check this automatically.

It has saved me a lot when pushing seamlessly unrelated changes.

jorgecarleitao:
Thanks! dam, yeap, I was trying to get something like that; the cargo hack one is a great suggestion!

I think we should turn the error into `ArrowError::InvalidArgumentError` so that we do not need the `Other` variant. could you do that instead (or let me know and I will do it)?

I will do a hot fix on this right after this PR is merged, and handle the CI afterwards. Sorry for the bogged release.

ritchie46:
No worries! I changed the error type. It should be good to go.

PR #646, opened 2021-11-30
Improved performance in cast Primitive to Binary/String (2x)

sundy-li:
Reuse the temp bytes buffer.

```
cast int32 to binary 512                                                                             
		time:   [10.650 us 10.761 us 10.885 us]
		change: [-47.175% -46.939% -46.649%] (p = 0.00 < 0.05)
		Performance has improved.

```


jorgecarleitao:
Thanks @sundy-li ! Great idea!

PR #648, opened 2021-11-30
Simplified Primitive and Boolean scalar

jorgecarleitao:
Made `.value()` return `Option<T>` instead of `T` and simplified their internals.


yjhmelody (approved):
LGTM

PR #651, opened 2021-12-01
Improved performance in cast Primitive to Binary/String again (4x)

sundy-li:
`Memcpy` style write, no extra copy.

```
cast int32 to binary 512                                                                             
                        time:   [7.2389 us 7.2844 us 7.3306 us]
                        change: [-32.305% -31.940% -31.566%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### DIFF
--- a/src/array/binary/mutable.rs
+++ b/src/array/binary/mutable.rs
@@ -105,12 +105,13 @@ impl<O: Offset> MutableBinaryArray<O> {
         }
     }
 
-    /// Reserves `additional` slots.
-    pub fn reserve(&mut self, additional: usize) {
+    /// Reserves `additional` elements and `additional_values` on the values buffer.
+    pub fn reserve(&mut self, additional: usize, additional_values: usize) {
         self.offsets.reserve(additional);
         if let Some(x) = self.validity.as_mut() {
             x.reserve(additional)
         }
+        self.values.reserve(additional_values);
     }
 
     #[inline]
@@ -388,6 +389,29 @@ impl<O: Offset> MutableBinaryArray<O> {
         let (offsets, values) = values_iter(iterator);
         Self::from_data(Self::default_data_type(), offsets, values, None)
     }
+
+    /// Write values directly into the underlying data.
+    /// # Safety
+    /// Caller must ensure that `len <= self.capacity()`
+    #[inline]
+    pub unsafe fn write_values<F>(&mut self, mut f: F)
+    where
+        F: FnMut(&mut [u8]) -> usize,
+    {
+        // ensure values has enough capacity and size to write
+        self.values.set_len(self.values.capacity());
+        let buffer = &mut self.values.as_mut_slice()[self.offsets.last().unwrap().to_usize()..];
+        let len = f(buffer);
+        let new_len = self.offsets.last().unwrap().to_usize() + len;
+        self.values.set_len(new_len);
+
+        let size = O::from_usize(new_len).ok_or(ArrowError::Overflow).unwrap();
+        self.offsets.push(size);
+        match &mut self.validity {
+            Some(validity) => validity.push(true),
+            None => {}
+        }
+    }
 }
 
 impl<O: Offset, T: AsRef<[u8]>> Extend<Option<T>> for MutableBinaryArray<O> {
@@ -399,7 +423,7 @@ impl<O: Offset, T: AsRef<[u8]>> Extend<Option<T>> for MutableBinaryArray<O> {
 impl<O: Offset, T: AsRef<[u8]>> TryExtend<Option<T>> for MutableBinaryArray<O> {
     fn try_extend<I: IntoIterator<Item = Option<T>>>(&mut self, iter: I) -> Result<()> {
         let mut iter = iter.into_iter();
-        self.reserve(iter.size_hint().0);
+        self.reserve(iter.size_hint().0, 0);
         iter.try_for_each(|x| self.try_push(x))
     }
 }
--- a/src/array/binary/mutable.rs
+++ b/src/array/binary/mutable.rs
@@ -105,13 +105,12 @@ impl<O: Offset> MutableBinaryArray<O> {
         }
     }
 
-    /// Reserves `additional` elements and `additional_values` on the values buffer.
-    pub fn reserve(&mut self, additional: usize, additional_values: usize) {
+    /// Reserves `additional` slots.
+    pub fn reserve(&mut self, additional: usize) {
         self.offsets.reserve(additional);
         if let Some(x) = self.validity.as_mut() {
             x.reserve(additional)
         }
-        self.values.reserve(additional_values);
     }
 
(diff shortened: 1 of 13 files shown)


sundy-li:
Why this pr makes MIRI tests fail?

jorgecarleitao:
it is unrelated. There is something going on on miri dependencies that are causing some CIs to fail. Could you change the "key" parameter in the `cache` action on the `.github/test.yaml` for a cache miss? I need to investigate what is causing this.

sundy-li:
[mergifiy](https://mergify.com/)  is a good bot to have.

jorgecarleitao:
I will take a bit more to review this since it uses `unsafe`.

src/compute/cast/primitive_to.rs
```diff
@@ -17,15 +16,14 @@ use super::CastOptions;
...
             match x {
-                Some(x) => {
-                    lexical_to_bytes_mut(*x, &mut buffer);
-                    builder.push(Some(buffer.as_slice()));
-                }
+                Some(x) => unsafe {
+                    builder.reserve(1, T::FORMATTED_SIZE_DECIMAL);
+                    builder.write_values(|bytes| lexical_core::write(*x, bytes).len());
+                },
                 None => builder.push_null(),
```
jorgecarleitao, line 27: this is a really cool idea!

I think we may go a step further, though: since the size is constant, the offsets will be `[0, N, 2N, ..., M*N]` and the values can be constructed directly from `lexical_core::write`, e.g. via `extend`. We also do not need to check for utf8 below because lexical_core guarantees this. We can even ignore the validity of the primitive array and continue writing whatever is in the null slot, and clone the validity.

I think this implementation is best done without a `MutableBinaryArray`, though: we benefit from operating on the buffers directly in this case.

src/array/utf8/mutable.rs
```diff
@@ -425,6 +425,29 @@ impl<O: Offset> MutableUtf8Array<O> {
...
+    /// # Safety
+    /// Caller must ensure that `len <= self.capacity()`
+    #[inline]
+    pub unsafe fn write_values<F>(&mut self, mut f: F)
+    where
+        F: FnMut(&mut [u8]) -> usize,
+    {
+        // ensure values has enough capacity and size to write
+        self.values.set_len(self.values.capacity());
+        let buffer = &mut self.values.as_mut_slice()[self.offsets.last().unwrap().to_usize()..];
```
jorgecarleitao, line 439: I think that this is unsound, even in `unsafe` code: a slice must always have initialized data on it. I propose a different implementation below that avoids introducing another API to the `MutableBuffer`. LMK what you think.

src/compute/cast/primitive_to.rs
sundy-li: >  ` since the size is constant, the offsets will be [0, N, 2N, ..., M*N]`

It's not constant,  `T::FORMATTED_SIZE_DECIMAL` is the maximum size to reverse.

src/array/utf8/mutable.rs
sundy-li: Yes, I do agree with you. But now `MutableBuffer ` is the only way to construct a BinaryArray,  maybe we should expose `values`, `offsets` to outside.

We can have temp `values`, `offset` vectors in the cast kernel and then construct the `MutableBuffer` by these two vectors.

Refer to clickhouse's style: 
https://github.com/ClickHouse/ClickHouse/blob/515cc74530d11e1b2b18a63141b66a15b94748ba/src/Columns/ColumnString.h

sundy-li:
Performance improved again:
```
cast int32 to binary 512                                                                             
        time:   [3.4333 us 3.4434 us 3.4536 us]
        change: [-52.301% -52.104% -51.908%] (p = 0.00 < 0.05)
        Performance has improved.
```

src/compute/cast/primitive_to.rs
```diff
@@ -70,23 +84,35 @@ where
...
+            let len = lexical_core::write_unchecked(*x, bytes).len();
+
+            offset += len;
+            offsets.push(O::from_isize(offset as isize).unwrap());
+        }
+        values.set_len(offset);
+        values.shrink_to_fit();
+    }
+
+    Utf8Array::<O>::from_data(
```
jorgecarleitao, line 110:
```suggestion
    Utf8Array::<O>::from_data_unchecked(
```

since the invariants are already uphold above

src/compute/cast/primitive_to.rs
```diff
@@ -17,21 +17,35 @@ use super::CastOptions;
...
+            let len = lexical_core::write_unchecked(*x, bytes).len();
+
+            offset += len;
+            offsets.push(O::from_isize(offset as isize).unwrap());
+        }
+        values.set_len(offset);
+        values.shrink_to_fit();
+    }
+
+    BinaryArray::<O>::from_data(
```
jorgecarleitao, line 43:
```suggestion
    BinaryArray::<O>::from_data_unchecked(
```

since the invariants are uphold above

src/array/binary/mutable.rs
```diff
@@ -105,12 +105,13 @@ impl<O: Offset> MutableBinaryArray<O> {
         }
     }
 
-    /// Reserves `additional` slots.
-    pub fn reserve(&mut self, additional: usize) {
+    /// Reserves `additional` elements and `additional_values` on the values buffer.
+    pub fn reserve(&mut self, additional: usize, additional_values: usize) {
```
jorgecarleitao, line 109: do we still need this change? imo we can keep it, but the I would suggest a separate PR to handle since it is backward incompatible.

jorgecarleitao (approved):
Looks great! Left some minor comments, but overall is ready to merge.

I think that the windows does not support some of the dev dependencies, unfortunately :(

src/array/binary/mutable.rs
sundy-li: I think it's better to keep the same api like `MutableUtf8`.

sundy-li: Let's do it in another pr.

sundy-li:
After using `from_data_unchecked`
```
cast int32 to binary 512                                                                             
                        time:   [2.9751 us 2.9841 us 2.9944 us]
                        change: [-12.521% -11.660% -10.883%] (p = 0.00 < 0.05)
                        Performance has improved.
```

sundy-li:
Some thoughts:  during every `XXXToBinary` or `XXXToUtf8`, we must write these unsafe codes?  :(

In `databend`, there are many functions to be implemented to work with strings, I think it better to have a wrap function to simplify the logic.

jorgecarleitao:
Yes, that would be ideal, so that we only have to write (and test) the `unsafe` bit once. An inline function should be enough, though.

jorgecarleitao:
The PR description is outdated, it is more like -75% now ^_^

PR #652, opened 2021-12-01
make Buffer::offset public

ritchie46:
I want to know if a certain primitive array has been sliced. This seems easiest to me by checking if the backing buffer has got an offset.

### DIFF
--- a/src/buffer/immutable.rs
+++ b/src/buffer/immutable.rs
@@ -121,8 +121,9 @@ impl<T: NativeType> Buffer<T> {
         self.data.ptr()
     }
 
-    /// Returns a offset of this buffer
-    pub(crate) fn offset(&self) -> usize {
+    /// Returns the offset of this buffer.
+    #[inline]
+    pub fn offset(&self) -> usize {
         self.offset
     }
 }


jorgecarleitao:
I made it on purpose to not expose since `offset` is an implementation detail of this crate (so far). Could you just briefly describe the use-case? Curious how you need this ^_^

ritchie46:
When I build an array of dtype `Categorical` and call `n_unique` or `unique`, I can return the hashmaps length or keys respectively.

This is valid as long a the array is not sliced. To see if it is sliced I check the offset. If it is non zero, we cannot take that fast path and we compute the unique values.

I could modify the slice operation if needed. But it is an implementation detail that has useful information for downstream crates. 

Another possibility would maybe be an function that returns the memory size of the original buffer (without offset and len taken into account)?

jorgecarleitao:
Categorical <-> to a dictionary, right?

ritchie46:
No, in polars the categorical itself. It is backend by a `UInt32Array` and an internal hashmap. The length of the hashmap is most cases the length of `unique(array)`. Until we slice. :)

jorgecarleitao:
ah, and if the array is sliced, the length of the hashmap is not by itself equal to the `unique` because some items may have been "sliced away" even though they are still part of the HashMap?

ritchie46:
> ah, and if the array is sliced, the length of the hashmap is not by itself equal to the `unique` because some items may have been "sliced away" even though they are still part of the HashMap?

Indeed :)

jorgecarleitao:
could you rebase to get a green CI?

PR #653, opened 2021-12-02
Fixed filter of predicate with validity

ritchie46:
The filter kernel took the values bitmap of the `mask/filter`, this works in most cases, as it often contains `false/unset` bits, but it doesn't have to.

This PR fixes the behavior and in case of null values it make sure that we first `mask & validities` before we filter.

### DIFF
--- a/src/compute/filter.rs
+++ b/src/compute/filter.rs
@@ -111,8 +111,10 @@ pub fn build_filter(filter: &BooleanArray) -> Result<Filter> {
 }
 
 /// Filters an [Array], returning elements matching the filter (i.e. where the values are true).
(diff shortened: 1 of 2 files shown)


jorgecarleitao:
We will need to check that this is consistent for all cases and update the docs accordingly (there is a big "WARNING" on the docs atm about this)

ritchie46:
> We will need to check that this is consistent for all cases and update the docs accordingly (there is a big "WARNING" on the docs atm about this)

What cases do you mean? The validity can only have two states, those are added in the test.

I will update the docs, missed that UB, and just passed filters with null values. xD.

PR #655, opened 2021-12-02
Avoid double utf8 checks on MutableUtf8 -> Utf8

jorgecarleitao:


ritchie46:
Was it called often? :grin:

PR #658, opened 2021-12-04
Added example to read parquet in parallel with rayon

jorgecarleitao:
This shaves a lot of time when reading many columns. The tradeoff, as usual, is that it loads all pages of a given row group into memory, to be parallelized via `par_iter`.

There is a middle ground here were we can use Rayon's par_bridge, but at that point we are mixing IO and CPU-bounded tasks, which for network IO / s3 makes things more difficult.


jorgecarleitao:
cc @ritchie46 , I think this is how polars can do it. ^_^

ritchie46:
Awesome! Thanks for this! :)

PR #670, opened 2021-12-09
Improved performance of reading utf8 required from parquet (-15%)

### LINKED ISSUE #666: reading of parquet binary type, is very slow:
I find [ https://github.com/jorgecarleitao/arrow2/blob/main/src/io/parquet/read/binary/mod.rs#L34]
capacity cannot be pre allocated here，Cause ‘extend_from_Slice’ is very slow。
image

jorgecarleitao:
Closes  #666

Two optimizations:
* pre-allocate
* inline reading u32 in little endian (parquet2's `get_length` is not inlined and causes some perf problems)

```
read required utf8 2^20 time:   [8.0038 ms 8.0496 ms 8.0977 ms]                                    
                        change: [-17.183% -16.507% -15.848%] (p = 0.00 < 0.05)
```

Thanks @ldn9638 for the idea!

### DIFF
--- a/benches/read_parquet.rs
+++ b/benches/read_parquet.rs
@@ -6,16 +6,23 @@ use criterion::{criterion_group, criterion_main, Criterion};
 use arrow2::error::Result;
 use arrow2::io::parquet::read;
 
-fn to_buffer(size: usize, dict: bool, multi_page: bool, compressed: bool) -> Vec<u8> {
+fn to_buffer(
+    size: usize,
+    nullable: bool,
+    dict: bool,
+    multi_page: bool,
+    compressed: bool,
+) -> Vec<u8> {
     let dir = env!("CARGO_MANIFEST_DIR");
 
     let dict = if dict { "dict/" } else { "" };
     let multi_page = if multi_page { "multi/" } else { "" };
     let compressed = if compressed { "snappy/" } else { "" };
+    let nullable = if nullable { "" } else { "_required" };
 
     let path = PathBuf::from(dir).join(format!(
-        "fixtures/pyarrow3/v1/{}{}{}benches_{}.parquet",
-        dict, multi_page, compressed, size
+        "fixtures/pyarrow3/v1/{}{}{}benches{}_{}.parquet",
+        dict, multi_page, compressed, nullable, size
     ));
 
     let metadata = fs::metadata(&path).expect("unable to read metadata");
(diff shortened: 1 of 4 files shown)


ldn9638:
Very cool !
But I still have a idea,For variable-length binary types. The current memory structure is 
`{size,data|size,data|size,data|size,data|.......}`
if modified to 
`{size_buf_size|size,size,size|data,data,data}.`

Will it be more efficient when reading? It can avoid `extend_from_slice` in BinaryIter

ldn9638:
The main reason is that if the row_group contains a lot of data_pages, it will continue to be very slow when the `page_stream_to_array` constructs   array.

jorgecarleitao:
The `[size][values]...[size][values]` comes from the PLAIN encoding from parquet. If the page is `DELTA_LENGTH_BYTE_ARRAY`, it will have the format that you mention (see also [ARROW-13388](https://issues.apache.org/jira/browse/ARROW-13388))

In general a column chunk should have a small number of pages (usually 2-3 for a dict-encoded and 1 for a non-dict encoded. More pages means more fragmentation and thus less compression. Since many parquet writers do not write or support filter pushdown at the page level, more fragmentation does not help in skipping pages.

> it will continue to be very slow when the page_stream_to_array constructs array.

I _think_ that this is the fastest possible under the constraints of the parquet format, since we can't tell the total size of a data page prior to reading its header (but I would love to be proven wrong and learn something new ^_^)

ldn9638:
Thank you for your reply.
Can I use `metadata. uncompressed_ size() - metadata. num_values() * 4` is the initialization capacity of the values container?

jorgecarleitao:
> Can I use `metadata. uncompressed_ size() - metadata. num_values() * 4` is the initialization capacity of the values container?

I think that the following holds:

```metadata.uncompressed_size() = sum_{pages} page.header + page.buffer```

pages may contain a dictionary page (we only know once we read the page), and individual pages may be encoded differently (e.g. mixing `PLAIN` with `DELTA_LENGTH_BYTE_ARRAY`) -- `metadata.uncompressed_size` will vary accordingly. So, it is a bit difficult to make a general statement for optimizations.

I think that the `DELTA_LENGTH_BYTE_ARRAY` encoding is the solution for this; it has a better compression ratio and is faster to read and write to and from arrow.

for reference, I took the equation above by looking at the writer, [here](https://github.com/jorgecarleitao/parquet2/blob/main/src/write/column_chunk.rs#L107) and [here](https://github.com/jorgecarleitao/parquet2/blob/main/src/write/column_chunk.rs#L176).

ldn9638:
Thanks for your reply. I'll study again

PR #675, opened 2021-12-11
Made IPC ZSTD-compressed consumable by pyarrow

jorgecarleitao:
I do not understand why the level must be zstd's default to work, but it works. This is either a bug on zstd's crate or on pyarrow's ZSTD's compressor :/


ritchie46:
That magic number goes somewhere in metadata I assume? Now we aligned with pyarrows 'level'?

jorgecarleitao:
Yes, the thing is that the "level" is just a number that tradeoffs performance for compression, so the fact that only one level works is a bit worrisome imo - the ability to read and write ZSTD-compressed data should be independent of the level. This PR just matches the level, but why other numbers do not work is unclear to me.

ritchie46:
Hmm.. a bit hard to test whats at fault here. As there are not many ipc implementations. xD

jorgecarleitao:
exactly! xD The evidence (prior to this PR) is:

* pyarrow reads what pyarrow writes
* pyarrow does not read what arrow2 writes
* arrow2 reads what pyarrow writes
* arrow2 reads what arrow2 writes

PR #690, opened 2021-12-17
Added support to write to Avro

jorgecarleitao:
This PR adds minimal support to write to [Apache Avro](https://en.wikipedia.org/wiki/Apache_Avro).

Things not yet supported: 
* many types
* compression


jorgecarleitao:
If anyone would like to chip in, there are 4 pieces missing:

1. write converter of an Arrow schema to an Avro schema (inverse of `convert_schema`)
2. write compressor (inverse of `Decompressor`) that optionally re-uses buffers
3. write serializer (inverse of the `deserialize`)
4. write the inverse of `BlockStreamIterator` (so, something that writes blocks of data to a `Write` object according to Avro's layout)

they can be written fairly indepedendently and are essentially the "inverse" operation of the read counter-parts

PR #692, opened 2021-12-19
Migrated from `avro-rs` to `avro-schema`

jorgecarleitao:
This PR replaces `avro-rs` by `avro-schema` (authored by me).

# Background

While working on reading from Avro some months ago, and now on #690, I noticed 4 issues with `avro-rs`, some of which are blocking:

1. I was unable to create a `Schema::Union` since both its properties and the constructor are private :/
2. I identified a bug which was causing `avro-rs` to write schemas incorrectly
3. `avro-rs` drags dependencies and code to deserialize from Avro, but we do not use it since we implement deserializers directly to arrow
4. `avro-rs` is unmaintained ([its new home](https://github.com/apache/avro/tree/master/lang/rust) seems so as well), and some of my PRs there are not being merged or dragged upstream

# This PR

Given the above, I wrote a new crate, `avro-schema`, containing the declaration of the Avro schema (as per spec) in Rust. This hopefully offers more flexibility and unblocks #690

The other added bonus are:
1. The create only depends on `serde` and `serde_json`
2. The crate is quite small (<800 LOC) and `#![forbid(unsafe_code)]`
3. addresses all the issues mentioned above

### DIFF
--- a/tests/it/io/avro/read.rs
+++ b/tests/it/io/avro/read.rs
@@ -9,7 +9,6 @@ use arrow2::datatypes::*;
 use arrow2::error::Result;
 use arrow2::io::avro::read;
 use arrow2::record_batch::RecordBatch;
-use arrow2::types::months_days_ns;
 
 fn schema() -> (AvroSchema, Schema) {
     let raw_schema = r#"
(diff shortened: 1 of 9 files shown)


tests/it/io/avro/read.rs
```diff
@@ -41,13 +40,6 @@ fn schema() -> (AvroSchema, Schema) {
                 "type": "enum",
                 "name": "",
                 "symbols" : ["SPADES", "HEARTS", "DIAMONDS", "CLUBS"]
-            }},
-            {"name": "duration",
-             "logicalType": "duration",
```
jorgecarleitao, line 46: this is deleted because this is an invalid schema (which `avro-rs` consumes and incorrectly writes an avro file with).

PR #693, opened 2021-12-20
Replaced `MutableBuffer` by `std::Vec`

jorgecarleitao:
This PR replaces

* `cache_aligned` cargo feature
* the custom implementation of `Vec` under the feature gate
* the custom allocator
* `MutableBuffer`

by Rust's `std::vec::Vec`.

We still need `Bytes` for the ffi, but this significantly simplifies the APIs for Rust users.

To migrate:

1. replace `Buffer::from(slice)` by `Buffer::from_slice(slice)` (`from` is now from `Vec`)
2. replace `.from_trusted_len_iter` and the like by `.extend` (since the std lib supports specialization and will pick the trusted len implementation by default)
3. replace `MutableBuffer` by `Vec`
4. replace `a.extend_constant(additional, value)` by `a.resize(a.len() + additional, value)`


ritchie46:
This is awesome! <3

jorgecarleitao:
@sundy-li , would this be ok for you?

sundy-li:
That's ok since they have similar pub methods.

jorgecarleitao:
`MutableBuffer` was needed due to the custom allocator aligned to cache lines, under the feature flag `cache_aligned`. This PR proposes removing the custom allocator (and consequently the `MutableBuffer`). I never saw any difference between them, and using `Vec` is safer and much lower maintenance (imo).

PR #694, opened 2021-12-20
Replaced `Display` by `Debug` for `Array`

jorgecarleitao:
Currently representing arrays in `Debug` is a bit verbose. This was useful to iterate over the struct definitions, but as the API stabilizes, we can move to less verbose representations.

This PR goes in this direction by removing the `impl Display` of the Array and by making the `Debug` one equal to the current `Display` one. This allows users to impl `Debug` on structs that contain arrays and have an easier to read string representation.

This follows the same design used by `std::Vec`, which does not present the struct details to the users.

This also removes the `Display` implementation to the datatypes, since it was equal to `Debug` (and there is no natural Display for them in Arrow).

### DIFF
--- a/examples/csv_read_async.rs
+++ b/examples/csv_read_async.rs
@@ -29,6 +29,6 @@ async fn main() -> Result<()> {
         0,
         deserialize_column,
     )?;
-    println!("{}", batch.column(0));
+    println!("{:?}", batch.column(0));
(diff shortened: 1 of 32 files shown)


jorgecarleitao:
A side product of this is that the `assert_eq` of two arrays is much easier to read.

PR #707, opened 2021-12-24
Add `and_scalar` and `or_scalar` for boolean

silathdiir:
Close https://github.com/jorgecarleitao/arrow2/issues/663

### DIFF
--- a/src/compute/boolean.rs
+++ b/src/compute/boolean.rs
@@ -3,11 +3,16 @@ use crate::array::{Array, BooleanArray};
 use crate::bitmap::{Bitmap, MutableBitmap};
 use crate::datatypes::DataType;
 use crate::error::{ArrowError, Result};
+use crate::scalar::BooleanScalar;
 
 use super::utils::combine_validities;
 
-/// Helper function to implement binary kernels
-fn binary_boolean_kernel<F>(lhs: &BooleanArray, rhs: &BooleanArray, op: F) -> Result<BooleanArray>
+/// Helper function to implement binary kernels for two boolean arrays
+fn binary_boolean_arrays_kernel<F>(
+    lhs: &BooleanArray,
+    rhs: &BooleanArray,
+    op: F,
+) -> Result<BooleanArray>
 where
     F: Fn(&Bitmap, &Bitmap) -> Bitmap,
 {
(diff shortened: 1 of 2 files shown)


jorgecarleitao (commented):
Thank you so much for this! The amount of testing that is here is really impressive.

I think that there is an optimization here: 

* When the scalar is `true`, `A & true = A` and `A | true = A`. 
* When the scalar is `false`, `A & false = false` and `A | false =  A`

i.e. we do not need to compute this item by item; we just need to clone and/or initialize all items as `false` / `unset` accordingly. This should be orders of magnitude faster since it is not `O(N)`.

What do you think?

silathdiir:
> I think that there is an optimization here:
> 
> * When the scalar is `true`, `A & true = A` and `A | true = A`.
> * When the scalar is `false`, `A & false = false` and `A | false =  A`
> 
> i.e. we do not need to compute this item by item; we just need to clone and/or initialize all items as `false` / `unset` accordingly. This should be orders of magnitude faster since it is not `O(N)`.
> 
> What do you think?

Yes. That should be great. I will update the code. Thanks.

src/compute/boolean.rs
```diff
@@ -133,3 +134,53 @@ pub fn is_not_null(input: &dyn Array) -> BooleanArray {
...
+/// let scalar = BooleanScalar::new(Some(true));
+/// let result = and_scalar(&array, &scalar);
+/// assert_eq!(result, BooleanArray::from_slice(&[false, false, true, true]));
+/// # }
+/// ```
+pub fn and_scalar(array: &BooleanArray, scalar: &BooleanScalar) -> BooleanArray {
+    match scalar.value() {
+        Some(true) => array.clone(),
+        Some(false) => {
+            let values = Bitmap::from_trusted_len_iter(std::iter::repeat(false).take(array.len()));
```
jorgecarleitao, line 156:
```suggestion
            let values = Bitmap::from_len_zeroed(array.len());
```

(`from_len_zeroed` uses the equivalent of C's `calloc`, while `from_trusted_len_iter` uses `malloc` followed by setting bit by bit, which is far more expensive ;) )

src/compute/boolean.rs
```diff
@@ -133,3 +134,53 @@ pub fn is_not_null(input: &dyn Array) -> BooleanArray {
...
+/// let array = BooleanArray::from_slice(&[false, false, true, true]);
+/// let scalar = BooleanScalar::new(Some(true));
+/// let result = or_scalar(&array, &scalar);
+/// assert_eq!(result, BooleanArray::from_slice(&[true, true, true, true]));
+/// # }
+/// ```
+pub fn or_scalar(array: &BooleanArray, scalar: &BooleanScalar) -> BooleanArray {
+    match scalar.value() {
+        Some(true) => {
+            let values = Bitmap::from_trusted_len_iter(std::iter::repeat(true).take(array.len()));
```
jorgecarleitao, line 180:
```suggestion
            let values = Bitmap::from_len_zeroed(array.len());
```

jorgecarleitao (commented):
Great one. Left two minor suggestions :)

src/compute/boolean.rs
silathdiir: Got it. I will update it with `let values = Bitmap::new_zeroed(array.len());`. Thanks.

src/compute/boolean.rs
silathdiir: For this `or` version, I want to initialize a Bitmap with all `true` bits to implement `A | true = true`. Is it right?

jorgecarleitao: Oh, yes, you are right!

Unfortunately we do not have a specific method for that, but it is worth checking `MutableBitmap::extend_constant` here, since it sets 8 bits per write, still much faster than `from_trusted_len_iter` ^_^

jorgecarleitao: For context, the reason we have all these seemly similar APIs is that for a single API (like rust's `Vec::extend`) we need trait specialization in the stable channel. This is why we currently name every method differently.

silathdiir: Thanks for instruction. I updated to use `MutableBitmap:: extend_constant`. Thanks.

jorgecarleitao:
Merged. Thanks a lot, @silathdiir!

PR #710, opened 2021-12-25
Allow to create primitive array by vec without extra memcopy

sundy-li:
`from_vec` is a faster way to create `primitive array` than `from_slice`.


jorgecarleitao:
Thanks a lot!

Issue #581, opened 2021-11-06
Decouple IPC-specific information from `Schema`


jorgecarleitao:
# Background

The schema declared in the Arrow IPC specification concerns aspects related to sharing of data under IPC. However, when the data is in memory, not all parts of the schema are necessary. Two important bits are:

* endianess information
* `dict_id`

As a user of the crate, I do not care about the schema endianess because my architecture has set it and the crate has already decoded values according to the incoming endianess (swap bytes as needed) in the IPC boundary. Likewise, `dict_id` is something used to keep track of dictionaries throughout the IPC file / stream, but is not necessary when the data has been loaded in memory.

The proof of this observation is that the c data interface do not have `dict_id` nor endianess - other mechanisms are used to ensure lossless roundtrips.

The existence of `dict_id` in `Field` makes it difficult to offer a good UX to use dictionaries effectively over the IPC. The gist is that `dict_id` is expected to be a number uniquely representing a dictionary array.

Currently, users are required to pass a `dict_id` for dictionary fields. However, setting this number right requires managing global state, as we need to track how many dictionaries were created and which ids were assigned to them. As a corollary, most people disregard that number, rendering transffering dictionaries via IPC useless and/or wrong, since settings the same id for two different dictionaries may lead to a "dictionary replacement" in the IPC.

`dict_id` is only used when writing to IPC, when building the relationships between dictionaries and corresponding index arrays. Likewise, endianess information is only necessary to declare to the IPC which endianess the file has been written as.

This issue is also tied with why it is currently difficult to support nested Dictionaries (#499) and some back and forth in what datatypes must contain to ensure loseless roundtrip over the IPC (e.g. #501 and #439)

# Goals

The goal of this issue is to investigate declaring a separate schema `IpcSchema, IpcDataType, IpcField` specifically for IPC, and only expose this schema when handling the Ipc format. Offer a `From` implementation that allows moving from `Schema` to `IpcSchema` and vice-versa. This would be very similar to what happens in parquet, where parquet has its own schema that we map to arrow. The core difference is that `IpcSchema <-> Schema` is lossless since by design this crate adopts all and only the logical types declared in arrow.

This allow us to remove `dict_id` from the fields, but, more importantly, allows offering the ability to write dictionary arrays effectively over the IPC boundary without user effort (by performing the calculation of the `dict_id` on the IPC boundary instead of by the user).

jorgecarleitao:
Closed by #713

Issue #673, opened 2021-12-11
Consider removing `RecordBatch`


jorgecarleitao:
For historical reasons, we have `RecordBatch`. `RecordBatch` represents a collection of columns with a schema. 

I see a couple of problems with `RecordBatch`:

1. it mixes metadata (`Schema`) with data (`Array`). In all IO cases we have, the `Schema` is known when the metadata from the file is read, way before data is read. I.e. the user has access to the `Schema` very early, and does not really need to pass it to an iterator or stream of data for the stream to contain the metadata. However, it is required to do so by our APIs, because our APIs currently return a `RecordBatch` (and thus need a schema on them) even though all the schemas are the same.

2. it is not part of the arrow spec. A RecordBatch is only mentioned in the IPC, and it does not contain a schema (only columns)

3. it is a struct that can easily be recreated by users that need it

4. It indirectly drives design decisions to use it as the data carrier, even though it is not a good one. For example, in DataFusion (apache/arrow-datafusion) the physical nodes return a stream of `RecordBatch`, which requires piping schemas all the way to the physical nodes so that they can in turn use them to create a RecordBatch. This could have been replaced by `Vec<Arc<dyn Array>>`, or even more exotic carriers (e.g. an enum with a scalar and vector variants).

houqp:
:+1: datafusion is already considering rolling its own enum based record batch abstraction. I also think it's a waste to clone and pass the same schema over and over again through out the code base.

jorgecarleitao:
@sundy-li  @ritchie46 , does any of you use the schemas on each of the batches coming from arrow2?

ritchie46:
Nothing we cannot refactor.  I think its a good idea. 👍

sundy-li:
I do agree it's better to remove the scheme inside the batch.

So there will be a better name `type Chunk = Vec<Arc<dyn Array>>` ?

ritchie46:
> So there will be a better name type Chunk = Vec<Arc<dyn Array>> ?

`ChunkedArrays` are vertical in pyarrow and polars, so that might be confusing.

'ArrayGroup`?

sundy-li:
That's ok, `chunk` is from some famous database naming style, we can still have our own name for that.

 Chunk is a list of columns with the same length
 
TIDB: https://github.com/pingcap/tidb/blob/master/util/chunk/chunk.go#L36-L50

[...]

PR #713, opened 2021-12-26
Moved dict_id to IPC-specific IO

jorgecarleitao:
This removes the `dict_id` from `Field`. It is a major simplification to the crate's design.

# Background

The arrow format supports dictionaries. When writing to IPC, each dictionary is required to be assigned an id. Dictionaries are written as separate (dictionary) messages and identified in a IPC recordbatch message via the id stored in the IPC message's schema.

Currently, dictionary ids are stored in `Field`. This is quite limiting for the following reasons:

## `dict_id` is not a logical construct

A dictionary id is encoding-related information specific for the IPC format (e.g. the C data interface does not have such a concept). However, it currently resides in `Field`, a logical construct used to track names, logical types and nullability (general schema information).

## `dict_id` must be declared at logical planning

Users currently have no mechanism to declare a dictionary to be re-used in the IPC during execution - since the id needs to be set on the `Field`, it means that logical planning needs to assign ids during logical planning, even before any access to the data. This is a major design limitation because it can happen that the specifics of the execution can _cause_ the dictionaries to be reusable - this information is often not available during logical planning.

For example, one implementation of single-batch "id assignment" is to look for all dictionary values in a batch and increment by 1 iff the pointer of the values is new over all others (thus avoiding serializing the values twice). That two columns are re-using a dictionary is only known once we have access to the dictionaries themselves, during execution.

Likewise, an implementation for multi-batch "id assignment" is to hash the dictionary values and have a hashmap between ids and hashes, and only emit dictionaries to the IPC when we find a new hash (which, again, is only known at execution time).

# This PR

This PR refactors the `dict_id` to a separate struct, `io::ipc::IpcField`, that tracks the dictionary ids of the fields from and to the IPC format. This tracking is only needed when reading from and writing to IPC, and is now part of the "metadata" that we store when reading from IPC.

When writing a batch to IPC, users are now required to provide a `io::ipc::IpcField`, which is used to describe to this crate which dictionaries are being re-used or not. We also offer a minimal implementation to generate these fields by assigning an incremental id per dictionary in the batch, `io::ipc::write::default_ipc_fields`.


jorgecarleitao:
This closes #581 and is a pre-requisite to address #673

jorgecarleitao:
cc @alamb , this was another simplification of the Field that allows re-using dictionary-encoding efficiently on Arrow IPC interface.

PR #717, opened 2021-12-28
Replaced `RecordBatch` by `Chunk`

### LINKED ISSUE #673: Consider removing `RecordBatch`

jorgecarleitao:
closed #673

This is a major refactor to the crates' IO interfaces, see #673 for details.

This PR:

Replaces `RecordBatch` by a new struct, `Chunk`, containing a vec of arrays with the same length. All IO interfaces now use `Chunk` and behave as follows:

* read or infer schema (logical plane)
* read columns (physical plane)

This allows users to not have to "leak" logical information to the physical plane unless necessary by the format.

All IO APIs were refactored to read and write `Chunk` (instead of `RecordBatch`). This removes much of the boilerplate to write a file.

# Migration path

* `RecordBatch` -> `Chunk<Arc<dyn Array>>`
* `RecordBatch::num_rows()` -> `Chunk::len()`
* `RecordBatch::columns()` -> `Chunk::columns()`
* `RecordBatch::column(i)` -> `Chunk::columns()[i]`
* `RecordBatch::num_columns()` -> `Chunk::columns().len()`
* `RecordBatch::schema()` -> no longer present. Use other APIs (usually metadata) to get the schema


jorgecarleitao:
Renamed to `Chunk` based on @sundy-li 's suggestion: https://github.com/jorgecarleitao/arrow2/issues/673#issuecomment-991845103

jorgecarleitao:
@yjshen , @houqp , could you take a look at this?

I envision some pain with this PR in datafusion, as datafusion currently passes logical information (`Schema`) down to the physical nodes.

Because this PR requires _less_ information to write, one way to go is to declare in DataFusion

```rust
pub struct RecordBatch {
     pub columns: Chunk<Arc<dyn Array>>;
     pub schema: Arc<Schema>;
```
and pass `columns` to the interfaces (this schema is now useless from arrow2's perspective, since the schema is known before the first batch is available).

For reading, likewise the schema is always known prior to start reading the first batch. Thus, we can just store an `Arc<Schema>` after reading the metadata/infering the schema and clone it for every batch that comes from IO.

yjshen:
The new Chunk API LGTM.

PR #718, opened 2021-12-28
clarify boolean value in DataType::Dictionary

ritchie46:
Same for `DataType::Map`?


jorgecarleitao:
yes, same for map, but there is just the keys I think.

PR #726, opened 2022-01-03
add ordering

Igosuki:
Hi guys, I'm currently trying out arrow2 with the latest master from datafusion.  The following updates were necessary because TypeSignature is ordered.

P.S. : The other commits are just me removing code only needed to add Decimal support.

### DIFF
--- a/src/array/fixed_size_binary/mutable.rs
+++ b/src/array/fixed_size_binary/mutable.rs
@@ -162,15 +162,6 @@ impl MutableFixedSizeBinaryArray {
             validity.shrink_to_fit()
         }
     }
-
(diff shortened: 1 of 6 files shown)


jorgecarleitao:
Thanks for this PR. Could you check https://github.com/jorgecarleitao/arrow2/pull/379?

Igosuki:
Ok, so the requirement for PartialOrd is lifted so we can close this !

jorgecarleitao:
Thanks @Igosuki , and also for all your work in the datafusion migration 🙇

Igosuki:
👍

PR #727, opened 2022-01-03
Avro arrays can be null

Igosuki:
This helps with deserializing the schema Union([Array, Null]) or Union([Null, Array]) which is a valid avro schema (and the default for sparks list arrays for instance).

### DIFF
--- a/src/io/avro/read/deserialize.rs
+++ b/src/io/avro/read/deserialize.rs
@@ -96,10 +96,15 @@ fn deserialize_value<'a>(
     let data_type = array.data_type();
     match data_type {
         DataType::List(inner) => {
-            let avro_inner = if let AvroSchema::Array(inner) = avro_field {
-                inner.as_ref()
-            } else {
-                unreachable!()
+            let avro_inner = match avro_field {
+                AvroSchema::Array(inner) => inner.as_ref(),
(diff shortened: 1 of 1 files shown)


jorgecarleitao:
Looks great. Thanks! Do you have a recommendation on how we could test this? (we do not need for this PR, but I am curious). E.g. we have a procedure to test that our parquet files are readable by spark, via pyspark. Could we use pyspark to write a bunch of avro files and test against that?

Igosuki:
@jorgecarleitao I test it using the avro files at https://github.com/apache/arrow-testing/tree/a8f7be380531758eb7962542a5eb020d8795aa20/data/avro (they are copy of the parquet files).
You're right that actually doing a round trip with arrow and pyspark would be a great integration test... But in the meantime, the test mod here is enough to test this PR : https://github.com/Igosuki/arrow-datafusion/blob/arrow22r/datafusion/src/avro_to_arrow/arrow_array_reader.rs happy to port the code here

jorgecarleitao:
Sounds good.

I have added #733 as a follow-up. Generally I prefer to run integrations with third parties against code (instead of files), so that we can also know (and reproduce) how the file was generated in the first place.

Could you fix the fmt so we can merge it green?

Igosuki:
Done

Issue #725, opened 2022-01-02
Consider switching flatbuffers' dependency by `planus`


jorgecarleitao:
flatbuffers (the rust implementation, not the format) has been giving us quite an amount of maintenance burden due to its usage of unsafe and general un-maintainanability (see some of the issues on the rust-advisory list as well as https://github.com/google/flatbuffers/issues).

This issue tracks investigations to migrate to [`planus`](https://crates.io/crates/planus). It was recently released and addresses the aformentioned issues with a simpler and easier to follow architecture.

If such switch happens, it is expected to be backward compatible, incur no performance cost, and will make this crate safer.

* https://github.com/DataEngineeringLabs/arrow-format/tree/planus tracks this movement on the arrow-format crate,
* https://github.com/jorgecarleitao/arrow2/tree/planus tracks this movement here

The current status is that I was able to generate the flatbuffers and have it compile, and I am now evaluating the changes required on this end of things (against that branch).

I am aiming at having all tests pass under this change. We can then benchmark the differences and make a decision based on data (maintenance vs performance vs security).

cc @TethysSvensson

jorgecarleitao:
Closed by #732

PR #732, opened 2022-01-04
Replace flatbuffers dependency by Planus

jorgecarleitao:
This PR will stay here for some time. It replaces the flatbuffers crate by planus, a recently released replacement to flatbuffers. See #725 for details.

This switches one for the other. All tests and integration tests pass, showing that it does indeed work as advertised 💯

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -10,8 +10,4 @@ ignore = [
     # Therefore, this advisory does not affect us.
     "RUSTSEC-2020-0071",
     "RUSTSEC-2020-0159", # same as previous
-
-    # this cannot be addressed, only mitigated.
-    # See [.github/workflows/security.yml] for details on how we mitigate this.
-    "RUSTSEC-2021-0122",
 ]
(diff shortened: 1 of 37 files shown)


jorgecarleitao:
Added MIRI check to a round-trip over IPC. All still green.

jorgecarleitao:
I though that this would take longer, but it is essentially done: tests pass, MIRI checks pass, simpler to use and a safer design.

One step missing is benchmarks of read and write, to check if there is no major regression.

@sundy-li , @ritchie46 , @houqp, @yjshen  , any thoughts on this?

My opinion is that flatbuffers is unmaintained and has soundness issues, which is a concern because IO is an obvious attack vector - it does not mean that `planus` won't become unmaintained (maybe @TethysSvensson or @kristoff3r could chip in here wrt to their plans?).

TethysSvensson:
@jorgecarleitao While I am not going to give any guarantees (this is open source after all), my current plan is to maintain and improve upon the library for at least the next year, since we are planning to use the library in product at [skybox.gg](https://skybox.gg).

What happens after that is too far into the future for me to predict and also depends on whether there is any community adoption.

Issue #738, opened 2022-01-07
Boolean kernels: `any` / `all`


ritchie46:
It would be nice to have fast columnwise `any`/ `all` operations.

This has probably some nice optimization options. Maybe reading the bits per `u8` or `u64` and check if `== 0` / `== usize::max`.

jorgecarleitao:
Closed by #739

PR #739, opened 2022-01-07
implement any kernel

ritchie46:
First part for #738. I'll continue with the `all` after first review.

### DIFF
--- a/src/compute/boolean.rs
+++ b/src/compute/boolean.rs
@@ -194,30 +194,16 @@ pub fn any(array: &BooleanArray) -> bool {
         array.into_iter().any(|v| v == Some(true))
     } else {
         let vals = array.values();
-        let (mut bytes, start, mut len) = vals.as_slice();
-        if start != 0 {
-            if array.values_iter().take(8 - start).any(|v| v) {
-                return true;
-            }
-            bytes = &bytes[1..];
-            len -= start;
-        }
-        // remaining part of last byte
-        let remainder = len % 8;
-        if remainder != 0 {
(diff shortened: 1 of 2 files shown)


src/compute/boolean.rs
```diff
@@ -185,3 +185,39 @@ pub fn or_scalar(array: &BooleanArray, scalar: &BooleanScalar) -> BooleanArray {
...
 }
+
+/// Check if any of the values in the array is `true`
+pub fn any(array: &BooleanArray) -> bool {
+    if array.is_empty() {
+        false
+    } else if array.validity().is_some() {
+        array.into_iter().any(|v| v == Some(true))
+    } else {
+        let vals = array.values();
```
jorgecarleitao, line 196: Here, we need to compute if any entry is set. isn't that equivalent to `array.values().null_count() > 0`?

ritchie46: Indeed. But this function returns on first `true`  value that is found, whereas `null_count` needs to scan the whole array.

jorgecarleitao: `null_count` is cached (yes, we also cache it for the values of the `BooleanArray`)

ritchie46: For every bitmap? Ooooh.. then that is definitely a fast path. :)

ritchie46:
Used the cached null counts and added the `all` kernel. Good to go on my part.

jorgecarleitao:
Thanks a lot, @ritchie46 !

PR #740, opened 2022-01-07
Added integration tests for writing compressed parquet

jorgecarleitao:
Showing that our written compressed parquet files are read by pyarrow and (py)spark.

pyspark does not ship zstd codecs, so we can't test that in this setup (we only test snappy).


jorgecarleitao:
related to https://github.com/jorgecarleitao/parquet2/issues/75

PR #743, opened 2022-01-08
Parquet: add as_any and null_count to dyn Statistics

ritchie46:
This adds the `as_any` method on `dyn Statistics` for parquet reading. I also added `null_count` similar to how `dyn Statistics` is defined in `parquet2`. 

I think the `distinct_count` would also be valuable on this trait.


jorgecarleitao (approved):
Thanks! (waiting for green)

jorgecarleitao:
fwiw, I haven't seen a parquet file with a populated distinct count xD

PR #749, opened 2022-01-10
Made IO `#[forbid(unsafe)]`

jorgecarleitao:
This PR adds the `bytemuck` dependency and replaces all instances of `unsafe` in the IO module by `bytemuck`'s API, also adding `#[forbid(unsafe)]` to make it more explicit that our IO has no `unsafe` on it.

It has no performance implications, it just formalizes (in traits) our use of plain old data (Pod) as defined in `bytemuck`.

Thanks to the wg secure code and in particular to @Shnatsel and @tarcieri for the recommendation that led to this PR, and also to the folks at [`#dark-arts`](https://discord.com/channels/273534239310479360/592856094527848449) for confirming that the two implementations of `Pod` in this PR are sound.

### DIFF
--- a/src/types/native.rs
+++ b/src/types/native.rs
@@ -1,6 +1,8 @@
 use std::convert::TryFrom;
 use std::ops::Neg;
 
+use bytemuck::{Pod, Zeroable};
+
 use super::PrimitiveType;
 
 /// Sealed trait implemented by all physical types that can be allocated,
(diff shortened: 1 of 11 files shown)


src/types/native.rs
```diff
@@ -108,6 +110,9 @@ impl days_ms {
     }
 }
 
+unsafe impl Pod for days_ms {}
+unsafe impl bytemuck::Zeroable for days_ms {}
```
Shnatsel, line 114: There is a `#[derive]` macro to implement this without any custom unsafe code. I suggest adding a comment on why that wasn't used here, and also moving the `unsafe impl`s closer to the struct definition to make then harder to miss when making changes to the type.

jorgecarleitao: Thanks! even better. Replaced by `derive`.

PR #751, opened 2022-01-11
Reduce memory usage in Parquet->Arrow decimal column chunk conversion

danburkert:
This PR reduces memory usage, both in terms of memory used as well as
allocations, in the Parquet->Arrow conversion of Decimal chunks. There
are two optimizations:

1. Instead of using `slice::concat` to expand buffers to 16 bytes, a
  stack-allocated 16 byte buffer is used instead. This removes an
  allocation per value.
2. Data is expanded from the encoded Parquet fixed-size binary pages
   into a byte buffer, which is then converted to a buffer of i128s. To
   reduce the size of the intermediate byte buffer, this conversion is
   now done page-by-page.

### DIFF
--- a/src/io/parquet/read/mod.rs
+++ b/src/io/parquet/read/mod.rs
@@ -4,7 +4,6 @@
 use std::{
     collections::VecDeque,
     io::{Read, Seek},
-    iter::FromIterator,
     sync::Arc,
 };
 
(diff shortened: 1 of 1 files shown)


avfeinberg (approved):

mdrach (approved):

danburkert:
For reasons I don't fully understand, optimization #2 was yielding the wrong results, so I've backed it out in the second commit and replaced it with a FromIter call.

jorgecarleitao (approved):
Thanks! Clippy missing but otherwise ready to ship

src/io/parquet/read/mod.rs
```diff
@@ -314,44 +314,33 @@ fn page_iter_to_array<I: FallibleStreamingIterator<Item = DataPage, Error = Parq
...
+                            .map(|value: Option<&[u8]>| {
+                                // Copy the fixed-size byte value to the start of a 16 byte stack
+                                // allocated buffer, then use an arithmetic right shift to fill in
+                                // MSBs, which accounts for leading 1's in negative (two's complement)
+                                // values.
+                                let mut bytes = [0u8; 16];
+                                bytes[..n].copy_from_slice(value?);
+                                Some(i128::from_be_bytes(bytes) >> 8 * (16 - n))
+                            });
+                    let i128_array = PrimitiveArray::<i128>::from_iter(i128_values);
```
jorgecarleitao, line 342:
```suggestion
                    let values =
                        fixed_size_binary_array
                            .values()
                            .iter()
                            .map(|value: &[u8]| {
                                // Copy the fixed-size byte value to the start of a 16 byte stack
                                // allocated buffer, then use an arithmetic right shift to fill in
                                // MSBs, which accounts for leading 1's in negative (two's complement)
                                // values.
                                let mut bytes = [0u8; 16];
                                bytes[..n].copy_from_slice(value);
                                i128::from_be_bytes(bytes) >> 8 * (16 - n)
                            })
                            .collect::<Vec<_>>();
                    let validity = fixed_size_binary_array.validity().cloned();
                    let i128_array = PrimitiveArray::<i128>::from_data(i128::DATA_TYPE, values.into(), validity);
```

since this operator is infallible, we can bypasse (the expensive) iteration and compute of a new validity.

src/io/parquet/read/mod.rs
```diff
@@ -314,44 +314,33 @@ fn page_iter_to_array<I: FallibleStreamingIterator<Item = DataPage, Error = Parq
...
+                    Ok(Box::new(i128_array.to(data_type)) as _)
```
jorgecarleitao, line 343: TIL `as _`, that is quite convenient!

jorgecarleitao (commented):
Noticed an extra optimization we can do here

src/io/parquet/read/mod.rs
jorgecarleitao, line 343:
```suggestion
                    Ok(Box::new(i128_array) as _)
```
(removed the `to` since we do it above)

TIL `as _`, that is quite convenient!

danburkert:
> Noticed an extra optimization we can do here

@jorgecarleitao nice, good call.  I had to insert a `chunks_exact()` in order to get that to compile, but I think that was the intention.

jorgecarleitao:
Woops, we are missing an iterator over the values of a `FixedSizeBinary`. PR here: https://github.com/jorgecarleitao/arrow2/pull/757

jorgecarleitao:
Thanks again, very clean :)

PR #747, opened 2022-01-09
Migrated to portable simd

jorgecarleitao:
This PR replaces the `packed_simd` dependency by the nightly-available `std::simd` (aka portable simd).

Closes #580

PR #773, opened 2022-01-18
Add extend_from_iter into growable trait

### LINKED ISSUE #772: Improve `Growable` by avoiding virtual function call.
I figured out that `dyn growable` runs by a slice iterator.
```
 let mut growable = growable::make_growable(&[lhs, rhs], false, lhs.len());
        let mut start_falsy = 0;
        let mut total_len = 0;
        for (start, len) in SlicesIterator::new(predicate.values()) 
  ...
}
```

what if we can put the iterator inside the `growable` trait?

```
 fn run(&mut self, iter: SlicesIterator  ) -> Box<dyn Array>;

```

sundy-li:
Did not achieve expected performance.
```
growable::primitive::non_null::non_null                                                                             
                        time:   [2.5629 us 2.5658 us 2.5692 us]

growable::dyn_primitive::non_null::non_null                                                                             
                        time:   [6.3092 us 6.3195 us 6.3328 us]
			
			
After:

growable::dyn_primitive::non_null::non_null                                                                             
                        time:   [6.0020 us 6.0147 us 6.0300 us]


```

We can't make `extend_from_iter` accept generic iterator, it'll make this trait none object-safety.

Fixes #772

### DIFF
--- a/src/array/growable/mod.rs
+++ b/src/array/growable/mod.rs
@@ -40,6 +40,13 @@ pub trait Growable<'a> {
     /// Extends this [`Growable`] with null elements, disregarding the bound arrays
     fn extend_validity(&mut self, additional: usize);
 
+    /// Extends this [`Growable`] by iterator.
+    fn extend_from_iter(&mut self, iter: Box<dyn Iterator<Item = (usize, usize, usize)>>) {
+        for (index, start, len) in iter {
+            self.extend(index, start, len);
+        }
+    }
+
     /// Converts this [`Growable`] to an [`Arc<dyn Array>`], thereby finishing the mutation.
     /// Self will be empty after such operation.
     fn as_arc(&mut self) -> std::sync::Arc<dyn Array> {
--- a/benches/growable.rs
+++ b/benches/growable.rs
@@ -19,6 +19,16 @@ fn add_benchmark(c: &mut Criterion) {
         })
     });
 
+    c.bench_function("growable::dyn_primitive::non_null::non_null", |b| {
+        b.iter(|| {
+            let mut a: Box<dyn Growable> =
+                Box::new(GrowablePrimitive::new(vec![&i32_array], false, 1026 * 10));
+
+            let iter = values.clone().into_iter().map(|start| (0, start, 10));
+            a.extend_from_iter(Box::new(iter));
+        })
+    });
+
     let i32_array = create_primitive_array::<i32>(1026 * 10, 0.0);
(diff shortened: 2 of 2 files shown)


sundy-li:
packed_simd is broken in the latest rust nightly toolchain.

Xuanwo:
BTW, packed_simd seems superseded by [stdsimd](https://github.com/rust-lang/stdsimd).

ritchie46:
> BTW, packed_simd seems superseded by [stdsimd](https://github.com/rust-lang/stdsimd).

We know. See #747. There still are some blockers on that.

src/array/growable/mod.rs
```diff
@@ -40,6 +40,13 @@ pub trait Growable<'a> {
     /// Extends this [`Growable`] with null elements, disregarding the bound arrays
     fn extend_validity(&mut self, additional: usize);
 
+    /// Extends this [`Growable`] by iterator.
+    fn extend_from_iter(&mut self, iter: Box<dyn Iterator<Item = (usize, usize, usize)>>) {
```
jorgecarleitao, line 44: Yeap, I think that this won't work as intended because `.extend` will still be dynamically dispatched.

We do something like what you intend here: https://github.com/jorgecarleitao/arrow2/blob/main/src/compute/filter.rs#L87

Essentially, specialize the implementation based on the physical type, and use a generic one (https://github.com/jorgecarleitao/arrow2/blob/main/src/compute/filter.rs#L101) for all others.

jorgecarleitao:
An idea: create a simple function `pub fn from_slice_iterator<A: AsRef<dyn Array>, I:Iterator>(&[A], iter: I) -> Box<dyn Array>` that matches the arrays' logical type, downcasts them, initializes the static-typed growable, and extend it from the iterator.

src/array/growable/mod.rs
sundy-li: https://users.rust-lang.org/t/is-this-a-curiously-recurring-template-pattern/31004/6

As User @ExpHP said:

```
trait Trait {
    fn foo(&mut self);
    fn call_foo_twice(&mut self) {
        self.foo(); // <-- this is statically dispatched
        self.foo();
    }

    fn wrap(self) -> Wrapper<Self> { // <-- we can name Self
        Wrapper(self)
    }
}
```

so `.extend ` is statically dispatched?

zirconium-n:
This should be done via extension trait to avoid object safety issue.

sundy-li:
closing to search for better ideas.

Issue #783, opened 2022-01-19
Add `FixedSizeListScalar`


jorgecarleitao:

jorgecarleitao:
Closed by #786

Issue #782, opened 2022-01-19
Add `FixedSizeBinaryScalar`


jorgecarleitao:
And plug it in the Scalar API.

jorgecarleitao:
Closed by #786

PR #786, opened 2022-01-23
Adding `FixedSizeListScalar` and `FixedSizeBinaryScalar`

illumination-k:
- Adding `FixedSizeListScalar` and `FixedSizeBinaryScalar`
- Supporting  `FixedSizeListScalar` and `FixedSizeBinaryScalar` in `new_scalar`
- Adding related test

related to #782 and #783


src/scalar/fixed_size_binary.rs
```diff
@@ -0,0 +1,52 @@
...
+
+impl FixedSizeBinaryScalar {
+    /// Returns a new [`FixedSizeBinaryScalar`].
+    /// # Panics
+    /// iff
+    /// * the `data_type` is not `FixedSizeBinary`
+    /// * the size of child binary is not equal
+    #[inline]
+    pub fn new<P: Into<Vec<u8>>>(data_type: DataType, value: Option<P>) -> Self {
+        Self {
```
jorgecarleitao, line 20:
```suggestion
        assert_eq!(data_type.to_physical_type(), crate::datatypes::PhysicalType::FixedSizeBinary);
        Self {
```

so that if the value us null, we still have a valid physical type on it.

src/scalar/fixed_size_binary.rs
```diff
@@ -0,0 +1,52 @@
...
+            value: value.map(|x| {
+                let x: Vec<u8> = x.into();
+                assert_eq!(data_type, DataType::FixedSizeBinary(x.len()));
```
jorgecarleitao, line 23:
```suggestion
                assert_eq!(data_type.to_logical_type(), DataType::FixedSizeBinary(x.len()));
```

so that if the `DataType` is an extension, we also accept it

src/scalar/fixed_size_list.rs
```diff
@@ -0,0 +1,69 @@
...
+use crate::{array::*, datatypes::DataType};
+
+use super::Scalar;
+
+/// The scalar equivalent of [`FixedSizeListArray`]. Like [`FixedSizeListArray`], this struct holds a dynamically-typed
+/// [`Array`]. The only difference is that this has only one element.
+#[derive(Debug, Clone)]
+pub struct FixedSizeListScalar {
+    values: Arc<dyn Array>,
+    is_valid: bool,
```
jorgecarleitao, line 13: What do you think about

```suggestion
    values: Option<Arc<dyn Array>>,
```

and remove `is_valid`?

jorgecarleitao (commented):
Looks awesome! I left a small number of comments / suggestions. Awesome PR, @illumination-k !

src/scalar/fixed_size_list.rs
illumination-k: I referred to `ListScalar`, but I think `Option<Arc<dyn Array>>` is better and value method should return `Option<&Arc<dyn Array>>` for consistency of other scalar.

`StructScalar` and `ListScalar` also have `is_valid`. Should we use `Option` and remove `is_valid` for them? (Maybe another PR)

jorgecarleitao:
Thanks a lot, @illumination-k !

PR #789, opened 2022-01-25
Read parquet row groups in chunks

### LINKED ISSUE #768: Allow parquet column chunks to be deserialized to multiple arrays (e.g. iterator)
As described by @tusvold [here](https://github.com/apache/arrow-datafusion/issues/1532#issuecomment-1012985001), there are important cases where it is more advantageous to deserialize a column chunk into more than one array.

Specifically, the main issue is that a decompressed parquet column chunk is still much smaller than its corresponding arrow array. Therefore, on large column chunks, the user may prefer to deserialize only part of the decompressed column chunk into arrow while still holding the chunk in memory.

The equation is something along the lines of 
```
size(decompressed parquet column chunk) + size(partial arrow array) <  size(complete chunk as arrow array)
```

To allow users to tradeoff memory usage and data fragmentation (of each array), we should offer something like

```
fn column_chunk_to_array(chunk, column_meta, length: Option<usize>) -> impl Iterator<Arc<dyn Array>>
```

that when `length: None` we recover our single array, when `length: Some(n)`, we iterate over the chunk and output `n` items at the time.

Thanks again to @tusvold for this important insight.

jorgecarleitao:
This PR allows reading of parquet columns in chunks, thereby allowing decompressing and deserializing pages on demand to reduce the memory footprint.

This is a draft as there is still some work to do, which I will continue

* [x] fix known issues in counting where we are at a given page (offset is incorrect atm)
* [x] add tests for a non-None chunk size
* [x] add back reading dictionary
* [x] add back support to read structs
* [ ] add back delta length encoding for binary
* [x] add back support for lists
* [x] triple check usage of `with_capacity` and `reserve` to avoid un-needed reallocations
* [x] bench ~~(performance should improve when the chunk size is None, there should be minimal diff with chunked)~~
* [x] DRY copy-pasted code between primitive, binary, fixed binary and boolean
* [x] remove old code

## Design

This PR follows the design of this crate:
* decompression/deserialization cannot cross column chunk boundaries, so that columns can be read (IO-bounded) and deserialized (CPU-bounded) independently and independently of each of the two operations.
* `unsafe` free
* generics to inline hot ops

The overall design of the changes enables pages to be deserialized to arrays of a different length, based on a new parameter `chunk_size: usize`:

```
parquet
[                             column chunk                       ]
[page][page][page][page][page][page][page][page][page][page][page]

arrow
[      array      ][      array      ][      array      ][ array ]
 <- chunk size  ->
```

Broadly, this PR does the following (for a single row group):

* read N column chunks to memory (`Vec<u8>`)
* for each column, compose an iterator with 2 iterator adapters:
    * `Vec<u8> -> Iterator<CompressedDataPage>`
    * `Iterator<CompressedDataPage> -> Iterator<&DataPage>` (decompression)
    * `Iterator<&DataPage> -> Iterator<Arc<dyn Array>>` (deserialization)

All these iterators are CPU-bounded. On the last step, we track where we are on the `DataPage` (through `PageState`, see below) and the temporary mutable array, and either:
* consume the page to the mutable and the rest of it to a ring of mutables when the chunk size is so small that one page contains multiple arrays, or
* iterate over more pages until we fill the temporary mutable up to `chunk_size`, at which point we freeze the mutable and return it

In more detail:
* column chunks are now read in a single `read_exact`, instead of page by page. This was a mistake on the previous design, since we should not mix reading columns from deserializing pages
* a `&'a DataPage` is mapped to a `PageState<'a>` based on its parquet physical type and encoding. This is the input state of a page and allows us to "suspend" an iteration over a page, thereby allowing a page to "extend" an array without having it completely consumed
* a new concept, `Decoder`, knows how to initialize mutable arrays and how to extend them state from `&mut PageState<'a>`, advancing the page state accordingly. This is mostly a DRY trait

Closes #768


jorgecarleitao:
This now reproduces all the existing functionality apart from:

* read structs
* read nested list (i.e. list of list)

I found a bug in `parquet2` in writing nested data while working on this. Thus, I plan to first release `parquet2` with the fix and a simpler API to write, and then move back to this one again.

jorgecarleitao:
@houqp, this is ready for a spin.

There is a regression (20-40%) when reading a whole binary/utf8 column chunk at once (i.e. no chunk size). This is related to some tricks in pre-computed capacities of binary/utf8 that benefit from reading the whole column (we can recover this behavior).

I deactivated structs of structs and lists of lists for now as I need to dig a bit into the dremel. The failing tests are just examples that I need to update. ^^

Issue #790, opened 2022-01-26
parquet -> arrow utf8-8 column conversion yields too many values


danburkert:
We're seeing an issue decoding parquet row groups containing UTF-8 string columns with many null values. Occassionally these conversions yield more values than specified in the parquet file metadata.  I traced this as far as `arrow2::io::parquet::read::binary::iter_to_array` with the problematic data set. After modifying that function to add some sanity check asserts, multiple test cases fail in the arrow2 test suite.

```rust
diff --git a/src/io/parquet/read/binary/mod.rs b/src/io/parquet/read/binary/mod.rs
index a0422f1a9..67812e2b1 100644
--- a/src/io/parquet/read/binary/mod.rs
+++ b/src/io/parquet/read/binary/mod.rs
@@ -53,6 +53,8 @@ where
             )?
         }
     }
+    debug_assert_eq!(validity.len(), capacity);
+    debug_assert_eq!(values.len(), capacity);
     Ok(utils::finish_array(data_type, values, validity))
 }
```

yields

```
---- io::parquet::write::list_large_binary_optional_v1 stdout ----
thread 'io::parquet::write::list_large_binary_optional_v1' panicked at 'assertion failed: `(left == right)`
  left: `12`,
 right: `15`', /home/ec2-user/src/rust/arrow2/src/io/parquet/read/binary/mod.rs:56:5

---- io::parquet::write::list_large_binary_optional_v2 stdout ----
thread 'io::parquet::write::list_large_binary_optional_v2' panicked at 'assertion failed: `(left == right)`
  left: `12`,
 right: `15`', /home/ec2-user/src/rust/arrow2/src/io/parquet/read/binary/mod.rs:56:5

---- io::parquet::write::list_utf8_optional_v1 stdout ----
thread 'io::parquet::write::list_utf8_optional_v1' panicked at 'assertion failed: `(left == right)`
  left: `12`,
 right: `15`', /home/ec2-user/src/rust/arrow2/src/io/parquet/read/binary/mod.rs:56:5

---- io::parquet::write::list_utf8_optional_v2 stdout ----
thread 'io::parquet::write::list_utf8_optional_v2' panicked at 'assertion failed: `(left == right)`
  left: `12`,
 right: `15`', /home/ec2-user/src/rust/arrow2/src/io/parquet/read/binary/mod.rs:56:5

---- io::parquet::write::utf8_required_v1 stdout ----
[...]

PR #791, opened 2022-01-26
Fix parquet binary dict page reading

danburkert:
As described in #790, there is a bug in parquet binary dict page decoding whereby too many values are yielded due to a mixup in the length arguments passed to various decoding functions, ultimately leading to extra bitpacked values being included in the output array.  This PR fixes the bug, and adds some `debug_assert!`s in key places to ensure that tests would catch the issue in the future.  I went ahead and added similar debug asserts to the other page type read functions, which exposed a few places where validity buffers are being allocated for non-nullable columns.

Unfortunately these new debug asserts are firing for nested page type tests, but I don't currently understand how the encoding works for these page types (it looks significantly more complicated than RLE & bitpack).  I suspect that these test failures are a true positive, but I can't be sure without understanding how nested encoding works.

### DIFF
--- a/src/io/parquet/read/binary/basic.rs
+++ b/src/io/parquet/read/binary/basic.rs
@@ -105,8 +105,6 @@ fn read_delta_optional<O: Offset>(
     values: &mut Binary<O>,
     validity: &mut MutableBitmap,
 ) {
-    let length = values.len() + additional;
-
     let Binary {
         offsets,
         values,
@@ -126,7 +124,7 @@ fn read_delta_optional<O: Offset>(
     extend_from_decoder(
         validity,
         &mut validity_iterator,
-        length,
+        additional,
         &mut Offsets::<O>(offsets),
         offsets_iterator,
     );
@@ -143,8 +141,6 @@ fn read_plain_optional<O: Offset>(
     values: &mut Binary<O>,
     validity: &mut MutableBitmap,
 ) {
-    let length = values.len() + additional;
-
     // values_buffer: first 4 bytes are len, remaining is values
     let values_iterator = utils::BinaryIter::new(values_buffer);
 
@@ -153,7 +149,7 @@ fn read_plain_optional<O: Offset>(
     extend_from_decoder(
         validity,
         &mut validity_iterator,
-        length,
+        additional,
         values,
         values_iterator,
     )
--- a/src/io/parquet/read/binary/basic.rs
+++ b/src/io/parquet/read/binary/basic.rs
@@ -189,7 +189,7 @@ pub(super) fn extend_from_page<O: Offset>(
 
     let (_, validity_buffer, values_buffer, version) = utils::split_buffer(page, descriptor);
 
-    match dbg!((&page.encoding(), page.dictionary_page(), is_optional)) {
+    match (&page.encoding(), page.dictionary_page(), is_optional) {
         (Encoding::PlainDictionary | Encoding::RleDictionary, Some(dict), true) => {
             read_dict_buffer::<O>(
                 validity_buffer,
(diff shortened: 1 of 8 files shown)


src/io/parquet/read/binary/basic.rs
```diff
@@ -192,7 +189,7 @@ pub(super) fn extend_from_page<O: Offset>(
 
     let (_, validity_buffer, values_buffer, version) = utils::split_buffer(page, descriptor);
 
-    match (&page.encoding(), page.dictionary_page(), is_optional) {
+    match dbg!((&page.encoding(), page.dictionary_page(), is_optional)) {
```
danburkert, line 192:
```suggestion
    match (&page.encoding(), page.dictionary_page(), is_optional) {
```

src/io/parquet/read/utils.rs
```diff
@@ -117,16 +117,13 @@ pub(super) fn extend_from_decoder<'a, T: Default, C: Pushable<T>, I: Iterator<It
...
     for run in decoder {
         match run {
             hybrid_rle::HybridEncoded::Bitpacked(pack) => {
                 // compute the length of the pack
                 let pack_size = pack.len() * 8;
-                let pack_remaining = page_length;
-                let length = std::cmp::min(pack_size, pack_remaining);
-
-                let additional = remaining.min(length);
+                let additional = pack_size.min(remaining);
```
danburkert, line 126: Using the full page length instead of the remaining length is one of the bugs.

src/io/parquet/read/binary/basic.rs
```diff
@@ -46,16 +46,14 @@ fn read_dict_buffer<O: Offset>(
...
-
     let values_iterator = values_iter(indices_buffer, dict, additional);
 
     let mut validity_iterator = hybrid_rle::Decoder::new(validity_buffer, 1);
 
     extend_from_decoder(
         validity,
         &mut validity_iterator,
-        length,
+        additional,
```
danburkert, line 56: parameter mixup; extend_from_decoder expects the additional length, but this was passing total length.

tustvold:
FWIW on the surface sounds similar to https://github.com/apache/arrow-rs/issues/1111

jorgecarleitao:
Great fix! Thanks!

My understanding is that the `column.num_values()` can't be reliably be used to estimate the capacity for nested types. imo It is a consequence of an ambiguous spec.

This has shown up some times:

* pyarrow has a bug related to this: https://issues.apache.org/jira/browse/PARQUET-2066
* this description is too ambiguous: https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift#L733
* mailing list discussing a related idea: https://www.mail-archive.com/dev@parquet.apache.org/msg16032.html (for pages, but the idea is the same)

My understanding is that for nested types `num_values` is 15 for the following list array:

```
[[0, 1], None, [2, None, 3], [4, 5, 6], [], [7, 8, 9], None, [10]]
```

which overshoots the number of values on the primitive array (`[0, 1, 2, None, 3, 4, 5,6,7,8,910].len() = 12`). The equality in the `debug_assert` is only valid for non-nested types.

I suggest we move the `debug_assert` to inside the non-nested branch and document why this is true in that case (imo it should raise an error, since data loss should be an error, but we can leave it for another PR - there is currently a mix of error and panic in this code base).

danburkert:
Got it, thanks for the explanation.  I've gone ahead and moved the checks to within the non-nested branch.

danburkert:
@jorgecarleitao one thing that occurred to me after having some time to digest your explanation - if the length/capacity as used in those methods isn't accurate for nested columns, then the capacity reservations for the data and validity buffers is going to be too large.  Obviously won't cause bugs or errors, but perhaps something that could be optimized.

danburkert:
And thanks for the quick help on this! 🙏

jorgecarleitao:
I agree. The reading of parquet is about to take an overall (#789), so I think we should revisit this after #789 lands.

PR #798, opened 2022-02-03
`MutableUtf8Array::extend_values`

ritchie46:
`MutableUtf8Array` already has `Extend` for iterators over `Option<S: AsRef<str>` and iterators over values and optional values from trusted length iterators, but not yet a method for iterators over values whose length cannot be trusted.

This PR adds that.

### DIFF
--- a/tests/it/array/utf8/mutable.rs
+++ b/tests/it/array/utf8/mutable.rs
@@ -80,3 +80,18 @@ fn test_extend_trusted_len() {
         Some(&Bitmap::from_u8_slice(&[0b00011011], 5))
     );
 }
+
+#[test]
+fn test_extend_values() {
+    let mut array = MutableUtf8Array::<i32>::new();
+
+    array.extend_values([Some("hi"), None, Some("there"), None].iter().flatten());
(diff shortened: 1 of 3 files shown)


tests/it/array/utf8/mutable.rs
```diff
@@ -80,3 +80,18 @@ fn test_extend_trusted_len() {
         Some(&Bitmap::from_u8_slice(&[0b00011011], 5))
     );
 }
+
+#[test]
+fn test_extend_values() {
+    let mut array = MutableUtf8Array::<i32>::new();
+
+    array.extend_values([Some("hi"), None, Some("there"), None].iter().flatten());
```
jorgecarleitao, line 88: why the flatten instead of `[Some("hi"), Some("there")].iter()`?

ritchie46: To create an iterator whose length cannot be trusted. The `None` values will be removed.

jorgecarleitao: ah, got it! 👍

Issue #804, opened 2022-02-04
Panic vs Error when reading Parquet


danburkert:
Currently much of the parquet reading section of the codebase is set up to panic when invalid files are read, for example:

* https://github.com/jorgecarleitao/arrow2/blob/main/src/array/specification.rs
* https://github.com/jorgecarleitao/arrow2/blob/8fcab5cc5e9fd5f9eab9f2d38c0e1c124da05a36/src/array/utf8/mod.rs#L68-L73
* https://github.com/jorgecarleitao/arrow2/blob/main/src/array/primitive/mod.rs#L62

This makes it more difficult to reading untrusted parquet files, or parquet files written by a system with a buggy parquet implementation (e.g. there are real-world cloud databases that will write out strings with invalid UTF-8 bytes).  It seems that many of the public APIs are already set up to return errors, and there is a Parquet error type corresponding to out-of-spec data.

Is removing panics from the library, and replacing panicking APIs with Result/Error equivalents an option for parquet2 & arrow2?

jorgecarleitao:
Yeap, everything needs to be migrated to errors - all io interfaces must not panic on invalid data.

I think that one starting point is to introduce `try_new` to all arrays, so that we can use them instead of `from_data` (that panics).

We then need to audit the code and replace all panics by errors in IO.

danburkert:
Great - I'm trying to solve one such instance in our system where a trusted, but buggy, warehouse is writing out parquet files with invalid utf-8 bytes.  I'll try and put together a PR fix for that specific scenario using the `try_new` guidance.

Eventually it'd be great to get to the point where the library is fuzzed, but obviously that's a bit off.  Thanks!

PR #807, opened 2022-02-04
Return error when reading Parquet column with invalid utf-8 bytes

danburkert:
This PR switches the Parquet reading internals to return an error
instead of panicking when data with invalid utf-8 bytes is read. Per the
advice in #804 I did this by introducing new `try_*` variants of methods
which panic, and switched the callstack over. Thus, I believe this is an
API compatible change (`TraitBinaryArray` is changed incompatibly, but
it appears to be crate private).

I've attached a very small .parquet file with an invalid utf-8 byte that
I used for testing. Without this change using the
`parquet_read_parallel` example binary with the file panics, whereas now
it errors. I think this file should be added to the test suite, but I
need some guidance as to exactly how to structure that.


jorgecarleitao:
looks great!

What do you think about renaming `try_from_data` to `try_new`? I believe that this is more idiomatic in Rust (imo `from_data` should also be changed to `new` at some point).

jorgecarleitao:
would it be possible to write the parquet file as a `vec![]` in the tests themselves? This allows the test to run against MIRI without disabling isolation (MIRI does not accept opening a file).

I think it would also be worthwhile to have a test against the file so that we show that invalid utf8 error.

danburkert:
> What do you think about renaming try_from_data to try_new? I believe that this is more idiomatic in Rust (imo from_data should also be changed to new at some point).

Sounds good.

> would it be possible to write the parquet file as a Vec<u8> in the tests themselves? This allows the test to run against MIRI without disabling isolation (MIRI does not accept opening a file).

I can definitely do that - is there a test suite file that you have in mind?

jorgecarleitao:
> I can definitely do that - is there a test suite file that you have in mind?

Somewhere here: https://github.com/jorgecarleitao/arrow2/blob/main/tests/it/io/parquet/read.rs ?

danburkert:
Methods are renamed, and test is added.

jorgecarleitao:
Thanks a lot, @danburkert.

PR #812, opened 2022-02-05
fix unescaped '"' in json writing

### LINKED ISSUE #811: Unescaped " in value produces invalid json when encoded
coming here from polars: 
reference issue: https://github.com/pola-rs/polars/issues/2548

When encoding any value containing `"` to json will produce invalid json. The `"` is not escaped correctly. An example is available here https://github.com/pola-rs/polars/issues/2548#issuecomment-1030593703 (It comes from polars which I hope is ok, but if needed I can try making a new one for only the arrow2 part)

Some digging lead me here (`"` is an ascii byte)
https://github.com/jorgecarleitao/arrow2/blob/9a8edd06a59d01023ec5a9f7ad1775f7e6c1969c/src/io/json/write/serialize.rs#L141
```rust
#[inline]
fn utf8_serialize(value: &str, buf: &mut Vec<u8>) {
    if value.as_bytes().is_ascii() {
        buf.reserve(value.len() + 2);
        buf.push(b'"');
        buf.extend_from_slice(value.as_bytes());
        buf.push(b'"');
    } else {
        // it may contain reserved keywords: perform roundtrip for
        // todo: avoid this roundtrip over serde_json
        serde_json::to_writer(buf, &Value::String(value.to_string())).unwrap();
    }
}
```

ritchie46:
fixes #811.

This is a quick hack that also searches for `"` before branching.

However, the second branch first did an unnecessary heap convertion from `&str` -> `String`. I removed it and the tests still succeed. 

I think that we maybe should not branch at all now we don' do this heap allocation? Because we are now doing a double scan over the data determine which path we must take.

### DIFF
--- a/src/io/json/write/serialize.rs
+++ b/src/io/json/write/serialize.rs
@@ -1,5 +1,4 @@
 use lexical_core::ToLexical;
-use serde_json::Value;
 use streaming_iterator::StreamingIterator;
 
 use crate::bitmap::utils::zip_validity;
@@ -139,7 +138,7 @@ fn list_serializer<'a, O: Offset>(
 
 #[inline]
 fn utf8_serialize(value: &str, buf: &mut Vec<u8>) {
-    if value.as_bytes().is_ascii() {
+    if value.as_bytes().is_ascii() && !value.contains('"') {
         buf.reserve(value.len() + 2);
         buf.push(b'"');
         buf.extend_from_slice(value.as_bytes());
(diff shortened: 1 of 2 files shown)


jorgecarleitao:
Thanks! I am working on a alloc free version, one sec :P

jorgecarleitao:
Alternative: https://github.com/jorgecarleitao/arrow2/pull/813

jorgecarleitao:
Actually, I think your proposal is better: do not branch and always use serde_json 👍

ritchie46:
Haha.. just let me know which PR that will be. :see_no_evil:

jorgecarleitao:
This one is way better. Thanks for the quick fix!

PR #814, opened 2022-02-05
add json serialization of timestamp/date32/date64

ritchie46:
This adds serialization of dates and timestamp. I started with the default `Display` implementation from chrono. We could add formatter later if that's required.


jorgecarleitao:
small conflict

jorgecarleitao:
Thanks, @ritchie46 !

Issue #733, opened 2022-01-05
Add CI job to roundtrip avro over pyspark


jorgecarleitao:
So that we can increase our confidence that our implementation is correct.

potter420:
Seem like we're missing some Decimal Type. Can I add it ?

jorgecarleitao:
Thank you! Note that we have something similar in reading  the parquet implementation that we may use - I _think_ it uses an equivalent encoding.

potter420:
I will reuse that as soon as I figure out nice way to extract correct non `Null` schema from `AvroSchema`. I plan to get the &[u8] from the value and proceed forward. Need a test case where the avro physical types is a `fixed` rather than `bytes`

PR #837, opened 2022-02-15
Add `Decimal128` to avro round trip (Read/Write)

potter420:
Hi, this is to partially address #733 .

Added read of `Decimal128` from both `fixed` and `bytes` schema.

Support writing to `bytes` schema to avro.

### DIFF
--- a/src/io/avro/read/deserialize.rs
+++ b/src/io/avro/read/deserialize.rs
@@ -257,13 +257,12 @@ fn deserialize_value<'a>(
                     block = &block[len..];
                     let mut bytes = [0u8; 16];
                     bytes[..len].copy_from_slice(value);
-                    let data = i128::from_be_bytes(bytes) >> (8 * (16 - len));
-                    // println!("{:?}", data);
+                    let data = u128::from_be_bytes(bytes) >> (8 * (16 - len));
                     let array = array
                         .as_mut_any()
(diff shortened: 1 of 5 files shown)


jorgecarleitao (commented):
Thanks a lot, @potter420 . Would it be possible to add 2 tests, one in the `read` where we show that we can read from avro-rs, and one in `write`, to show that we can roundtrip decimals?

potter420:
@jorgecarleitao For some reason I can't pass the `ipc` and `parquet` test locally, despite running `write_parquet.py`.
The `avro` test case show some errors for certain special number:
```rust
    let value = [188u8, 97u8, 78u8]; //12345678
    let mut bytes = [0u8; 16];
    bytes[..3].copy_from_slice(&value);
    println!("{:?}", bytes);
    let d = (u128::from_be_bytes(bytes) >> (8*(16-3))) as i128;
```

Reason: 188 is 10111100, resulting `i128::from_be_bytes(bytes)` turn the original bytes into a negative number (since 1 is there). When doing the bit shift, the resulting number will be filled with 1. 
`[255, 255, 255, .... , 188u8, 97u8, 78u8]`

Using `u128` however does not result in this behavior, as we 0 filling correctly.

So the test failed. Should I change that in the parquet implementation as well?

jorgecarleitao:
Flawless. Thanks a lot, @potter420 !

PR #851, opened 2022-02-17
Add lifetime parameterization to async IPC reader

### LINKED ISSUE #850: Add lifetime parameterization to async IPC reader
Currently the async IPC reader (`io::ipc::read::stream_async::AsyncStreamReader`) has a `'static` lifetime. It would be more ergonomic to parameterize its lifetime:

```rust
let mut buffer = futures::io::Cursor::new(Vec::new());
let reader = AsyncStreamReader::new(&mut buffer);
...
drop(reader);
let _out = reader.into_inner(); // This fails to compile because of the 'static mutable borrow
```

dexterduck:
Closes #850.

Pretty small ergonomic change.


jorgecarleitao:
Ahhhh, you won the war I started (and I lost a battle with) the compiler. Thank you! 🙇

PR #854, opened 2022-02-18
Simplified API for FFI

jorgecarleitao:
Inspired by [this observation](https://github.com/jorgecarleitao/arrow2/issues/673#issuecomment-1044788231) from @[multimeric](https://github.com/multimeric), I just realized that we do not really need a `Field` to read an array - the `DataType` is sufficient.

This PR leverages this observation to simplify the API a bit further. The updated examples show the idea.

### DIFF
--- a/arrow-pyarrow-integration-testing/src/lib.rs
+++ b/arrow-pyarrow-integration-testing/src/lib.rs
@@ -66,7 +66,8 @@ fn to_rust_array(ob: PyObject, py: Python) -> PyResult<Arc<dyn Array>> {
     )?;
 
     let field = unsafe { ffi::import_field_from_c(schema.as_ref()).map_err(PyO3ArrowError::from)? };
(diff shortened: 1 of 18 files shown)


jorgecarleitao:
> Is the idea with this PR that the Schema isn't needed, and so a Chunk is therefore equivalent to a RecordBatch?

Yes, that is the core observation - exporting an Array does not require a schema nor fields, only its datatype. Because every array stores such a datatype in this implementation of Arrow, we can just expose the API for that.

This does not close #854 , they are related but independent :)

jorgecarleitao:
> Also I don't understand how you can pass the data type into the C function that expects a schema pointer. How does that work at all? Are we not losing information stored in the Schema as well?

What is "the C function"? What is "Schema"?

There are two things supported by the C data interface:
1. A `Field` (`Ffi_ArrowSchema`)
2. An `Array` (`Ffi_ArrowArray`)

PR #855, opened 2022-02-18
Fixed error in writing compressed IPC arrow

jorgecarleitao:
There is some ambiguity in the format over how to count buffer lengths, but I think this handles it.

Will work on `apache/arrow` to clear up the ambiguity and add an integration test for this.


jorgecarleitao:
cc @ghuls (related to https://github.com/pola-rs/polars/issues/2639)

ghuls:
Thanks, I will test it on a big file to see if PyArrow likes it now.

Issue #865, opened 2022-02-25
CSV writing: distinguish between missing and empty strings


ritchie46:
Currently the csv writer encodes missing data and empty strings equally. Because a csv field may be quoted, we could use that as the empty string value.

2 empty strings
```
"",""\n
```

2 missing data
```
,\n
```

## Current behavior:

If we encode a single column, we seem to quote both missing and emtpy:

```python
f = io.BytesIO()
pl.DataFrame({
    "s1": ["", None],
}).to_csv(f)

f.seek(0)
print(f.read())
```
```
b's1\n""\n""\n'
```

However if we encode multiple columns, we change our behavior and we don't encode any.

```python
f = io.BytesIO()
pl.DataFrame({
    "s1": ["", None],
    "s2": ["", None],
}).to_csv(f)

f.seek(0)
print(f.read())
```
```
b's1,s2\n,\n,\n'
```

I think we should also be consistent no matter how many columns we encode.

ritchie46:
This turns out to be harder than I thought in current setup. We dispatch the `&[u8]` to the csv crate, which handles escaping. It has no notion of a missing string vs and empty string.

One thing that could help this is doing the csv encoding only in the string serializers. This gives us more control and will likely also be faster as we don't encode any other data types. At the moment the whole row is encoded, which in case of no string data is a wasted scan.

When writing the whole row, we can just slam a delimiter between the serialized fields and end with a new line.

PR #866, opened 2022-02-25
change csv-writer

ritchie46:
This addresses #865. It is a breaking change, but it gives us more control and very likely improves performance because we only use `csv-core` for `utf8` fields. 

Because we have a lot less variants than the original csv crate has to deal with. E.g. 

* we know the schema, 
* we know that we will have an equal number of rows and fields

### DIFF
--- a/src/io/csv/write/mod.rs
+++ b/src/io/csv/write/mod.rs
@@ -48,11 +48,13 @@ pub fn serialize<A: AsRef<dyn Array>>(
                 row.extend_from_slice(field);
                 row.push(options.delimiter);
             });
-        // replace last delimiter with new line
-        let last_byte = row.len() - 1;
-        row[last_byte] = b'\n';
-        rows.push(row.clone());
-        row.clear();
+        if !row.is_empty() {
+            // replace last delimiter with new line
+            let last_byte = row.len() - 1;
+            row[last_byte] = b'\n';
+            rows.push(row.clone());
+            row.clear();
+        }
         Result::Ok(())
     })?;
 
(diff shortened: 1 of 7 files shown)


ritchie46:
Still have to do the parallel example.

src/io/csv/write/serialize.rs
```diff
@@ -187,6 +206,45 @@ fn timestamp_with_tz<'a>(
...
+        array.iter(),
+        move |x, buf| {
+            match x {
+                // Empty strings are quoted.
+                // This will ensure a csv parser will not read them as missing
+                // in a delimited field
+                Some("") => buf.extend_from_slice(b"\"\""),
+                Some(s) => {
+                    let bytes = s.as_bytes();
+                    buf.reserve(bytes.len() * 2);
```
jorgecarleitao, line 226: shouldn't this be `+ 2`?

jorgecarleitao (commented):
Great PR!

Indeed this will be more performant. I left one comment.

src/io/csv/write/serialize.rs
ritchie46: We don't know how much quoting will be done. But now that I think of it reserving with `buf.len()` would be best I think.

That would be exponential scaling. What do you think?

src/io/csv/write/mod.rs
```diff
@@ -26,55 +26,85 @@ fn new_serializers<'a, A: AsRef<dyn Array>>(
...
-    Ok(records)
+            // `unwrap` is infalible because `array.len()` equals `Chunk::len`
+            .for_each(|iter| {
+                let field = iter.next().unwrap();
+                row.extend_from_slice(field);
+                row.push(options.delimiter);
+            });
+        // replace last delimiter with new line
+        let last_byte = row.len() - 1;
+        row[last_byte] = b'\n';
```
ritchie46, line 53: I shall update this part to deal with the empty row case so that we don't underflow.

jorgecarleitao:
Indeed. Thanks a lot, @ritchie46 !

PR #871, opened 2022-02-28
Improved performance of filter performance via Simd selection [3x]

sundy-li:
Benchmarks:

Before:

```rust
filter f32              time:   [269.62 us 269.88 us 270.12 us]                       
filter f32 high selectivity                                                                            
                        time:   [243.83 us 244.41 us 244.95 us]
filter f32 nonull       time:   [108.89 us 108.95 us 109.03 us]                              
filter f32 nonull high selectivity                                                                             
                        time:   [18.886 us 18.909 us 18.933 us]
 
```

Now:

``` 
filter f32            time:   [180.22 us 180.34 us 180.45 us]                       
filter f32 high selectivity                                                                            
                        time:   [94.458 us 94.495 us 94.540 us]
			
filter f32 nonull       time:   [58.747 us 58.775 us 58.809 us]                              
filter f32 nonull high selectivity                                                                             
                        time:   [10.456 us 10.477 us 10.494 us]
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -249,13 +249,15 @@ harness = false
 name = "comparison_kernels"
 harness = false
 
-[[bench]]
-name = "read_parquet"
-harness = false
+## disabled because we don't have the path of env `CARGO_MANIFEST_DIR`
 
-[[bench]]
-name = "write_parquet"
-harness = false
+# [[bench]]
+# name = "read_parquet"
+# harness = false
+
+# [[bench]]
+# name = "write_parquet"
+# harness = false
 
 [[bench]]
 name = "aggregate"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -249,15 +249,14 @@ harness = false
 name = "comparison_kernels"
 harness = false
 
-## disabled because we don't have the path of env `CARGO_MANIFEST_DIR`
 
-# [[bench]]
-# name = "read_parquet"
-# harness = false
+[[bench]]
+name = "read_parquet"
+harness = false
 
-# [[bench]]
-# name = "write_parquet"
-# harness = false
+[[bench]]
+name = "write_parquet"
+harness = false
 
 [[bench]]
 name = "aggregate"
--- a/src/types/bit_chunk.rs
+++ b/src/types/bit_chunk.rs
@@ -1,7 +1,9 @@
 use std::{
     fmt::Binary,
-    ops::{BitAnd, BitAndAssign, BitOr, Not, Shl, ShlAssign, ShrAssign},
+    ops::{BitAndAssign, Not, Shl, ShlAssign, ShrAssign},
 };
+
+use num_traits::PrimInt;
 
 use super::NativeType;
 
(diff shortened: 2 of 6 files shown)


src/types/bit_chunk.rs
```diff
@@ -113,6 +99,54 @@ impl<T: BitChunk> Iterator for BitChunkIter<T> {
...
+
+        self.remaining -= 1;
+        Some(v)
+    }
+
+    #[inline]
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        (self.remaining, Some(self.remaining))
+    }
+}
```
jorgecarleitao, line 148:
```suggestion
}

// # Safety
// a mathematical invariant of this iterator
impl<T: BitChunk> crate::trusted_len::TrustedLen for BitChunkOnes<T>
```

so that we can confidently use it without checking for lengths.

Cargo.toml
```diff
@@ -249,13 +249,15 @@ harness = false
 name = "comparison_kernels"
 harness = false
 
-[[bench]]
```
jorgecarleitao, line 252: could you re-activate these, or, what is the issue?

src/types/bit_chunk.rs
```diff
@@ -113,6 +99,54 @@ impl<T: BitChunk> Iterator for BitChunkIter<T> {
...
+/// Refer: https://lemire.me/blog/2018/03/08/iterating-over-set-bits-quickly-simd-edition/
+/// # Example
+/// ```
+/// use arrow2::types::BitChunkOnes;
+/// let a = 0b00010000u8;
+/// let iter = BitChunkOnes::new(a);
+/// let r = iter.collect::<Vec<_>>();
+/// assert_eq!(r, vec![4]);
+/// ```
+pub struct BitChunkOnes<T: BitChunk> {
```
jorgecarleitao, line 113: beautiful abstraction - small and well encapsulated; really cool!

src/types/bit_chunk.rs
```diff
@@ -125,4 +159,13 @@ mod tests {
         let r = iter.collect::<Vec<_>>();
         assert_eq!(r, (0..16).map(|x| x == 0 || x == 12).collect::<Vec<_>>(),);
     }
+
+    #[test]
+    fn test_ones() {
+        let a = [0b00000001, 0b00010000]; // 0th and 13th entry
+        let a = u16::from_ne_bytes(a);
+        let mut iter = BitChunkOnes::new(a);
```
jorgecarleitao, line 167:
```suggestion
        let mut iter = BitChunkOnes::new(a);
        assert_eq!(iter.size_hint(), (2, Some(2));
```

jorgecarleitao (commented):
Amazing PR, @sundy-li ! Left a small number of comments but it is otherwise ready imo.

Cargo.toml
sundy-li: I don't find any doc to set env `CARGO_MANIFEST_DIR` in this repo.

If we activate this, it will make other benches fail.

jorgecarleitao: isn't `CARGO_MANIFEST_DIR` set automatically by cargo on build?

There shouldn't be interaction between benches (could you give the error?). The `read_parquet` does need some parquet files to be generated (via `python parquet_integration/write_parquet.py`) at runtime, but it should still compile.

Note that it is possible to run only specific benches. E.g. in this case I would use

```
cargo bench --features compute_filter --bench filter_kernels
```

which should be quite fast to compile.

sundy-li: `CARGO_MANIFEST_DIR=/tmp cargo bench`  works now.

Cargo.toml
```diff
@@ -249,6 +249,7 @@ harness = false
 name = "comparison_kernels"
 harness = false
 
+
```
rapiz1, line 252: A blank line slipped in

PR #872, opened 2022-03-01
remove superfluous assert

ritchie46:
I was looking at the concat kernel and saw we loop over the `&[& dyn Array]` array to check if the `DataTypes` are equal, then we dispatch to `make_growable` to do the same loop, but then we `assert` if they are not equal. Then we dispatch to the implementations where we `downcast` via `Any`, so yet another assert on the `DataType`.

This PR removes one of the loops where we `assert`.

For a future PR I am thinking about splitting it in a fallible and one that only asserts at the downcast. I have quite a few cases where I create a very large number of small arrays of the same `DataType` in polars. Those loops then need to loop a large number of elements.

### DIFF
--- a/src/array/growable/mod.rs
+++ b/src/array/growable/mod.rs
@@ -73,7 +73,6 @@ pub fn make_growable<'a>(
 ) -> Box<dyn Growable<'a> + 'a> {
     assert!(!arrays.is_empty());
     let data_type = arrays[0].data_type();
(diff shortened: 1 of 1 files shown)


sundy-li:
LGTM， how about making it `debug_assert` ?

ritchie46:
> LGTM， how about making it `debug_assert` ?

Yes, normally I would agree with that. But because we already `unwrap` in `dowcast` and thus panic, even that feels a bit redundant to me.

PR #874, opened 2022-03-02
expose `ListValuesIter`

ritchie46:
I want to implement a trait for an arrow defined iterator. This makes that possible.

### DIFF
--- a/src/array/list/iterator.rs
+++ b/src/array/list/iterator.rs
@@ -13,7 +13,7 @@ pub struct ListValuesIter<'a, A: IterableListArray> {
 
 impl<'a, A: IterableListArray> ListValuesIter<'a, A> {
     #[inline]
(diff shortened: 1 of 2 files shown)


src/array/list/iterator.rs
```diff
@@ -13,7 +13,7 @@ pub struct ListValuesIter<'a, A: IterableListArray> {
 
 impl<'a, A: IterableListArray> ListValuesIter<'a, A> {
     #[inline]
-    pub fn new(array: &'a A) -> Self {
+    pub(crate) fn new(array: &'a A) -> Self {
```
jorgecarleitao, line 16: doesn't this need to be public for you to use it?

ritchie46: We get it by calling `values_iter` on a list array. Don't think we need a `new` publicly? Don't have much opinion about it though.

PR #876, opened 2022-03-02
Async sink interfaces for IPC and Parquet IO

dexterduck:
Following up on discussion here: https://github.com/jorgecarleitao/parquet2/pull/78.

This PR implements new types that expose the `futures::Sink` trait for Arrow IPC and Parquet writers. It also implements an async stream type for reading the IPC file format (since currently async reading is only available for the IPC stream format). Specific new types:

- `io::ipc::read::file_async::FileStream` - implements `futures::Stream` for IPC files.
- `io::ipc::write::file_async::FileSink` - implements `futures::Sink` for IPC files.
- `io::ipc::write::file_async::StreamSink` - implements `futures::Sink` for IPC streams.
- `io::parquet::write::FileSink` - implements `futures::Sink` for Parquet files.

Definitely interested in feedback on these changes! 
In particular I felt pretty unsure about file structure and naming for the new types so happy to change those in any way that makes sense. I also added tests directly into some of the modules, but I wasn't sure if there was a different place I should be putting those.


jorgecarleitao:
Holy molly amazing PR ❤️

IMO this is basically ready to merge; the API looks great, it is ready easy to follow it, and overall really great work.

Some comments:

* In the parquet, we have a `FileStreamer`. Could you comment what is the difference / whether we can replace that by this? If they are solving the same problem, imo we should just remove the other one. ^^
* Could you move all the tests to `tests/it/io/*`? This crate has most of its tests outside the main crate, following https://matklad.github.io/2021/02/27/delete-cargo-integration-tests.html
* Could you split this PR in two: parquet and IPC; imo they deserve two entries in the change log and corresponding PRs.

dexterduck:
Sure, I'll split it up, move the tests, and re-submit!

`FileSink` essentially does the same thing as `FileStreamer`, so it could replace `FileStreamer` if we're okay with the breaking change to the interface.

jorgecarleitao:
Thanks a lot, @dexterduck 🙇 

yes, for now it is ok to break APIs in `io/` - you found/proposed a better one; let's use it :)

PR #878, opened 2022-03-02
IPC sink types and IPC file stream

dexterduck:
IPC component of this PR: #876 

Adds the following new types:
- `io::ipc::read::file_async::FileStream` - implements `futures::Stream` for IPC files.
- `io::ipc::write::file_async::FileSink` - implements `futures::Sink` for IPC files.
- `io::ipc::write::file_async::StreamSink` - implements `futures::Sink` for IPC streams.


dexterduck:
@jorgecarleitao along the same lines as the Parquet integration, the `Sink` implementations could replace the file/stream writers for IPC as well, although it would require a wrapper type in order to preserve the ability to do per-chunk projections. 

Potentially something to the effect of:

```rust
// Wrapper type
pub struct Projection {
  pub chunk: Chunk<Arc<dyn Array>>,
  pub fields: Option<Vec<IpcField>>,
}

impl From<Chunk<Arc<dyn Array>>> for Projection {
  fn from(chunk: Chunk<Arc<dyn Array>>) -> Self {
    Self { chunk, fields: None }
  }
}

// Usage
let mut sink = FileSink::new(..);
let chunks: Vec<Chunk<Arc<dyn Array>>> = vec![];
for chunk in chunks {
  sink.feed(chunk.into()).await?;
}
```

jorgecarleitao:
* The `IPCField` is required to re-use dictionaries across batches. So, we do need it in the API.
* I am not very confortable with code duplication - could you either change the existing code (e.g. implement `Sink` for the structs we have) or remove them? Having 2 implementations of the same thing is not ideal imo

src/io/ipc/write/common.rs
```diff
@@ -379,3 +380,55 @@ pub struct EncodedData {
 pub(crate) fn pad_to_8(len: usize) -> usize {
     (((len + 7) & !7) - len) as usize
 }
+
+/// An array [`Chunk`] with optional accompanying IPC fields.
+#[derive(Debug, Clone, PartialEq)]
+pub struct Record<'a> {
+    columns: Cow<'a, Chunk<Arc<dyn Array>>>,
```
jorgecarleitao, line 387: Curious: what is the purpose of `Cow`?

jorgecarleitao (approved):
Looks great! Thank you so much, @dexterduck - learned a lot in this PR!

src/io/ipc/write/common.rs
dexterduck: The idea is to allow passing the arguments either by value or by reference. Obviously, better to avoid unnecessary clones if possible, but also wanted to avoid requiring a reference since then if you want to pass owned values you have to use the somewhat non-ergonomic `sink.feed((&chunk).into())` or `sink.feed((&chunk, &fields[..]).into())`. In contrast, with the chosen implementation you can use `sink.feed(chunk.into())` or `sink.feed((chunk, fields).into())` in all cases. 

I originally used a generic implementation that parameterized `Record` to accept a type implementing `Borrow<_>`, but it turns out that having a single type that implements sink for multiple other types (e.g. `FileSink` implementing both `Sink<Record<Chunk<_>>>` and `Sink<Record&<Chunk<_>>>`) doesn't work because the compiler can't infer the correct generic type when you call `sink.flush()` or `sink.close()`.

jorgecarleitao: Thank you for the explanation. Cool tip!

PR #891, opened 2022-03-07
Bug fix in offset for sliced unions

ncpenke:


src/array/union/mod.rs
```diff
@@ -191,7 +191,10 @@ impl UnionArray {
             fields: self.fields.clone(),
             fields_hash: self.fields_hash.clone(),
             types: self.types.clone().slice(offset, length),
-            offsets: self.offsets.clone(),
+            offsets: self
+                .offsets
+                .clone()
+                .map(|offsets| offsets.slice(offset, length)),
```
ncpenke, line 197: @jorgecarleitao For dense unions wasn't sure if slicing the offsets or modifying using (index + offset) in field_slot was more correct.

jorgecarleitao:
Thanks a lot @ncpenke , this is indeed a bug.

PR #893, opened 2022-03-07
Bug fix in like for scalar compute

zhyass:


houqp:
good catch, mind adding a test case to guard against future regressions?

zhyass:
> good catch, mind adding a test case to guard against future regressions?

Added

jorgecarleitao:
Thanks a lot, @zhyass 🙇

PR #912, opened 2022-03-14
Fixup for new and renamed core::simd fn

workingjubilee:
Pardon our dust!

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -76,7 +76,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: nightly-2022-03-03
+          toolchain: nightly-2022-03-16
           override: true
       - uses: Swatinem/rust-cache@v1
         with:
@@ -99,7 +99,7 @@ jobs:
           submodules: true # needed to test IPC, which are located in a submodule
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: nightly-2022-03-03
+          toolchain: nightly-2022-03-16
           override: true
       - uses: Swatinem/rust-cache@v1
         with:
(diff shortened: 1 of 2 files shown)


workingjubilee:
Oh, the nightly version is pinned, I see.

workingjubilee:
Ah, so the pin will have to be to the next nightly which hasn't been cut yet, because Miri failed a test and the rebuild will happen then.

jorgecarleitao:
thanks a lot @workingjubilee for this PR! Indeed, we pin nightly due to MIRI, since not all nightlies support it.

We could use two nightlies, one for MIRI and one for the `simd`, so that we prove that we can compile under the most recent `std::simd`.

workingjubilee:
Reasonable but simpler to just move forward an additional nightly here. Miri actually supports an increasing number of portable SIMD stuff, so eventually it will be an option to run Miri over the entire codebase, SIMD optimizations included. Also I seem to have mixed up how nightlies are versioned/dated, so it has to tick forward again.

ritchie46:
> Reasonable but simpler to just move forward an additional nightly here. Miri actually supports an increasing number of portable SIMD stuff, so eventually it will be an option to run Miri over the entire codebase, SIMD optimizations included. Also I seem to have mixed up how nightlies are versioned/dated, so it has to tick forward again.

I am for this. I think it is better to always support latest nightly. Currently it is a very specific version which we have to communicate. Whilst saying this I assume the api will stabilize of course. :)

workingjubilee:
Yeah, we're just in an awkward position where we're getting major actual testing/usage and thus picking up things we missed along the way (which lead to this change) and also staring down a difficult design "fork".

workingjubilee:
If I have my numbers right, this should now use a version of the Rust toolchain that runs all the tests, both SIMD and Miri.

jorgecarleitao:
Thank you again @workingjubilee for this PR!

Out of curiosity, how did you know that this would affect arrow2, or is there a mechanism to see which crates are using `std::simd`? Curious because I would be interested in knowing this mechanism ^^

workingjubilee:
Magic powers! (I guessed)

PR #915, opened 2022-03-14
Make sure that NullArray return a single buffer pointer to null.

anliakho2:
Null column must produce n_buffers=1  with a single nullpointer in buffers array. Otherwise `import_array_from_c` would fail with validation error.

### DIFF
--- a/src/array/null.rs
+++ b/src/array/null.rs
@@ -109,7 +109,7 @@ impl std::fmt::Debug for NullArray {
 
 unsafe impl ToFfi for NullArray {
     fn buffers(&self) -> Vec<Option<std::ptr::NonNull<u8>>> {
-        vec![]
+        vec![None]
     }
 
     fn offset(&self) -> Option<usize> {
--- a/src/array/null.rs
+++ b/src/array/null.rs
@@ -109,6 +109,8 @@ impl std::fmt::Debug for NullArray {
 
 unsafe impl ToFfi for NullArray {
     fn buffers(&self) -> Vec<Option<std::ptr::NonNull<u8>>> {
+        // `None` is technically not required by the specification, but older C++ implementations require it, so leaving
+        // it here for backward compatibility
         vec![None]
     }
 


jorgecarleitao:
Thanks.

I wonder who requires a buffer in the null array. E.g. we test this against the C++ implementation and it does not seem to require it: https://github.com/jorgecarleitao/arrow2/blob/main/arrow-pyarrow-integration-testing/tests/test_sql.py#L42

anliakho2:
Also I want to clarify that I'm using a higher level pyarrow api `pyarrow.RecordBatch._import_from_c`

jorgecarleitao:
interesting, because we test exactly this against pyarrow (7.0) [here](https://github.com/jorgecarleitao/arrow2/blob/main/arrow-pyarrow-integration-testing/tests/test_sql.py#L42), but we do not get any error.

Note that I do not disagree with the implementation, I am just trying to find a way to test this

anliakho2:
The test is doing `array._import_from_c` and I'm working on higher level and doing 'RecordBatch._import_from_c'

Ahh found it, it's arrow5, it had hard check for at least one buffer https://github.com/apache/arrow/blob/4591d76fce2846a29dac33bf01e9ba0337b118e9/cpp/src/arrow/c/bridge.cc#L1302
It got removed since arrow6, which now supports single buffer but would also work with 0 buffers

src/array/null.rs
```diff
@@ -109,7 +109,7 @@ impl std::fmt::Debug for NullArray {
 
 unsafe impl ToFfi for NullArray {
     fn buffers(&self) -> Vec<Option<std::ptr::NonNull<u8>>> {
-        vec![]
+        vec![None]
```
jorgecarleitao, line 112:
```suggestion
        // `None` is technically not required by the specification, but older C++ implementations require it, so leaving
        // it here for backward compatibility
        vec![None]
```

jorgecarleitao:
Thanks a lot for sticking and for looking for the root cause, @anliakho2 .

anliakho2:
Pleasure is all mine @jorgecarleitao! And really appreciate quick turnaround and pointers to the test code covering this :)

PR #923, opened 2022-03-20
Migrate to latest parquet2

jorgecarleitao:
This PR adds support for bloom filters (read) and column/offset indexes (read and write).

Work needed:

- [x] map parquet types to arrow types to be able to map min and max in the column indexes to arrow, just like we do for statistics.
- [x] generalize deserialization to support offsetted parquet pages [difficult]

The writing of offsets works out of the box and is active whenever the option `write_statistics` is active, since page offsets are just page-level statistics written outside of a page header.

See https://github.com/jorgecarleitao/parquet2/issues/102 and https://github.com/jorgecarleitao/parquet2/pull/107 for more details

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,7 +53,7 @@ hex = { version = "^0.4", optional = true }
 
 # for IPC compression
 lz4 = { version = "1.23.1", optional = true }
-zstd = { version = "0.10", optional = true }
+zstd = { version = "0.11", optional = true }
 
 rand = { version = "0.8", optional = true }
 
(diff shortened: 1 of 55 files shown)


jorgecarleitao:
Read and write indexes is now complete.

The next and last step is to make deserialization work with offsetted pages. The last commit adds support for this for required primitives and binary pages as well as a test against this.

The tedious work of supporting offsetted pages for the remaining page representations begins.. 🚂🚂🚂🚂

jorgecarleitao:
- [x] flat
- [ ] flat dictionary-encoded
- [ ] flat dictionary-encoded to dictionary
- [ ] nested

jorgecarleitao:
After 2 or 3 re-writes on both parquet2 and arrow2, I think that the design is finally here.

There is no performance degradation in using indexes - indexes only improve performance.

Indexed reads are not yet supported for nested types (I think that this is also true in parquet-mr) - this is feasible, but requires a bit more work since the indexed reading requires skipping values, and this must be correctly done in nested types. parquet2 still does not have auxiliary iterators for this.

jorgecarleitao:
This is still not ready for end-users (we need to pipe the API to the frontend), but it does no harm and is heavily testing. Thus, merging as is.

PR #930, opened 2022-04-03
Add support for union scalars

ncpenke:


src/array/union/mod.rs
```diff
@@ -296,6 +293,13 @@ impl Array for UnionArray {
         None
     }
 
+    fn is_valid(&self, index: usize) -> bool {
```
jorgecarleitao, line 296: is this necessary? A union does not have a "validity". I could not find where this was being used.

src/array/union/mod.rs
```diff
@@ -272,10 +272,7 @@ impl UnionArray {
 
     /// Returns the slot `index` as a [`Scalar`].
     pub fn value(&self, index: usize) -> Box<dyn Scalar> {
-        let type_ = self.types()[index];
-        let field = self.field(type_);
-        let index = self.field_slot(index);
-        new_scalar(field.as_ref(), index)
+        new_scalar(self, index)
```
jorgecarleitao, line 275: this will return a scalar with a `DataType` of `self`. I think we want the scalar to have the datatype of the field. As such, imo this should still do what it does, and `new_scalar` have a different logic (see comment below)

src/scalar/mod.rs
```diff
@@ -144,7 +146,22 @@ pub fn new_scalar(array: &dyn Array, index: usize) -> Box<dyn Scalar> {
...
+            Box::new(UnionScalar::new(
+                array.data_type().clone(),
+                type_,
+                if field_value.is_valid() {
+                    Some(field_value.into())
+                } else {
+                    None
+                },
+            ))
+        }
```
jorgecarleitao, line 163:
```suggestion
        Union => {
            let array = array.as_any().downcast_ref::<UnionArray>().unwrap();
            Box::new(UnionScalar::new(
                array.data_type().clone(),
                array.value(index),
            ))
        }
```

This would be my suggestion

src/scalar/union.rs
```diff
@@ -0,0 +1,48 @@
+use std::sync::Arc;
+
+use crate::datatypes::DataType;
+
+use super::Scalar;
+
+/// A single entry of a [`crate::array::UnionArray`].
+#[derive(Debug, Clone, PartialEq)]
+pub struct UnionScalar {
+    value: Option<Arc<dyn Scalar>>,
```
jorgecarleitao, line 10:
```suggestion
    value: Arc<dyn Scalar>,
```

unions do not have a validity. Thus, the value of an index is always "Some".

src/scalar/union.rs
```diff
@@ -0,0 +1,48 @@
...
+    type_: i8,
```
jorgecarleitao, line 11: I do not think this is used, right?

src/scalar/union.rs
```diff
@@ -0,0 +1,48 @@
...
+
+impl Scalar for UnionScalar {
+    #[inline]
+    fn as_any(&self) -> &dyn std::any::Any {
+        self
+    }
+
+    #[inline]
+    fn is_valid(&self) -> bool {
+        self.value.is_some()
```
jorgecarleitao, line 41:
```suggestion
        true
```

union does not have validity, so it is always valid

jorgecarleitao (commented):
Thank you, @ncpenke , looks great!

I left some comments as I think we can simplify the scalar further due to the observation that `Union` does not have a validity - values are always valid.

src/array/union/mod.rs
ncpenke: I was going off the C++ implementation here https://github.com/apache/arrow/blob/a935c81b595d24179e115d64cda944efa93aa0e0/cpp/src/arrow/array/array_base.cc#L113, where the type of the array seems to be inherited by the scalar (rather than the field). 

There's also a test case https://github.com/apache/arrow/blob/d7fbf5260433dee08ae219e8aec34169cd610652/python/pyarrow/tests/test_scalars.py#L663 that seems to be consistent with this.

The reason passing in both the type and the union's field seemed important is to disambiguate cases where the union contains fields that have the same type. For example if we want to represent the following rust enum in Arrow:

```rust
enum E {
    UnitVariant1,
    UnitVariant2,
    IntVariant1(i64),
    IntVariant2(i64),
}
```

To retrieve the original enum from the UnionScalar, we need to be distinguish between the different fields of the same type. Presumably the different fields could have different names, so we could use that but it seemed much cleaner to have access to the typeid and fields.

Does this make sense? In general, unions seem to be an underused area in Arrow, and documentation has been light so I could wrong. Please let me know your thoughts.

src/scalar/union.rs
ncpenke: Please see comment above, the type would be useful to convert back to the original type (I'm using it in arrow2_convert for deserialization back to the original enum, though can work around it by using the field name instead, though that could be less clean).

src/scalar/union.rs
ncpenke: I think I misinterpreted the format spec. I thought unions did have validity based on the example for dense unions where one of the values is null (https://arrow.apache.org/docs/format/Columnar.html#union-layout) just not an explicit validity mask. I had initially used this in new_scalar to follow a similar pattern to the struct, but ended up going with a different implementation to avoid extra lookups.

src/array/union/mod.rs
ncpenke: @jorgecarleitao After looking into this more, I propose we leave the `data_type` and `type_ changes` as-is and `type_` to `type_id`. This would be more expensive memory-wise, but would be consistent with the C++ library implementations and allow for better equality checks (for example the type id of two UnionScalars can be compared instead of the field name of the underlying schema). 

All the other scalar types inherit the list's data_type. If UnionScalar does the same, then that would be consistent and also provide more information for consumers of the scalar.

jorgecarleitao: I understand. What I am trying to say is that given a union with fields `f1` and `f2`, the datatype of the _value_ at position `i` should be of data type f1 or f2 depending on type. The same way that the value of a `ListArray` at position `i` contains an array of its inner field, not a `ListScalar`.

IMO this is independent of how we should represent a UnionScalar (with which I agree with you).

The scalar returned by `new_scalar(&dyn UnionScalar::new(...), index)` is a `UnionScalar` with data type equal to `DataType::Union(...)`, since `new_scalar` converts an index of an array into its scalar representation.

ncpenke:
@jorgecarleitao Latest changes with your feedback - thanks!

src/scalar/union.rs
```diff
@@ -0,0 +1,54 @@
...
+    #[inline]
+    pub fn new(data_type: DataType, type_: i8, value: Arc<dyn Scalar>) -> Self {
+        Self {
+            value,
+            type_,
+            data_type,
+        }
+    }
+
+    /// Returns the values irrespectively of the validity.
```
jorgecarleitao, line 26:
```suggestion
    /// Returns the inner value
```

(there is no validity in the Union (and thus in the `UnionScalar`)

jorgecarleitao (approved):
Looks awesome, super clean and easy to follow. 

Thanks a lot. Left a minor documentation suggestion.

src/scalar/union.rs
ncpenke: Much clearer - committed it

PR #934, opened 2022-04-07
use cached null count in filters

ritchie46:
We can use the cached null counts to determine fast paths in `filter` operations.

Additional note:

I used the following hack to clone `&dyn Array`. I don't like it TBH. If you are ok with it I will add a `boxed(&self) -> Box<dyn Array>` to the `Array` trait.

```rust
    if false_count == 0 {
        assert_eq!(array.len(), filter.len());
        // a hack to clone
        return Ok(array.with_validity(array.validity().cloned()));
    }
```


jorgecarleitao:
Thanks!

There is a function `arrow2::array::clone`, but I like the idea of adding it to the trait, like you suggested

PR #935, opened 2022-04-07
don't allocate string when writing to csv

ritchie46:
Writing to csv is pretty slow. Polars can read two datetime columns in 500 ms, whereas writing the same data to csv takes 33.6 seconds.

Upon looking at the code I saw we allocate a `String` for every value we `format`. 

This PR makes sure we write immediately to the `Vec<u8>` buffer.


jorgecarleitao:
oof, thanks a lot, @ritchie46 , really good stuff!

PR #946, opened 2022-04-13
parquet: split pages by array size

### LINKED ISSUE #932: Parquet: Snappy buffer size limitation.
Writing large parquet files can lead to:

```
ArrowError(ExternalFormat("External format error: underlying snap error: snappy: input buffer (size = 68397383472) is larger than allowed (size = **4294967295)**
```

I believe we already discussed this earlier in a polars issue, but I cannot reproduce it. I open this issue to explore ways to circumvent it. 
I assume we can prevent this by passing smaller `Chunks` to the writer?

Is there a way upfront to estimate the size of a buffer that will be send to snappy?

ritchie46:
This is an attempt to resolve #932 and https://github.com/pola-rs/polars/issues/3120.

As I understand from this post, a row groups column may consist of multiple pages. http://cloudsqale.com/2020/05/29/how-parquet-files-are-written-row-groups-pages-required-memory-and-flush-operations/

So I tried to split an incoming `dyn Array` into `DynIter<Page>` of smaller pages instead of a single large one. The threshold is estimated by the estimated byte size of the array and the limit of `i32::MAX` bytes + some leeway.

Let me know if I am completely wrong here. :see_no_evil:

### DIFF
--- a/src/compute/aggregate/mod.rs
+++ b/src/compute/aggregate/mod.rs
@@ -1,10 +1,15 @@
 //! Contains different aggregation functions
+#[cfg(feature = "compute_aggregate")]
 mod sum;
+#[cfg(feature = "compute_aggregate")]
 pub use sum::*;
 
+#[cfg(feature = "compute_aggregate")]
 mod min_max;
+#[cfg(feature = "compute_aggregate")]
 pub use min_max::*;
 
 mod memory;
 pub use memory::*;
+#[cfg(feature = "compute_aggregate")]
 mod simd;
--- a/src/compute/mod.rs
+++ b/src/compute/mod.rs
@@ -11,7 +11,7 @@
 //! Some dynamically-typed operators have an auxiliary function, `can_*`, that returns
 //! true if the operator can be applied to the particular `DataType`.
 
-#[cfg(feature = "compute_aggregate")]
+#[cfg(any(feature = "compute_aggregate", feature = "io_parquet"))]
 #[cfg_attr(docsrs, doc(cfg(feature = "compute_aggregate")))]
 pub mod aggregate;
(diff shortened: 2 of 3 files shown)


jorgecarleitao:
ahaha, Awesome!

Maybe for now we could just split the array in multiple pages? Most of the issues are coming from a too large page (a parquet page has a max size of `i32::MAX` elements).

This would simplify the dependency tree between compute and parquet?

ritchie46:
> Maybe for now we could just split the array in multiple pages? Most of the issues are coming from a too large page (a parquet page has a max size of i32::MAX elements).

Yeah, sounds good. Do you suggest making a function `array_to_pages` (note the plural)?

jorgecarleitao:
Exactly. Probably an iterator adapter: `ArrayToPages<'a>(array: &'a dyn Array)` that slices the array in chunks of say `i32::MAX` bytes.

Note that technically the size of a page is larger than arrow. For example, a page of N strings of length 4 is `(N + 1) * 4) + M` (+1 because the size of each string is stored as an i32), M is the size of definition levels (`ceil(N/8)` for a non-nested array).

It is not trivial to compute this, so I think that for now we should just give some number of bytes and eventually support it as an option like column-level "Encoding"s.

ritchie46:
Hmm.. It is not super clear to me what we should do differently than the iterator adapter I currently wrote. As we exact byte size is not know until we create it, we need some proxy it seems to me.

Currently we estimate the size and choose a relatively low threshold. Do you want to logic on another location?

jorgecarleitao:
I get it now - I was thinking we had to split the array multiple times, but it is highly unlikely that we need to do it more than once.

Issue #948, opened 2022-04-14
Feature: add `pop`  to `MutableBitmap`


ygf11:
`pop` operation is just opposite to `push`.

Supporting it can make `MutableBitmap` more flexible.

ygf11:
Let me have a try, will create a pr later:)

jorgecarleitao:
Closed by #949

PR #949, opened 2022-04-16
Added pop to mutable bitmap and some mutable array

ygf11:
Part of #948

### DIFF
--- a/src/array/boolean/mutable.rs
+++ b/src/array/boolean/mutable.rs
@@ -104,6 +104,26 @@ impl MutableBooleanArray {
         }
     }
 
+    /// Pop a entry to from [`MutableBooleanArray`].
+    /// Note If the values' length is zero, this method will return None.
+    pub fn pop(&mut self) -> Option<bool> {
+        if self.values.len() == 0 {
+            return None;
+        }
+
+        let value = self.values.pop();
+        match self.validity {
+            Some(ref mut validity) => {
+                if validity.pop() {
+                    Some(value)
+                } else {
+                    None
+                }
+            }
+            None => Some(value),
+        }
+    }
+
     /// Extends the [`MutableBooleanArray`] from an iterator of values of trusted len.
     /// This differs from `extend_trusted_len` which accepts in iterator of optional values.
     #[inline]
--- a/src/array/boolean/mutable.rs
+++ b/src/array/boolean/mutable.rs
@@ -107,14 +107,10 @@ impl MutableBooleanArray {
     /// Pop an entry from [`MutableBooleanArray`].
     /// Note If the values is empty, this method will return None.
     pub fn pop(&mut self) -> Option<bool> {
-        if self.values.is_empty() {
-            return None;
-        }
-
-        let value = self.values.pop();
+        let value = self.values.pop()?;
         match self.validity {
             Some(ref mut validity) => {
-                if validity.pop() {
+                if validity.pop().unwrap() {
                     Some(value)
                 } else {
                     None
--- a/src/array/boolean/mutable.rs
+++ b/src/array/boolean/mutable.rs
@@ -110,7 +110,7 @@ impl MutableBooleanArray {
         let value = self.values.pop()?;
         self.validity
             .as_mut()
-            .map(|x| if x.pop()? { Some(value) } else { None })
+            .map(|x| x.pop()?.then(|| value))
             .unwrap_or_else(|| Some(value))
     }
 
--- a/src/array/boolean/mutable.rs
+++ b/src/array/boolean/mutable.rs
@@ -108,16 +108,10 @@ impl MutableBooleanArray {
     /// Note If the values is empty, this method will return None.
     pub fn pop(&mut self) -> Option<bool> {
         let value = self.values.pop()?;
-        match self.validity {
-            Some(ref mut validity) => {
-                if validity.pop().unwrap() {
-                    Some(value)
-                } else {
-                    None
-                }
-            }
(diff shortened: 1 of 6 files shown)


src/bitmap/mutable.rs
```diff
@@ -94,6 +94,21 @@ impl MutableBitmap {
         self.length += 1;
     }
 
+    /// Pop the last bit from the [`MutableBitmap`].
+    /// # Panics
+    /// Panics iff the length is zero.
```
jorgecarleitao, line 99: should we also make this to return `Option<bool>`, to be consistent with [`Vec::pop`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.pop)?

jorgecarleitao (commented):
This looks really good. Left a comment on the design, but overall pretty much ready.

Thanks a lot, @ygf11 , excellent first contribution!

src/array/primitive/mutable.rs
```diff
@@ -137,6 +137,25 @@ impl<T: NativeType> MutablePrimitiveArray<T> {
         }
     }
 
+    /// Pop a value from the array.
+    /// Note if the values is empty, this method will return None.
+    pub fn pop(&mut self) -> Option<T> {
+        if self.values.is_empty() {
```
Dandandan, line 143: We could match on `self.values.pop()` instead, which does itself perform the same check.

```rust
let value = self.values.pop()?;
...
```

src/bitmap/mutable.rs
ygf11: It is both ok to me.

Since `MutableBitmap` is semantically equivalent to [`Vec<bool>`] as the comment said, let's use `Option<bool>`.
https://github.com/jorgecarleitao/arrow2/blob/d29e135c00cb862f939d67e965b6bb167e550951/src/bitmap/mutable.rs#L10-L11

src/array/primitive/mutable.rs
ygf11: When we get `None` from `self.values.pop()`, there are two conditions:
1. the `array` is empty.
2. the last value of the `array` is None.

So when we match on `self.values.pop()?` first, it will skip `validity.pop()` in the `2nd` condition.

Dandandan: I'm not saying we should skip matching on the validity, but it seems to me we can replace L143-146 by `let value = self values.pop()?`.

src/array/boolean/mutable.rs
```diff
@@ -104,6 +104,26 @@ impl MutableBooleanArray {
         }
     }
 
+    /// Pop an entry from [`MutableBooleanArray`].
+    /// Note If the values is empty, this method will return None.
+    pub fn pop(&mut self) -> Option<bool> {
+        if self.values.is_empty() {
```
Dandandan, line 110: Here the same.

src/array/primitive/mutable.rs
ygf11: If the last value of array is `None`, that means we need to handle `validity bitmap` next, 
but `let value = self.values.pop()?;` will directly return without handle the `validity bitmap` in this condition.

Dandandan: `self.values` is of type `Vec<T>`, where `T` is a primitive type, so by definition it will not return a `None` from this `Vec<T>` itself (it will return `None` on a empty `Vec`).

jorgecarleitao: @ygf11, @Dandandan argument holds because `self.values.len() == self.validity.len()` when the validity exists (it is an invariant of the array)

ygf11: I'm sorry, I recognized `Vec<T>` as `Vec<Option<T>>` here.

ygf11: Addressed.

src/array/boolean/mutable.rs
ygf11: Addressed.

jorgecarleitao (approved):
Looks great! Thanks a lot @ygf11

src/array/boolean/mutable.rs
```diff
@@ -104,6 +104,22 @@ impl MutableBooleanArray {
...
+        match self.validity {
+            Some(ref mut validity) => {
+                if validity.pop().unwrap() {
+                    Some(value)
+                } else {
+                    None
+                }
+            }
+            None => Some(value),
+        }
```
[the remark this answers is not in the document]
ygf11: Thanks, Get it:D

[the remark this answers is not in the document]
Dandandan: Some nice bool function you can use:

`x.pop()?.then_some(value)`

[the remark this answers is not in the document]
ygf11: I never know this function, learn much!
But seems this bool function is unstable ..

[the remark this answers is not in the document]
Dandandan: Ah `then_some` is unstable indeed.

Maybe use then: `x.pop()?.then(||value)`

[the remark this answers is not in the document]
ygf11: Addressed.

jorgecarleitao:
all good from your side, @Dandandan?

Dandandan:
> all good from your side, @Dandandan?

All good! 💯👍

jorgecarleitao:
Thanks again @ygf11 for this PR. Thanks @Dandandan for the review 🙇

PR #972, opened 2022-04-30
Added support to append to existing IPC Arrow file

### LINKED ISSUE #963: Consider adding append options in IPC writer
IPC files are usually pretty huge, so I do not want to read the whole, append new records, and then write the whole buffer, but I wish only to append new records. But unfortunately, now, I cannot find an easy way to append new records in IPC format.

The IPC format has a footer so we can add new records. 
IMO the option to append new records in the IPC writer is an excellent feature. If there is an easy way, adding the example also seems great.

jorgecarleitao:
This PR adds support to append new record batches to an existing Arrow IPC file.

The implementation is simple:

1. read the file's metadata
2. seek to the end of the last block
3. `write` to it

Closes #963 - thanks to @illumination-k for fielding the idea!

### DIFF
--- a/src/io/flight/mod.rs
+++ b/src/io/flight/mod.rs
@@ -25,7 +25,10 @@ pub fn serialize_batch(
     fields: &[IpcField],
     options: &WriteOptions,
 ) -> (Vec<FlightData>, FlightData) {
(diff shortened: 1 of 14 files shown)


ritchie46:
Wow, cool!

PR #974, opened 2022-05-01
Fixed empty reader panic for NDJSON type infer

### LINKED ISSUE #911: json::coerce_data_type fails if there are no lines to be read from reader
In ```ndjson::read::file::infer```, if the reader yields no rows then the method will panic because coerce_data_type expects at least one element in the array.

Solution : return an error, or `Ok(DataType::Null)` if there are no records

Roberto-XY:
Hi there,
I am currently in the process of learning some Rust & what better way then to contribute ;)

- Now returns an error on empty readers instead of panic in `infer`. In `deserialize` as well, to have consistent behavior.
- #911 also suggested handling that as `DataType::Null`. I decided against that, since an empty string is not valid JSON. But `null` is, which also panicked—fixed that. Or do you prefer a separate PR?
- added some explicit tests for null to `coerce_data_type`

Fixes #911

### DIFF
--- a/src/io/json/read/infer_schema.rs
+++ b/src/io/json/read/infer_schema.rs
@@ -78,12 +78,17 @@ fn infer_number(n: &serde_json::Number) -> DataType {
 }
 
 /// Coerce an heterogeneous set of [`DataType`] into a single one. Rules:
+/// * The empty set is coerced to `Null`
 /// * `Int64` and `Float64` are `Float64`
 /// * Lists and scalars are coerced to a list of a compatible scalar
 /// * Structs contain the union of all fields
 /// * All other types are coerced to `Utf8`
 pub(crate) fn coerce_data_type<A: Borrow<DataType>>(datatypes: &[A]) -> DataType {
     use DataType::*;
+
+    if datatypes.is_empty() {
+        return DataType::Null;
+    }
 
     let are_all_equal = datatypes.windows(2).all(|w| w[0].borrow() == w[1].borrow());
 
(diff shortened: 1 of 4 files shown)


src/io/json/read/infer_schema.rs
```diff
@@ -186,4 +191,15 @@ mod test {
...
         );
     }
+
+    #[test]
+    fn test_coersion_of_nulls() {
+        assert_eq!(coerce_data_type(&[DataType::Null]), DataType::Null);
+        assert_eq!(
+            coerce_data_type(&[DataType::Null, DataType::Boolean]),
+            DataType::Utf8
+        );
```
Roberto-XY, line 201: Is this intended? Should it not be just `DataType::Boolean`?

Similarly, `{"a": null}"` is currently infered as `DataType::Struct(vec![])`and not `DataType::Struct(vec![Field::new("a", DataType::Null, true)])`

jorgecarleitao: Good observation. I think it is intended - `infer_array` does not pass `DataType::Null` to `coerce_data_type` since arrow arrays are nullable and thus we handle `Null` as part before passing to `coerce_data_type`. The idea is [here](https://github.com/jorgecarleitao/arrow2/blob/main/src/io/json/read/infer_schema.rs#L56).

With that said, imo it is not the cleanest design and maybe we should clean it up?

jorgecarleitao (approved):
This looks wonderful. Thanks a lot, very clean and super easy to follow.

Feel welcome to use this project to learn Rust - e.g. via incomplete PRs, PR reviews and/or challenging the design or code. :)

Issue #976, opened 2022-05-04
ValueDictionary offset length < 2 potential panic


TurnOfACard:
I found an issue where a small dictionary offset (length of 1) in `arrow2::io::parquet::read::deserialize::binary::basic::State::OptionalDictionary`  causes the parquet reader to panic.

I imagine restricting the slice bound is a fix. In general, there doesn't appear to be a way to propagate errors without some API changes:

* the function the code is located in (extend_from_state) does not return a Result
* either does its caller: extend_from_new_page

I assume that we should modify the return values of [`extend_from_state`](https://github.com/jorgecarleitao/arrow2/blob/9a386637ccf36355e593bb04407cf31382532914/src/io/parquet/read/deserialize/nested_utils.rs#L316) and [`extend_from_new_page`](https://github.com/jorgecarleitao/arrow2/blob/9a386637ccf36355e593bb04407cf31382532914/src/io/parquet/read/deserialize/binary/basic.rs#L306)?

https://github.com/jorgecarleitao/arrow2/blob/9a386637ccf36355e593bb04407cf31382532914/src/io/parquet/read/deserialize/binary/basic.rs#L331-L340

jorgecarleitao:
Thanks for the report!

The invariant is that `dict_offsets.len()` equals `page.num_values() + 1`, from [here](https://github.com/jorgecarleitao/parquet2/blob/main/src/page/page_dict/binary.rs#L52). So, for `offset.len() == 1`, the dict page is empty? In that case the values page should also also be empty?

Do you have the parquet file available?

TurnOfACard:
Ah, it seems that helps explain part of it.

Long story short, I was using a file produced by arrow2. The first `PageType` was `PageType::DictionaryPage`, and the second was `PageType::DataPageV2`. For the `DictionaryPage`, the `compressed_page_size` was == 9, but `uncompressed_page_size` was == 0 (which may be another problem). Encoding was `RleDictionary`.

As a result of this, `parquet2::page::page_dict::deserialize` was passed an empty `buf` with `num_values` == 0 (plausible if the `uncompressed_page_size` is zero) it will return `(vec![], vec![0])`.

This is ultimately passed to the code below, which will always try to slice dict_values which potentially may be zero-length.

https://github.com/jorgecarleitao/arrow2/blob/9a386637ccf36355e593bb04407cf31382532914/src/io/parquet/read/deserialize/binary/basic.rs#L331-L340

[...]

PR #981, opened 2022-05-06
Aligns MutableDictionaryArray's with MutablePrimitiveArrays with TryPush

TurnOfACard:
Also removes the ability to build an invalid map when pushing individual values. try_push_valid relied on the user pushing a value to the values stack.

Could have contributed to an invalid file was generated and bug found in #976.

### DIFF
--- a/src/array/dictionary/mutable.rs
+++ b/src/array/dictionary/mutable.rs
@@ -3,9 +3,8 @@ use std::{collections::hash_map::DefaultHasher, sync::Arc};
 
 use hash_hasher::HashedMap;
 
-use crate::array::TryExtend;
 use crate::{
-    array::{primitive::MutablePrimitiveArray, Array, MutableArray},
+    array::{primitive::MutablePrimitiveArray, Array, MutableArray, TryExtend, TryPush},
     bitmap::MutableBitmap,
     datatypes::DataType,
     error::{ArrowError, Result},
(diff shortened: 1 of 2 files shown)


src/array/dictionary/mutable.rs
```diff
@@ -181,3 +193,63 @@ where
...
+            }
+        } else {
+            self.push_null();
+            Ok(())
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
jorgecarleitao, line 218: Could we move this to `tests/it/array`? we have all our tests there so that the compilation of the crate is not triggered when only the tests change.

jorgecarleitao (approved):
Looks awesome. Great find!

src/array/dictionary/mutable.rs
TurnOfACard: I can do that. It will need two new `pub(crate)` functions on `MutableDictionaryArray`, which are required to test `TryPush`:

```rust

    /// Returns the dictionary map
    pub(crate) fn map(&self) -> &HashedMap<u64, K> {
        &self.map
    }

    pub(crate) fn keys(&self) -> &MutablePrimitiveArray<K> {
        &self.keys
    }
```

I'll also add an explicit return for the doctest to satisfy Miri.

Issue #988, opened 2022-05-11
Serialising NaN float produces invalid JSON.


SimonSchneider:
Serialising an arrow vector (?) of dataType Float with a `NaN` value produces invalid JSON.

I hope it's ok that I have a test from Polars as that is the way I interact with arrow.
```
    #[test]
    fn serialize_row_with_nan_float() -> Result<()> {
        let mut df = df! {
            "a" => [4.0, 0.0,f64::NAN]
        }?;
        let mut buf = Vec::new();
        JsonWriter::new(&mut buf)
            .with_json_format(JsonFormat::Json)
            .finish(&mut df)?;
        let encoded = std::str::from_utf8(&buf).unwrap();
        dbg!(encoded);
        let test_res = serde_json::json!([4.0, 0.0, f64::NAN]);
        dbg!(test_res);
        let res: serde_json::Value = serde_json::from_slice(&buf)?;
        dbg!(res);
        Ok(())
    }
```
produces:
```
[src/test.rs:775] encoded = "[{\"a\":4.0},{\"a\":0.0},{\"a\":NaN}]"
[src/test.rs:778] test_res = Array([
    Number(
        4.0,
    ),
    Number(
        0.0,
    ),
    Null,
])
Error: expected value at line 1 column 27
```
Above can be seen that the arrow JSON wrote: `{\"a\":NaN}` but the normal rust `serde_json` writes `Null` for the `NaN` value.

If needed I can try rewriting the test into `arrow2` specific code.

jorgecarleitao:
Indeed - well spotted. This is a good first issue - let me know if you would like to PR :)

PR #990, opened 2022-05-16
fix: encode float::NAN as null in json

SimonSchneider:
fixes: #988


jorgecarleitao (approved):
Thanks a lot for the PR. It looks great 💯

Thanks a lot @ritchie46 for the review 🙇

PR #993, opened 2022-05-18
Rename `ArrowError` to `Error`

jorgecarleitao:
The rest of the ecosystem uses `Error` to describe the crate own errors, see e.g. `std::io::Error`, `tokio::time::error::Error`, `serde::de::Error`, `odbc_api::Error`, `lexical_core::Error`.

I would like to propose that we align our error name with the rest of the ecosystem, to reduce the mental load of using this crate and make the crate a bit more standard also in this aspect.

Any thoughts?


ritchie46:
Souns good to me. Easily upgraded with `use arrow::error::Error as ArrowError`.

PR #994, opened 2022-05-18
delay null_counts

ritchie46:
On several `extend` (which may be called in hot loops) there were still `null_count` check on mutable arrays.

Upon freezing a `MutableArray` to `Array` there was also this pattern

```
// first null_count
if mutable_bitmap::null_count > 0 {
  // second null_count on conversion to bitmap
  mutable_bitmap::into_bitmap
}
```

I rewrote it by first doing the `bitmap` coversions (which computes the nulls). And then if the `null_count == 0` then we set `validity = None`.

So basically this PR delays all `null_counts` until we freeze the array as we will compute it upon conversion and can then decide if we can drop the validity or not.

### DIFF
--- a/src/array/utf8/mutable.rs
+++ b/src/array/utf8/mutable.rs
@@ -13,6 +13,7 @@ use crate::{
 
 use super::Utf8Array;
 use crate::array::physical_binary::*;
+use crate::bitmap::Bitmap;
 
 struct StrAsBytes<P>(P);
 impl<T: AsRef<str>> AsRef<[u8]> for StrAsBytes<T> {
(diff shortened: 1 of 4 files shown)


src/array/utf8/mutable.rs
```diff
@@ -36,12 +37,24 @@ impl<O: Offset> From<MutableUtf8Array<O>> for Utf8Array<O> {
         // Safety:
         // `MutableUtf8Array` has the same invariants as `Utf8Array` and thus
         // `Utf8Array` can be safely created from `MutableUtf8Array` without checks.
+        let validity = other
```
ritchie46, line 40: If `null_count == 0` we don't need to set the `validity`

src/array/utf8/mutable.rs
```diff
@@ -368,10 +381,6 @@ impl<O: Offset> MutableUtf8Array<O> {
             self.validity.as_mut().unwrap(),
             iterator,
         );
-
```
ritchie46, line 371: This extend might be called a lot, which means we have quadratic behavior on the `null_count` compute.

ritchie46:
@jorgecarleitao I added an extra test and that required met to clone the `MutableArray`. I derived `Clone` for `MutablePrimitiveArray` and `MutableBitmap` as that seemed a reasonable trait to have to me.

Was there any reason it was not deriving clone other than just not yet added?

jorgecarleitao:
Awesome, @ritchie46 !

No reason - at the time the `MutableBuffer` was not Clone and thus none was clone - meanwhile we moved to `Vec` which is clone and thus everything can be clone ^^

PR #1004, opened 2022-05-22
Rewrite `min_max_string` and `min_max_binary`

HaoYang670:
Signed-off-by: remzi <13716567376yh@gmail.com>

# Changes in the PR
1. Replace `fold` with `reduce`
2. use macro to abstract the common expression
3. Add test to cover the `no null` cases.
4. Add benchmark for `min utf8`
5. Less code is better
6. Functional style

# No obvious performance impact
```
min 2^10 utf8           time:   [33.766 ns 33.786 ns 33.810 ns]                           
                        change: [+0.0167% +0.2385% +0.4298%] (p = 0.02 < 0.05)
                        Change within noise threshold.
  3 (3.00%) high mild
  4 (4.00%) high severe

min null 2^10 utf8      time:   [33.352 ns 33.370 ns 33.391 ns]                                
                        change: [-4.7561% -4.6289% -4.5317%] (p = 0.00 < 0.05)
                        Performance has improved.
  4 (4.00%) high mild
  4 (4.00%) high severe
```

### DIFF
--- a/benches/aggregate.rs
+++ b/benches/aggregate.rs
@@ -43,13 +43,13 @@ fn add_benchmark(c: &mut Criterion) {
             b.iter(|| bench_min(&arr_a))
         });
 
-        let arr_a = create_string_array::<i32>(100, size, 0.0, 0);
+        let arr_a = create_string_array::<i32>(1, size, 0.0, 0);
 
         c.bench_function(&format!("min 2^{} utf8", log2_size), |b| {
             b.iter(|| bench_min(&arr_a))
         });
 
-        let arr_a = create_string_array::<i32>(100, size, 0.1, 0);
+        let arr_a = create_string_array::<i32>(1, size, 0.1, 0);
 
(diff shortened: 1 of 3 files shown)


benches/aggregate.rs
```diff
@@ -42,6 +42,18 @@ fn add_benchmark(c: &mut Criterion) {
         c.bench_function(&format!("min null 2^{} f32", log2_size), |b| {
             b.iter(|| bench_min(&arr_a))
         });
+
+        let arr_a = create_string_array::<i32>(1, size, 0.0, 0);
```
HaoYang670, line 46: I set the length of each string to be 1 to make the time of comparing each element to be determined.

jorgecarleitao:
Thanks a lot for this PR.

I went through this and also ran the benches on main vs this PR with different string lengths and array lengths and the performance is equivalent on my machine.

Given that this is less code and more tested, we ship it :)

PR #1005, opened 2022-05-23
Speed up `min_max_boolean`

HaoYang670:
Signed-off-by: remzi <13716567376yh@gmail.com>

Changes in the PR:
For a Boolean array without `null` value. We can simply get the max value by testing `array.values.null_count < array.length` and get the min value by testing `array.values.null_count == 0`. 
Theoretically, the time complexity is `O(1)` as `null_count` is pre computed.


jorgecarleitao:
Thanks a lot, @HaoYang670 , indeed it is a quite faster case :)

PR #1006, opened 2022-05-23
Make convert mod public to expose parquet_to_arrow_schema utility function

martingallagher:
Exposes the utility function to go from `parquet2`'s `SchemaDescriptor` to `arrow2` `Schema`

Maybe a `arrow2::schema::Schema::try_from_message` is more desirable?


jorgecarleitao:
Hey, thanks a lot for this!

Would it be possible to use `pub use convert::parquet_to_arrow_schema` instead? It makes it a bit easier to follow what is being exported exactly when reading the code.

Issue #903, opened 2022-03-11
Panic when printing debug arrays with invalid timezones


dbr:
Bit tricky to untangle into a simple repro case, but hopefully enough info:

If I create a native datetime object via a convoluted path of Python into arrow2, something vaguely along these lines:

```python
import pandas
import datetime

df = pandas.DataFrame.from_dict({'a': [datetime.datetime.now()]})

custom_method_from_pandas_to_pyarrow_to_arrow2(df)
```

..then I get the following panic:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidArgumentError("timezone \"\" cannot be parsed")', /home/dbr/.cargo/registry/src/github.com-1ecc6299db9ec823/arrow2-0.9.2/src/array/display.rs:68:82
```
..which is happening on the 0.9 equivelant of this line:

https://github.com/jorgecarleitao/arrow2/blob/4b893b71ddb73bfe9e7a0d76fbf6fe4ca0e9a930/src/array/primitive/fmt.rs#L69

Oddly, this wasn't happening when we were using 0.8.1.. but the `get_display` code seems basically identical in that version.. That said I've changed quite a lot of our arrow/Python integration code between then, so I need to investigate further what caused this - regardless, it seems surprising the `get_display` method would panic like this

jorgecarleitao:
Thanks a lot for reporting it 🙇 

Do you have a suggestion on what the intended behavior should be? Some ideas:

* write something like `"ERROR: can't parse timezone ""` in each array's item
* present the value in rfc3339 but present an "ERROR" in the timezone

ideas?

houqp:
Based on arrow spec, empty string as timezone should be considered as invalid datatype? It seems reasonable to panic or return a Result::Err when invalid data are encountered. The timestamp datatype validation probably should have happened much earlier in the pipeline, ideally when the timestamp array is being created.

dbr:
> Do you have a suggestion on what the intended behavior should be?

Hmm I'm not sure - with the 0.8 API of `arrow2::array::get_display(col)(idx)` I would expect it to return a `Result` if it does any non-trivial parsing (as is done with the timezones)

However with the recent changes to use this code in `Debug`, it's harder to say. Panicing is not ideal at all for us (since the data is coming from user, and we are using the `get_display` call in a UI)

[...]

PR #1009, opened 2022-05-26
Treat empty timezone string as no-timezone

dbr:
As per https://github.com/jorgecarleitao/arrow2/issues/903#issuecomment-1138123945


jorgecarleitao:
Thank you! The code is much cleaner, also!

PR #1013, opened 2022-05-26
Fixed panic on debug print of invalid timezones

### LINKED ISSUE #903: Panic when printing debug arrays with invalid timezones

jorgecarleitao:
Closes #903

### DIFF
--- a/src/array/primitive/fmt.rs
+++ b/src/array/primitive/fmt.rs
@@ -66,15 +66,26 @@ pub fn get_write_value<'a, T: NativeType, F: Write>(
                     }
                     #[cfg(feature = "chrono-tz")]
                     Err(_) => {
-                        let timezone = temporal_conversions::parse_offset_tz(tz).unwrap();
-                        dyn_primitive!(array, i64, |time| {
-                            temporal_conversions::timestamp_to_datetime(time, *time_unit, &timezone)
-                        })
+                        let timezone = temporal_conversions::parse_offset_tz(tz);
+                        match timezone {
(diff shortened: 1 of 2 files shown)


dbr:
Tested, and change looks good to me :partying_face: 

It's a bit of a shame to lose the "Invalid Offset format (must be [-]00:00) or chrono-tz feature not active" message, but not panicing is a huge improvement, and keeping this message doesn't seem worth the significant change required I think!

PR #1016, opened 2022-05-27
Added support for custom sort `build_compare_fn`

b41sh:
The `build_compare` function only supports build compare functions for common data types and will generate errors for other data types like `DataType::Extension`. We have some custom data types stored as `DataType::Extension` need to be sorted, but there is no proper way to implement them.

This PR adds the following functions to support the user implement custom `build_compare` function for their own data types.
```rust
pub fn lexsort_to_indices_impl<I: Index>(
    columns: &[SortColumn],
    limit: Option<usize>,
    build_compare_fn: &dyn Fn(&dyn Array, &dyn Array) -> Result<DynComparator>,
) -> Result<PrimitiveArray<I>> {
}

pub(crate) fn build_compare_impl(
    array: &dyn Array,
    sort_option: SortOptions,
    build_compare_fn: &dyn Fn(&dyn Array, &dyn Array) -> Result<DynComparator>,
) -> Result<DynComparator> {
}

pub fn build_comparator_impl<'a>(
    pairs: &'a [(&'a [&'a dyn Array], &SortOptions)],
    build_compare_fn: &dyn Fn(&dyn Array, &dyn Array) -> Result<DynComparator>,
) -> Result<Comparator<'a>> {
}
```

### DIFF
--- a/src/compute/merge_sort/mod.rs
+++ b/src/compute/merge_sort/mod.rs
@@ -470,6 +470,15 @@ type IsValid<'a> = Box<dyn Fn(usize) -> bool + 'a>;
 pub fn build_comparator<'a>(
     pairs: &'a [(&'a [&'a dyn Array], &SortOptions)],
 ) -> Result<Comparator<'a>> {
+    build_comparator_impl(pairs, &build_compare)
+}
+
+/// returns a comparison function between any two arrays of each pair of arrays, according to `SortOptions`.
+/// Implementing custom `build_compare_fn` for unsupportd data types.
+pub fn build_comparator_impl<'a>(
+    pairs: &'a [(&'a [&'a dyn Array], &SortOptions)],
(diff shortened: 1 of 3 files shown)


jorgecarleitao:
Thanks a lot!

Just to understand, are the extension types using the same comparison as the datatype that they are wrapping, or is the comparison entirely different?

b41sh:
> Thanks a lot!
> 
> Just to understand, are the extension types using the same comparison as the datatype that they are wrapping, or is the comparison entirely different?

The datatype is actually similar to JSON and stored as `LargeBinary` in arrow2. We define a set of comparison rules for it, which is entirely different from the inner datatype.

jorgecarleitao:
Thanks for the explanation. Got it. Yeap, makes a lot of sense this implementation, then!

PR #1042, opened 2022-06-02
Proposal: Migrate from `Arc<dyn Array>` to `Box<dyn Array>`

jorgecarleitao:
This PR is a backward incompatible change to extend the support for clone-on-write semantics
initially added by @ritchie46 on #794 to arrays.

The migration is simple - replace `Arc<dyn Array>` by `Box<dyn Array>`.

# Background

`Arc` main goal is to enable sharing of data, typically because cloning is expensive. The tradeoff is that `Arc`s usually represent immutable data.

All our arrays are quite small structs. Their underlying data is usually a `DataType` and one or two `Arc`s that store the data itself.

Thus, `Arc`ing `Array` does not add much value. OTOH, it makes the API clumsy, since we need to remember to arc everything. It also aludes users to the immutable nature of the Array. Furthermore, it makes it challenging to offer a good support for clone-on-write semantics to arrays since we need to check if the array is being shared, and clone its internals if yes.

Currently,
* all our nested/child arrays are stored as `Arc<dyn Array>`
* all our IO interfaces, including FFI, received and output an `Arc<dyn Array>`

My understanding is that there three main reasons for us to be in this situation:

1. Rust support for cloning `Box<dyn T>` where `T: Clone` is not great - it is essentially not possible in `std`.
2. historically `arrow-rs` used to store everything under `Arc` (e.g. `ArrayData`, `Array`, data itself).
3. The C++ implementation, which `arrow-rs` is inspired on, seems to use `Arc` (`shared_ptr`)

# This PR

This PR makes
* child arrays to be `Box<dyn Array>` instead of `Arc<dyn Array>`
* all IO consume and return `Box<dyn Array>` instead of `Arc<dyn Array>`

The major benefit of this behavior is that it makes it quite easy to add first class support clone-on-write semantics to the different Array APIs. As an example, with this PR we can now more easily:

* read a primitive array from parquet (`Box<dyn Array>`)
* evaluate a complex runtime expression on it (e.g. `((a * 10) + 2) * exp(-10 * a))`) without any extra allocation

This PR also adds an example illustrating how the API is used.

### DIFF
--- a/examples/parquet_write_parallel/src/main.rs
+++ b/examples/parquet_write_parallel/src/main.rs
@@ -1,6 +1,5 @@
 //! Example demonstrating how to write to parquet in parallel.
 use std::collections::VecDeque;
-use std::sync::Arc;
 
 use rayon::prelude::*;
 
@@ -12,7 +11,7 @@ use arrow2::{
     io::parquet::{read::ParquetError, write::*},
 };
 
-type Chunk = AChunk<Arc<dyn Array>>;
+type Chunk = AChunk<Box<dyn Array>>;
 
 struct Bla {
     columns: VecDeque<CompressedPage>,
@@ -134,7 +133,14 @@ fn create_batch(size: usize) -> Result<Chunk> {
         })
         .collect();
 
+<<<<<<< HEAD
     Chunk::try_new(vec![c1.arced(), c2.arced()])
+=======
+    Chunk::try_new(vec![
+        Box::new(c1) as Box<dyn Array>,
+        Box::new(c2) as Box<dyn Array>,
+    ])
+>>>>>>> 619252ff0 (Arc->Box)
 }
 
 fn main() -> Result<()> {
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ bench = false
 [dependencies]
 either = "1.6"
 num-traits = "0.2"
+dyn-clone = "1"
 bytemuck = { version = "1", features = ["derive"] }
 chrono = { version = "0.4", default_features = false, features = ["std"] }
 chrono-tz = { version = "0.6", optional = true }
--- a/arrow-parquet-integration-testing/src/main.rs
+++ b/arrow-parquet-integration-testing/src/main.rs
@@ -1,3 +1,6 @@
+use std::fs::File;
+use std::{collections::HashMap, io::Read};
+
 use arrow2::array::Array;
 use arrow2::io::ipc::IpcField;
 use arrow2::{
@@ -15,15 +18,12 @@ use arrow2::{
 };
 use clap::Parser;
 use flate2::read::GzDecoder;
-use std::fs::File;
-use std::sync::Arc;
-use std::{collections::HashMap, io::Read};
 
 /// Read gzipped JSON file
 pub fn read_gzip_json(
     version: &str,
     file_name: &str,
-) -> Result<(Schema, Vec<IpcField>, Vec<Chunk<Arc<dyn Array>>>)> {
+) -> Result<(Schema, Vec<IpcField>, Vec<Chunk<Box<dyn Array>>>)> {
     let path = format!(
         "../testing/arrow-testing/data/arrow-ipc-stream/integration/{}/{}.json.gz",
         version, file_name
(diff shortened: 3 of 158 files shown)


jorgecarleitao:
@houqp, @Dandandan, @dbr, @sundy-li would you be available to review / comment / help on this one? We are trying to significantly improve the clone-on-write story, which is a bottleneck in Arrow, but would like to double-check with you that this does no harm.

jorgecarleitao:
may be of interest to @wesm also - I recall that Voltron was looking around avoiding allocations on compute kernels - the main point is (in C++ notation): since data is already inside a `shared_ptr` inside `ArrayData`, we can use `unique_ptr<Array>` instead of `shared_ptr<Array>` - cloning the Array is cheap anyways. The advantage is that we can easily access the array's content and thus can mutate the array itself in-place.

In cases (basically everything coming from parquet, IPC, etc) where data is not shared (which is often), this has dramatic effects since it can lead to a allocation-free compute for math.

E.g. expressions like `exp(-(arr + 1)) * 2 + 1` (5 ops) can be done without allocations, still leverage SIMD, and do not require emitting new LLVM via e.g. Grandiva.

jorgecarleitao:
#1061 motivates this PR (~2x speedup on horizontal ops, ~5-10x on horizontal logic ops)

dbr:
This change sounds perfectly fine from our point of view - we mostly interact with arrow2 data either in read-only fashion or via FFI (and the changes to FFI seem pretty trivial to handle) - so this doesn't really impact us directly, and if it's faster and generally better, then, :partying_face: 

One thing I don't quite understand is,

> Thus, Arcing Array  [...] makes it challenging to offer a good support for clone-on-write semantics to arrays since we need to check if the array is being shared, and clone its internals if yes

I thought `Arc` provides clone-on-write? Specifically `Arc::make_mut(...)` docs say "If there are other Arc pointers to the same allocation, then make_mut will clone the inner value to a new allocation to ensure unique ownership. This is also referred to as clone-on-write"

jorgecarleitao:
@dbr, great question.

Let's consider the example added on this PR with an `Arc` instead of a `Box`:

```
// This example demos how to operate on arrays in-place.
use std::sync::Arc;
use arrow2::array::{Array, PrimitiveArray};

// say we have have received an array
let mut array: Arc<dyn Array> = PrimitiveArray::from_vec(vec![1i32, 2]).arced();

// we can apply a transformation to its values without allocating a new array as follows:
// 1. downcast it to the correct type (known via `array.data_type().to_physical_type()`)
let array = Arc::make_mut(&mut array)
    .as_any_mut()
    .downcast_mut::<PrimitiveArray<i32>>()
    .unwrap();

// 2. call `apply_values` with the function to apply over the values
array.apply_values(|x| x.iter_mut().for_each(|x| *x *= 10));
```

This unfortunately does not compile because `Arc::<T>::make_mut` requires `T: Clone`, but `dyn Array` is not clonable (since it is not `Size`). The crate `dyn_clone` offers a safe implementation of `Clone` for `Box<dyn T>` when `T: Clone`. We could get away with something like

```
// say we have have received an array
let array: Arc<dyn Array> = PrimitiveArray::from_vec(vec![1i32, 2]).arced();

// we can apply a transformation to its values without allocating a new array as follows:
// 1. clone the internals
let mut array: Box<dyn Array> = arrow2::array::clone(array.as_ref());

// 2. downcast it to the correct type (known via `array.data_type().to_physical_type()`)
let array = array
    .as_any_mut()
    .downcast_mut::<PrimitiveArray<i32>>()
    .unwrap();

// 3. call `apply_values` with the function to apply over the values
array.apply_values(|x| x.iter_mut().for_each(|x| *x *= 10));
```

but here we need to be careful - if we would have named the variable `array1` instead of `array`, in step 3. both arrays are in scope and share the underlying data (due to `clone`). Thus, `apply_values` would result in a clone of `O(N)`, since the values are now shared between the two arrays.

sundy-li:
Looks great to me.

elferherrera:
Silly question about this change. I understood that the use of `Arc` in arrow was to have a `Sync` and `Send` array. Will `Box` provide the same functionality?

joshuataylor:
Not sure if changelog is generated at version release time, but the only change I needed to make for my code that relied on chunks was to change Arc -> Box and it worked. :rocket:

examples/parquet_write_parallel/src/main.rs
```diff
@@ -134,7 +133,14 @@ fn create_batch(size: usize) -> Result<Chunk> {
         })
         .collect();
 
+<<<<<<< HEAD
```
cjermain, line 136: It looks like merge conflicts were still in the commit. @jorgecarleitao

jorgecarleitao: 6jiwtn

PR #1046, opened 2022-06-03
Fixed using lower limit than size of first parquet row group

arxra:
If the first row-group of parquet data had more rows than the limit, no data would be returned even if the chunk size was available in that limit and first row group

### DIFF
--- a/src/io/parquet/read/file.rs
+++ b/src/io/parquet/read/file.rs
@@ -80,7 +80,7 @@ impl<R: Read + Seek> FileReader<R> {
             metadata: schema_metadata,
         };
 
-        let mut row_groups = RowGroupReader::new(
+        let row_groups = RowGroupReader::new(
             reader,
             schema,
             groups_filter,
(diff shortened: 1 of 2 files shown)


src/io/parquet/read/file.rs
```diff
@@ -80,20 +80,21 @@ impl<R: Read + Seek> FileReader<R> {
...
-        let row_groups = RowGroupReader::new(
+        let mut row_groups = RowGroupReader::new(
             reader,
             schema,
             groups_filter,
             metadata.row_groups.clone(),
             chunk_size,
             limit,
         );
+        let current_row_group = row_groups.next().transpose()?;
```
jorgecarleitao, line 91: I think we should consider something different here - this causes `try_new` to be `O(N)` since it advances the iterator.

jorgecarleitao (commented):
Thanks a lot for the PR!

I agree that there is an issue here. I left a comment as I think we should avoid advancing the iterator on `try_new`. Is there any way around that?

arxra:
@jorgecarleitao This alternative instead looks at the current_row_group, which if None should not update the amount of remaining rows as no rows will be read on that iteration! This would still work if given a empty row group, but most importantly its None on initialization for reading the first one.

jorgecarleitao (approved):
Great solution to this! Thanks a lot again.

(minor fmt error - let me know if you would like me to fix it)

arxra:
Fixed :)

PR #1051, opened 2022-06-06
Added support to read and write f16

### LINKED ISSUE #1050: Float16 Support
Hello,

Thanks for the great crate!

I have some files in Apache feather that are using FP16. What are your thoughts on integrating FP16 support with https://github.com/starkat99/half-rs? Are there major technical limitations / downsides to this?

jorgecarleitao:
Some stats: the size of the lib (in `.so`) is increased by 90Kb

* main `.so`: `9.384.144`
* this `.so`: `9.492.912`

it adds no extra dependencies.

The idea is to provide the very minimal functionality to read and write `f16` from and to IPC, since `f16` is not really used in Rust. 

Users that wish more functionality over `f16`, can use [`half`](https://docs.rs/half/latest/half/)'s `f16::from_bits` and `to_bits`:

```rust
use arrow2::array::Float16Array;
use arrow2::types::f16;

fn main() {
    let data = vec![f16(half::f16::from_f32(0.1f32).to_bits()); 10];
    let array = Float16Array::from_vec(data);
    println!("{:?}", array);
}
```

However, when you only need to read `f16` from IPC and convert it to `f32`, you can use `arrow2::types::f16::as_f32()`, e.g.

```rust
use arrow2::array::{Float16Array, Float32Array};
use arrow2::datatypes::DataType;
use arrow2::types::f16;

fn main() {
    let array = Float16Array::from_vec(vec![f16(100); 10]);

    let f32_values = array
        .values()
        .iter()
        .map(|x| x.as_f32())
        .collect::<Vec<_>>();

    let new_array = Float32Array::new(
        DataType::Float32,
        f32_values.into(),
        array.validity().cloned(),
    );
    println!("{:?}", new_array);
}
```

Closes #1050


ritchie46:
Right, so we don't have to implement the full array trait and not all compute and thus can save binary bloat?

jorgecarleitao:
This does implement the trait for `f16`, mutable and non-mutable, growable, etc. The primary concern here was to not bring `half`'s dependency to core, which I believe is no justified just to read f16 from IPC.

In other words, our core has so few requirements that we can get away with 100 LOC to support a primitive type in core. ^^

ritchie46:
> In other words, our core has so few requirements that we can get away with 150 LOC to support a primitive type in core. ^^

That certainly is impressive! 

I guess we only need to watch out to not include it in a macro! that downcasts to primitive for any of the compute kernels. We could accidentally compile for f16 as well, whereas that binary code would not be used.

PR #1059, opened 2022-06-08
prevent unneeded offset check

ritchie46:
Calling `ListArray::new` does `O(n)` work where `n = offset_length`. 

However, this work can be prevented in a few cases because we convert from a struct that has the same invariants.

The same logic applies to a Utf8 conversion, where we just determined the offsets and converted `&str` to bytes, so a `utf8` check is redundant.

### DIFF
--- a/src/array/list/mutable.rs
+++ b/src/array/list/mutable.rs
@@ -50,12 +50,16 @@ impl<O: Offset, M: MutableArray + Default> Default for MutableListArray<O, M> {
 
 impl<O: Offset, M: MutableArray> From<MutableListArray<O, M>> for ListArray<O> {
     fn from(mut other: MutableListArray<O, M>) -> Self {
-        ListArray::new(
-            other.data_type,
-            other.offsets.into(),
-            other.values.as_arc(),
-            other.validity.map(|x| x.into()),
-        )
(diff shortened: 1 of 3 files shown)


ritchie46:
Hmm.. it fails on `forbid_unsafe` in avro. I can remove it there.

ritchie46:
We could actually abstract these unsafe calls by having two newtype structs

```rust
#[repr(transparent)]
struct OffsetBuffer<O: Offset>(Vec<O>)

impl OffsetBuffer {

  fn add_lenght(&mut self, len: usize);
  
  unsafe fn from_vec();
}

// this could also be just `String`
// but maybe this is easier with delaying Utf8 validation
#[repr(transparent)]
struct StringBuffer(Vec<u8>)

impl StringBuffer {

   fn push_str(&mut self, val: &str);
   
   unsafe fn push_bytes(&mut self, bytes: &[u8]);
   
   unsafe fn from_vec();
}
```

A `MutableList` can then call `new` and pass on its internal state with the given invariants. That way you can keep the `io` module unsafe free without doing redundant checks.

jorgecarleitao:
Great idea - we can offer a safe API around those!

PR #1066, opened 2022-06-10
Added cast to/from `Null` from/to every type

jorgecarleitao:
We limited what types we accepted on the cast, but do support any type :)


ritchie46:
That's useful!

PR #1067, opened 2022-06-10
Improved performance in lexical write (~5%)

ritchie46:
It might be interesting to explore a slightly adapted version of `BufStreamingIterator`.

If I am correct it currently writes to an internal buffer, and then copies that buffer to the `W: Write` buffer. If we would accept that final buffer in the `next()` method as argument, we could directly write to that.

Will explore that idea later.

### DIFF
--- a/src/util/lexical.rs
+++ b/src/util/lexical.rs
@@ -22,8 +22,16 @@ pub fn lexical_to_bytes_mut<N: lexical_core::ToLexical>(n: N, buf: &mut Vec<u8>)
 
         //  Safety:
         //  Omits an unneeded bound check as we just ensured that we reserved `N::FORMATTED_SIZE_DECIMAL`
-        let len = lexical_core::write_unchecked(n, slice).len();
-        buf.set_len(len);
+        #[cfg(debug_assertions)]
+        {
+            let len = lexical_core::write(n, slice).len();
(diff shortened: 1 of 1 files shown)


jorgecarleitao:
Looks reasonable! Do you have a bench on performance delta? Curious how it impacts.

ritchie46:
Yup

```
bench_1                 time:   [1.0596 ms 1.0599 ms 1.0601 ms]                     
                        change: [-5.8026% -5.3279% -4.9843%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  4 (4.00%) low mild
  2 (2.00%) high mild

```

```rust
pub fn compute(vals: &[i32], buf: &mut Vec<u8>)  {
    for v in vals {
        memcheck::lexical_to_bytes_mut_unchecked(*v, buf)
    }
}

fn add_benchmark(c: &mut Criterion) {

    let mut rng = thread_rng();
    let values = (0..100000)
        .map(|_| rng.sample(Standard))
        .collect::<Vec<i32>>();

    let mut buf = vec![];

    c.bench_function("bench_1", |b| {
        b.iter(|| black_box(compute(&values, &mut buf)))
    });
}
```

PR #1075, opened 2022-06-15
Panic-free read of IPC files

jorgecarleitao:
This PR safeguards the IPC interface from panicking or aborting (OOM) on invalid data.

* All number casts (e.g. `i32` to `usize`) and operations (e.g. `*`) are now protected to avoid overflow and thus trigger OOM
* Many of the errors were converted to an enum for more clarity over what failed exactly
* the sum of size of all IPC buffers can no longer be larger than the size of the file, thus protecting us from OOM attacks


jorgecarleitao:
Thanks @kristoff3r and @TethysSvensson for the quick fix and release of https://github.com/planus-org/planus/issues/99 !

PR #1076, opened 2022-06-15
Improve `arity_assign`: `~2x` improvement if we share data.

ritchie46:
This continues upon #1073 (you may close that one if you prefer in one PR, otherwise I rebase).

I did a local benchmark to see what the cost of the new `arity_assign` would be in the case we share data. In polars expressions that is every first operation:

```
bench_1                 time:   [608.34 us 609.30 us 610.73 us]                    
                        change: [+44.741% +45.228% +45.698%] (p = 0.00 < 0.05)
                        Performance has regressed.
  2 (2.00%) high mild
  5 (5.00%) high severe
```

This showed, that in case of shared data, we first do a `memcpy` and this is `~2x` slower than the kernels where we allocate a new memory buffer.

Luckily we can have best of both worlds. :)

This PR branches depending on the shared state:

* in case we share state -> we alloc (and thus choose the fastest path possible)
* in case we own -> we mutate in place (and thus choose the fastest path possible)

I also added a distinction between `set_validity` and `with_validity` as we now have mutable data, it is not always logical return a new `PrimitiveArray`.

### DIFF
--- a/benches/assign_ops.rs
+++ b/benches/assign_ops.rs
@@ -14,7 +14,7 @@ fn add_benchmark(c: &mut Criterion) {
         c.bench_function(&format!("apply_mul 2^{}", log2_size), |b| {
             b.iter(|| {
                 criterion::black_box(&mut arr_a)
-                    .apply_values(|x| x.iter_mut().for_each(|x| *x *= 1.01));
+                    .apply_values_mut(|x| x.iter_mut().for_each(|x| *x *= 1.01));
                 assert!(!arr_a.value(10).is_nan());
             })
         });
@@ -30,7 +30,7 @@ fn add_benchmark(c: &mut Criterion) {
         let mut arr_a = create_primitive_array::<f32>(size, 0.2);
         let mut arr_b = create_primitive_array_with_seed::<f32>(size, 0.2, 10);
         // convert to be close to 1.01
-        arr_b.apply_values(|x| x.iter_mut().for_each(|x| *x = 1.01 + *x / 20.0));
+        arr_b.apply_values_mut(|x| x.iter_mut().for_each(|x| *x = 1.01 + *x / 20.0));
 
         c.bench_function(&format!("apply_mul null 2^{}", log2_size), |b| {
             b.iter(|| {
--- a/src/array/primitive/mod.rs
+++ b/src/array/primitive/mod.rs
@@ -511,3 +511,9 @@ pub type UInt16Vec = MutablePrimitiveArray<u16>;
 pub type UInt32Vec = MutablePrimitiveArray<u32>;
 /// A type definition [`MutablePrimitiveArray`] for `u64`
 pub type UInt64Vec = MutablePrimitiveArray<u64>;
+
+impl<T: NativeType> Default for PrimitiveArray<T> {
+    fn default() -> Self {
+        PrimitiveArray::new(DataType::Null, vec![].into(), None)
+    }
+}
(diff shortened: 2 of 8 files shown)


src/array/primitive/mod.rs
```diff
@@ -245,12 +245,41 @@ impl<T: NativeType> PrimitiveArray<T> {
...
+    pub fn with_values(&self, values: Vec<T>) -> Self {
+        let mut out = self.clone();
+        out.set_values(values);
+        out
+    }
+
+    /// Update the values buffer of this [`PrimitiveArray`].
+    /// # Panics
+    /// This function panics iff `values.len() != self.len()`.
+    pub fn set_values(&mut self, values: Vec<T>) {
```
jorgecarleitao, line 276:
```suggestion
    pub fn set_values(&mut self, values: Buffer<T>) {
```

src/array/primitive/mod.rs
```diff
@@ -245,12 +245,41 @@ impl<T: NativeType> PrimitiveArray<T> {
...
+        assert_eq!(
+            values.len(),
+            self.len(),
+            "values length should be equal to this arrays length"
+        );
+        self.values = values.into();
```
jorgecarleitao, line 282:
```suggestion
        self.values = values;
```

src/array/primitive/mutable.rs
```diff
@@ -81,6 +81,32 @@ impl<T: NativeType> MutablePrimitiveArray<T> {
...
+    ///
+    /// This is an API to leverage clone-on-write
+    /// # Implementation
+    /// This function is `O(f)` if the data is not being shared, and `O(N) + O(f)`
+    /// if it is being shared (since it results in a `O(N)` memcopy).
+    /// # Panics
+    /// This function panics if the function modifies the length of the [`MutableBitmap`].
+    pub fn apply_validity<F: Fn(&mut MutableBitmap)>(&mut self, f: F) {
+        if let Some(validity) = &mut self.validity {
+            f(validity);
```
jorgecarleitao, line 107: we need to check that the validity's length remains the same, I believe.

jorgecarleitao (approved):
Love it! This is great!

src/array/primitive/mutable.rs
ritchie46: Check I will add that. 👍

src/compute/arity_assign.rs
```diff
@@ -48,10 +49,28 @@ where
...
-            .for_each(|(l, r)| *l = op(*l, *r))
-    });
+    match owned_lhs.into_mut() {
+        Either::Left(mut immutable) => {
+            let values = immutable
+                .values()
+                .iter()
+                .zip(rhs.values().iter())
+                .map(|(l, r)| op(*l, *r))
+                .collect::<Vec<_>>();
```
jorgecarleitao, line 63:
```suggestion
                .collect()
                .into();
```

ritchie46: The compiler cannot infer the type if we don't specify we want to collect into a `Vec`

ritchie46:
@jorgecarleitao in light of #1078 and the possible `into_iter().collect()` over bits call, 

I added the allocation branch for the validity as well.

benches/assign_ops.rs
```diff
@@ -14,7 +14,7 @@ fn add_benchmark(c: &mut Criterion) {
         c.bench_function(&format!("apply_mul 2^{}", log2_size), |b| {
             b.iter(|| {
                 criterion::black_box(&mut arr_a)
-                    .apply_values(|x| x.iter_mut().for_each(|x| *x *= 1.01));
+                    .apply_values_mut(|x| x.iter_mut().for_each(|x| *x *= 1.01));
```
ritchie46, line 17: `apply_values_mut` will do the memcopy no matter what state.

I've added an `apply_values` which let's the caller decide if the the mutable path is worth it. Both functions clean up and check if the invariants hold.

PR #1078, opened 2022-06-16
Fixed panic in bitmap assign_mut

ritchie46:
~The `length` taken into account to compute the remainder bytes could differ from the length of the buffer (e.g. in slicing operations).~

~This ensures we do the computations on the length of the buffer and we've added a test to ensure we don't regress.~

When calling `into_mut` on a `Bitmap` and the data has an `offset` we `memcpy` the bytes into a `MutableBitmap`. However this is incorrect, because a `MutableBitmap` does not have an offset.

This fixes that by iterating every `bit` and recreating the proper bytes for the `MutableBitmap`.

This is expensive, so maybe we should consider adding an `offset` on `MutableBitmap` as well.

### DIFF
--- a/src/bitmap/immutable.rs
+++ b/src/bitmap/immutable.rs
@@ -239,7 +239,12 @@ impl Bitmap {
     pub fn make_mut(self) -> MutableBitmap {
         match self.into_mut() {
             Either::Left(data) => {
-                MutableBitmap::from_vec(data.bytes.as_ref().to_vec(), data.length)
+                if data.offset > 0 {
+                    // we have to recreate the bytes because a MutableBitmap does not have an `offset`.
+                    data.iter().collect()
+                } else {
+                    MutableBitmap::from_vec(data.bytes.as_ref().to_vec(), data.length)
+                }
             }
             Either::Right(data) => data,
         }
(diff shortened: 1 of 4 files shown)


src/bitmap/immutable.rs
```diff
@@ -239,7 +239,12 @@ impl Bitmap {
     pub fn make_mut(self) -> MutableBitmap {
         match self.into_mut() {
             Either::Left(data) => {
-                MutableBitmap::from_vec(data.bytes.as_ref().to_vec(), data.length)
+                if data.offset > 0 {
+                    // we have to recreate the bytes because a MutableBitmap does not have an `offset`.
+                    data.iter().collect()
```
jorgecarleitao, line 244: I think we can improve this further by using `iter_chunk`, which will iterate in chunks of `u64`. We have a function somewhere to create a `MutableBitmap` from chunks.

jorgecarleitao:
Just to double check, before the test would panic (not trigger UB), right?

ritchie46:
> Just to double check, before the test would panic (not trigger UB), right?

Yes

src/bitmap/immutable.rs
ritchie46: I tried this, but that seems to not include the `remainder`

```rust
                 let chunks = BitChunks::<u64>::new(data.bytes.as_ref(), data.offset, data.length);
                    chunk_iter_to_vec(chunks)
```

PR #1085, opened 2022-06-19
Removed un-necessary allocation in `assign_ops`

jorgecarleitao:
This PR removes unnecessary allocation on `assign_ops`.

When converting `PrimitiveArray -> MutablePrimitiveArray`, we need to check that both the values and validities are unique refs. However, when operating over values and validities independently, we do not need a mutable ref to both at the same time.

A common case where this was hitting was when `rhs` had a validity and `lhs` did not. In this case, we first cloned the `rhs`'s validity for `lhs`, but them tried to get a mut ref to both the validity and the values (to only use the values). This would naturally fail, causing us to have to re-use the `lhs`.

This PR adds `PrimitiveArray::get_mut_values` and removes (un-released) functions from `PrimitiveArray`, since the function `get_mut_values` is more generic.

### DIFF
--- a/src/array/primitive/mod.rs
+++ b/src/array/primitive/mod.rs
@@ -1,7 +1,7 @@
 use crate::{
     bitmap::{
         utils::{zip_validity, ZipValidity},
-        Bitmap, MutableBitmap,
+        Bitmap,
     },
     buffer::Buffer,
     datatypes::*,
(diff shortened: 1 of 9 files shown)


src/compute/arity_assign.rs
```diff
@@ -18,14 +20,22 @@ where
     I: NativeType,
     F: Fn(I) -> I,
 {
-    array.apply_values_mut(|values| values.iter_mut().for_each(|v| *v = op(*v)));
+    if let Some(values) = array.get_mut_values() {
+        // mutate in place
+        values.iter_mut().for_each(|l| *l = op(*l));
+    } else {
+        // alloc and write to new region
+        let values = array.values().iter().map(|l| op(*l)).collect::<Vec<_>>();
```
jorgecarleitao, line 28: this is the same optimization that @ritchie46 used on #1076 for binary.

src/compute/arity_assign.rs
```diff
@@ -48,54 +58,39 @@ where
...
             });
         }
-    }
+    };
 
-    // we need to take ownership for the `into_mut` call, but leave the `&mut` lhs intact
-    // so that we can later assign the result to out `&mut lhs`
-    let owned_lhs = std::mem::take(lhs);
-
-    *lhs = match owned_lhs.into_mut() {
```
jorgecarleitao, line 79: this is the important change - this was trying to get a `MutableBitmap`, but we do not need one - we only need a mut reference to the values.

ritchie46:
> However, when operating over values and validities independently, we do not need a mutable ref to both at the same time.

Good observation.

src/array/primitive/mod.rs
```diff
@@ -283,59 +283,24 @@ impl<T: NativeType> PrimitiveArray<T> {
...
-    /// This function is `O(f)` if the data is not being shared, and `O(N) + O(f)`
-    /// if it is being shared (since it results in a `O(N)` memcopy).
     /// # Panics
-    /// This function panics if the function modifies the length of the [`MutableBitmap`].
-    pub fn apply_validity<F: Fn(&mut Bitmap)>(&mut self, f: F) {
-        if let Some(validity) = self.validity.as_mut() {
-            f(validity);
+    /// This function panics if the function `f` modifies the length of the [`Bitmap`].
+    pub fn apply_validity<F: Fn(Bitmap) -> Bitmap>(&mut self, f: F) {
+        if let Some(validity) = std::mem::take(&mut self.validity) {
```
ritchie46, line 293: Right, so we let the called decide if it's worth it to make it a mutable bitmap or not? That reduces some function bloat indeed. :+1:

src/compute/arity_assign.rs
ritchie46: Yep, good catch.

src/array/primitive/mutable.rs
```diff
@@ -90,27 +90,9 @@ impl<T: NativeType> MutablePrimitiveArray<T> {
     /// This function is `O(f)` if the data is not being shared, and `O(N) + O(f)`
     /// if it is being shared (since it results in a `O(N)` memcopy).
     /// # Panics
-    /// This function panics, if `f` modifies the length of `&mut [T]`
+    /// This function panics iff `f` panics
```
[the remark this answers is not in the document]
jorgecarleitao: Can we mutate the length of `&mut [T]`? I thought we could only mutate its contents. We would have this problem if we were passing a `&mut Vec<T>`.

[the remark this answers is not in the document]
ritchie46: Yep you are right. Nvm👍

ritchie46 (approved):

PR #1096, opened 2022-06-23
Added notes about creating parquet files and submodules in the development documentation

joshuataylor:
This is copied out of the CI setup, I was confused why I was getting no files found for parquet.

I also added notes about [recurse-submodules](https://git-scm.com/docs/gitsubmodules) with Git.


jorgecarleitao:
Thanks, @joshuataylor !

PR #1097, opened 2022-06-23
Added support for projections in reading IPC streams

### LINKED ISSUE #1095: Support projection to IPC streams

joshuataylor:
Adds projections to IPC Streams, which is required for https://github.com/pola-rs/polars/issues/3778 which in turn is required for https://github.com/elixir-nx/explorer/issues/265 :rocket: 

As this is my first PR, and I'm new to Rust, please give any feedback you find helpful :raised_hands: 

Closes #1095

`StreamReader::new` has a new argument, `projection`. Add `None` to it to retain the original behavior.

### DIFF
--- a/src/io/ipc/read/stream.rs
+++ b/src/io/ipc/read/stream.rs
@@ -172,7 +172,7 @@ fn read_next<R: Read>(
             if let Some((_, map, _)) = projection {
                 // re-order according to projection
                 chunk
-                    .map(|chunk| apply_projection(chunk, &map))
+                    .map(|chunk| apply_projection(chunk, map))
                     .map(|x| Some(StreamState::Some(x)))
             } else {
                 chunk.map(|x| Some(StreamState::Some(x)))
--- a/src/io/ipc/read/stream.rs
+++ b/src/io/ipc/read/stream.rs
@@ -1,3 +1,4 @@
+use std::collections::HashMap;
 use std::io::Read;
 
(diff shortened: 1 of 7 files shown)


jorgecarleitao (approved):
Flawless. This is ready to merge from the `src` and `tests`. Amazing work, @joshuataylor !

There is a small fix we need to apply - in `integration-testing/src/bin/arrow-stream-to-file.rs:27:35`

we need to add `None` to the call, since the API changed.

src/io/ipc/read/stream.rs
```diff
@@ -155,18 +157,26 @@ fn read_next<R: Read>(
...
                 0,
                 file_size,
-            )
-            .map(|x| Some(StreamState::Some(x)))
+            );
+
+            if let Some((_, map, _)) = projection {
+                // re-order according to projection
+                chunk
+                    .map(|chunk| apply_projection(chunk, &map))
```
jorgecarleitao, line 175:
```suggestion
                    .map(|chunk| apply_projection(chunk, map))
```

joshuataylor:
Ah, missed those, added it. Will remember to keep an eye on CI.

jorgecarleitao:
Thanks! The CI for integration has been failing lately between C# and X. We use the same steps as apache/arrow and that is an issue from there. So, as long as there is not there all is good (I do not expect any impact - if it compiles we are good).

PR #1099, opened 2022-06-24
Simplified `Bytes` (internal)

jorgecarleitao:
This PR decouples `Bytes` from everything except `std`.

This ports @jhorstmann's [PR in arrow-rs](https://github.com/apache/arrow-rs/commit/688dd4c69b8f38d5d14a7bd71a8345fdcc6654e7). Although not as relevant here since we use `Vec` anyways, it enable us to potentially move `Bytes` to its own crate.

I am not fan of using `Arc` instead of `Box` - the struct itself does not need an `Arc`, will try to address this in a future PR.

Relevant thread in rust-users: https://users.rust-lang.org/t/implementing-a-potentially-foreign-allocated-potentially-mutable-memory-region/77388


jorgecarleitao:
@jhorstmann I had to require `Send + Sync` to the owner, or `Bytes` would not be `Send + Sync`. `unsafe impl Send for Bytes` in arrow-rs may be unsound.

ritchie46:
So we remove 1 layer of indirection to the data? And only decide at `Drop` what to do with the Vec we created? That saves a lot of branches!

jorgecarleitao:
> So we remove 1 layer of indirection to the data? And only decide at `Drop` what to do with the Vec we created? That saves a lot of branches!

I believe so - although it is unclear whether the compiler was not optimizing it already.

> P.S. which Arc are you referring to?

The one at `Foreign(Arc<dyn RefUnwindSafe + Send + Sync>)` - this arc is only here due to some constraint in the FFI that imo can go away.

ritchie46:
> I believe so - although it is unclear whether the compiler was not optimizing it already

I don't think it can. It could be either branch, so it has to check. I think.

> The one at Foreign(Arc<dyn RefUnwindSafe + Send + Sync>) - this arc is only here due to some constraint in the FFI that imo can go away.

Right, and if we remove it, the clone call goes over FFI go that smart atomic pointer?

src/buffer/bytes.rs
```diff
@@ -1,163 +1,118 @@
...
         len: usize,
-        deallocation: Deallocation,
+        owner: Arc<dyn RefUnwindSafe + Send + Sync>,
     ) -> Self {
-        assert!(matches!(deallocation, Deallocation::Foreign(_)));
         // This line is technically outside the assumptions of `Vec::from_raw_parts`, since
         // `ptr` was not allocated by `Vec`. However, one of the invariants of this struct
         // is that we do not expose this region as a `Vec`; we only use `Vec` on it to provide
         // immutable access to the region (via `Vec::deref` to `&[T]`).
+        // MIRI does not complain, which seems to agree with the line of thought.
```
ritchie46, line 44: Maybe we could also use `Box<[T]>` and call `Vec::from_raw_parts` only on `get_vec` on native allocated data.

jorgecarleitao: I tried this but in `get_vec` we would be creating a temporary Vec and returning `&mut Vec`, which does not compile :(

ritchie46: Oh yeah.. :(

jorgecarleitao:
> Right, and if we remove it, the clone call goes over FFI go that smart atomic pointer?

We do not know about smart pointers in FFI - in this case the idea is that we can `Arc` the struct containing the ref counter of the FFI struct (i.e. instead of `Arc<Arc<`, we use `Box<Arc<`, just like we did for `Array`.

PR #1105, opened 2022-06-25
Reduced reallocations when reading from IPC (`~12%`)

ritchie46:
We used `data.clear()` which sets the length to `0` and then use `data.resize(len, 0)`, so that every bytes was overwritten with zero before reading.

However, we can simply keep the `vec` at the maximum length and write to a slice, growing if needed. The bytes that are already initialized don't have to be reset because we will overwrite them in the subsequent `read` operation.

I want to add the same for parquet in a following PR.

### DIFF
--- a/examples/ipc_file_read.rs
+++ b/examples/ipc_file_read.rs
@@ -33,7 +33,7 @@ fn read_batch(path: &str) -> Result<(Schema, Chunk<Box<dyn Array>>)> {
     let schema = metadata.schema.clone();
 
     // advanced way: read the dictionary
-    let dictionaries = read::read_file_dictionaries(&mut file, &metadata, Default::default())?;
+    let dictionaries = read::read_file_dictionaries(&mut file, &metadata, &mut Default::default())?;
 
     let chunk_index = 0;
 
@@ -43,6 +43,7 @@ fn read_batch(path: &str) -> Result<(Schema, Chunk<Box<dyn Array>>)> {
         &metadata,
         None,
         chunk_index,
+        &mut Default::default(),
         &mut Default::default(),
     )?;
 
--- a/examples/ipc_file_read.rs
+++ b/examples/ipc_file_read.rs
@@ -33,7 +33,7 @@ fn read_batch(path: &str) -> Result<(Schema, Chunk<Box<dyn Array>>)> {
     let schema = metadata.schema.clone();
 
     // advanced way: read the dictionary
-    let dictionaries = read::read_file_dictionaries(&mut file, &metadata)?;
+    let dictionaries = read::read_file_dictionaries(&mut file, &metadata, Default::default())?;
 
     let chunk_index = 0;
 
--- a/examples/ipc_file_read.rs
+++ b/examples/ipc_file_read.rs
@@ -43,7 +43,7 @@ fn read_batch(path: &str) -> Result<(Schema, Chunk<Box<dyn Array>>)> {
         &metadata,
         None,
         chunk_index,
-        &mut vec![],
+        &mut Default::default(),
     )?;
 
     Ok((schema, chunk))
--- a/src/io/ipc/read/common.rs
+++ b/src/io/ipc/read/common.rs
@@ -85,6 +85,7 @@ pub fn read_record_batch<R: Read + Seek>(
     reader: &mut R,
     block_offset: u64,
     file_size: u64,
+    scratch: &mut ReadBuffer,
 ) -> Result<Chunk<Box<dyn Array>>> {
     assert_eq!(fields.len(), ipc_schema.fields.len());
     let buffers = batch
@@ -136,6 +137,7 @@ pub fn read_record_batch<R: Read + Seek>(
                         Error::from(OutOfSpecKind::InvalidFlatbufferCompression(err))
                     })?,
                     version,
+                    scratch,
                 )?)),
                 ProjectionResult::NotSelected((field, _)) => {
                     skip(&mut field_nodes, &field.data_type, &mut buffers)?;
@@ -162,6 +164,7 @@ pub fn read_record_batch<R: Read + Seek>(
                         Error::from(OutOfSpecKind::InvalidFlatbufferCompression(err))
                     })?,
                     version,
+                    scratch,
                 )
             })
             .collect::<Result<Vec<_>>>()?
@@ -221,6 +224,7 @@ fn first_dict_field<'a>(
 
 /// Read the dictionary from the buffer and provided metadata,
 /// updating the `dictionaries` with the resulting dictionary
+#[allow(clippy::too_many_arguments)]
 pub fn read_dictionary<R: Read + Seek>(
     batch: arrow_format::ipc::DictionaryBatchRef,
     fields: &[Field],
@@ -229,6 +233,7 @@ pub fn read_dictionary<R: Read + Seek>(
     reader: &mut R,
     block_offset: u64,
     file_size: u64,
+    scratch: &mut ReadBuffer,
 ) -> Result<()> {
     if batch
         .is_delta()
@@ -270,6 +275,7 @@ pub fn read_dictionary<R: Read + Seek>(
                 reader,
                 block_offset,
                 file_size,
+                scratch,
             )?;
             let mut arrays = columns.into_arrays();
             arrays.pop().unwrap()
--- a/src/io/ipc/read/common.rs
+++ b/src/io/ipc/read/common.rs
@@ -9,6 +9,7 @@ use crate::datatypes::{DataType, Field};
 use crate::error::{Error, Result};
 use crate::io::ipc::read::OutOfSpecKind;
 use crate::io::ipc::{IpcField, IpcSchema};
+use crate::io::ReadBuffer;
 
 use super::deserialize::{read, skip};
 use super::Dictionaries;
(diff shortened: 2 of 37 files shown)


ritchie46:
The failed test doesn't seem to be related.

jorgecarleitao:
Thanks, this is awesome! 

On my github it shows a diff that seems related to a missing rebase?

ritchie46:
Will have to change a few functions to that we make can know the tot read length. Before this was the `Vec`'s length. With this change, the length of the `Vec` can be longer than the read length. I will ping when ready.

src/io/ipc/read/reader.rs
```diff
@@ -36,6 +36,16 @@ pub struct FileMetadata {
     pub(crate) size: u64,
 }
 
+/// prepare `scratch` to read the message
+pub(super) fn prepare_scratch(scratch: &mut Vec<u8>, message_length: usize) -> &mut [u8] {
+    // ensure that we have enough scratch space
+    if message_length > scratch.len() {
+        scratch.resize(message_length, 0);
+    }
```
jorgecarleitao, line 44: I wonder if

```suggestion
    if message_length > scratch.len() {
        *scratch = vec![0; message_length];
    }
```

wouldn't be even faster. i.e. instead of having to re-alloc old data that will be overwritten anyways, we call `alloc_zeroed` (the min requirements for safety in `Read`).

We do something similar in parquet2: https://github.com/jorgecarleitao/parquet2/blob/main/src/read/page/reader.rs#L199

ritchie46: I think you are right. I suggest to first check the capacity. If the new size fits in capacity we only append the `new_size - size` zeros.

And otherwise go for the `alloc_zeroed` strategy.

jorgecarleitao: Or maybe one step further may be to:

```rust
if length > scratch.capacity() {
    *scratch = vec![0; message];
} else length > scratch.len() {
    scratch.resize(length, 0)
} else {
    scratch.truncate(length)
}
scratch
```

so, if we need to alloc, we just alloc-zero. If we only need a bit more zeroed space, we zeroed it.

ritchie46: > Or maybe one step further may be to:
> 
> ```rust
> if length > scratch.capacity() {
>     *scratch = vec![0; message];
> } else if length > scratch.len() {
>     scratch.resize(length, 0);
> } else {
>     scratch.truncate(length);
> }
> scratch
> ```
> 
> so, if we need to alloc, we just alloc-zero. If we only need a bit more zeroed space, we zeroed it.

Haha.. my thougts in code! ;)

jorgecarleitao: we would do a bunch of resize and truncates, since every message will be slightly different length. In this context we could skip the `truncate` and only expose `&mut scratch[..length]` as this PR already does.

We just need to be mindful that data on the scratch can be read between messages if the message is not valid (a bit like out of bound reads but for owned). imo this is not an issue because the user already offered `scratch` to the reader and should thus assume that `Read` may read any of that content.

src/io/ipc/read/reader.rs
```diff
@@ -116,9 +116,9 @@ fn read_dictionary_block<R: Read + Seek>(
 pub fn read_file_dictionaries<R: Read + Seek>(
     reader: &mut R,
     metadata: &FileMetadata,
+    scratch: &mut ReadBuffer,
```
ritchie46, line 119: This extra argument allows us to reuse the `ReadBuffer` that is owned by the `IPCFileReader`.

examples/ipc_file_read.rs
```diff
@@ -43,7 +43,7 @@ fn read_batch(path: &str) -> Result<(Schema, Chunk<Box<dyn Array>>)> {
         &metadata,
         None,
         chunk_index,
-        &mut vec![],
+        &mut Default::default(),
```
jorgecarleitao, line 46: would it be make sense / be possible to make the struct private? I think we could it via

```
struct ReadBuffer<'a> {
    data: &'a mut Vec<u8>,
    length: usize,
}
```

users would still use Vec in the public, and we would wrap the mut reference.

We do lose the `length` between calls, but I think that it would be less mental workload knowing about `ReadBuffer`.

src/io/ipc/read/read_basic.rs
```diff
@@ -105,9 +107,8 @@ fn read_compressed_buffer<T: NativeType, R: Read + Seek>(
     let mut buffer = vec![T::default(); length];
 
     // decompress first
-    // todo: move this allocation to an external buffer for re-use
-    let mut slice = vec![0u8; buffer_length];
-    reader.read_exact(&mut slice)?;
+    scratch.set_len(buffer_length);
```
ritchie46, line 110: Had to add scratch to most functions to reach this point. Here we can reuse the scratch for every decompression.

examples/ipc_file_read.rs
ritchie46: Could we make a public API that allows `Vec` as input?

We already implement `Default` and `From<Vec>`. 

So we could make an API that accepts `Vec` or `ReadBuffer` but is easily created with `Default::default()`.


For more lower level users (e.g. polars, I still want to parallize the reading) it would be easier if it is public. But we could still hide that from the default readers API ofcourse.

src/io/ipc/read/common.rs
```diff
@@ -286,6 +292,61 @@ pub fn read_dictionary<R: Read + Seek>(
     Ok(())
 }
 
+/// A small wrapper around `[Vec<u8>]` that allows us to reuse memory once it is initialized.
```
jorgecarleitao, line 295:
```suggestion
/// A small wrapper around [`Vec<u8>`] that allows us to reuse memory once it is initialized.
```

src/io/ipc/read/common.rs
```diff
@@ -286,6 +292,61 @@ pub fn read_dictionary<R: Read + Seek>(
...
+/// This may improve performance of the `[Read]` trait.
```
jorgecarleitao, line 296:
```suggestion
/// This may improve performance of the [`Read`] trait.
```

src/io/ipc/read/common.rs
```diff
@@ -286,6 +292,61 @@ pub fn read_dictionary<R: Read + Seek>(
...
+    pub fn new(length: usize) -> Self {
+        let data = vec![0; length];
+        Self { data, length }
+    }
+
+    /// Set the length of the [`ReadBuf`]. Contrary to the
+    /// method on `Vec` this is `safe` because this function guarantees that
+    /// the underlying data always is initialized.
+    pub fn set_len(&mut self, length: usize) {
+        dbg!(length, self.data.len(), self.data.capacity());
```
jorgecarleitao, line 315:
```suggestion
```

src/io/ipc/read/common.rs
```diff
@@ -286,6 +292,61 @@ pub fn read_dictionary<R: Read + Seek>(
...
+    length: usize,
+}
+
+impl ReadBuffer {
+    /// Create a new [`ReadBuf`] initialized to `length`
+    pub fn new(length: usize) -> Self {
+        let data = vec![0; length];
+        Self { data, length }
+    }
+
```
jorgecarleitao, line 310:
```suggestion
```

imo there is no need for this method - `From<Vec` covers it 👍

src/io/ipc/read/deserialize.rs
```diff
@@ -7,6 +7,7 @@ use arrow_format::ipc::MetadataVersion;
 use crate::array::*;
 use crate::datatypes::{DataType, Field, PhysicalType};
 use crate::error::Result;
+use crate::io::ipc::read::common::ReadBuffer;
```
jorgecarleitao, line 10: I have been trying to make `common` private - could we pub export this at `io::ipc::read` so users can just import from it?

src/io/ipc/read/file_async.rs
```diff
@@ -221,8 +221,8 @@ where
     R: AsyncRead + AsyncSeek + Unpin,
 {
     let mut dictionaries = Default::default();
-    let mut data = vec![];
-    let mut buffer = vec![];
+    let mut data = ReadBuffer::new(0);
+    let mut buffer = ReadBuffer::new(0);
```
jorgecarleitao, line 231:
```suggestion
    let mut data: ReadBuffer = vec![].into();
    let mut buffer: ReadBuffer = vec![].into();
```

jorgecarleitao (approved):
I love it.

Left some minor comments, but otherwise ready to go.

Any benches worth mentioning?

src/io/ipc/read/deserialize.rs
ritchie46: Will do. :+1:

src/io/ipc/read/common.rs
ritchie46: Ok, will remove that.

ritchie46:
> I love it.
> 
> Left some minor comments, but otherwise ready to go.
> 
> Any benches worth mentioning?

Yes!


# main

```
 Performance counter stats for './target/release/memcheck':

          3.009,99 msec task-clock                #    1,000 CPUs utilized          
                15      context-switches          #    4,983 /sec                   
                 0      cpu-migrations            #    0,000 /sec                   
           957.156      page-faults               #  317,993 K/sec                  
    10.602.766.501      cycles                    #    3,523 GHz                    
    16.565.436.781      instructions              #    1,56  insn per cycle         
     2.804.597.000      branches                  #  931,764 M/sec                  
        38.538.003      branch-misses             #    1,37% of all branches        

       3,010826723 seconds time elapsed

       1,885070000 seconds user
       1,124638000 seconds sys

```

linear (or no?) growing strategy
```
 
 Performance counter stats for './target/release/memcheck':

          2.824,92 msec task-clock                #    1,000 CPUs utilized          
                 8      context-switches          #    2,832 /sec                   
                 0      cpu-migrations            #    0,000 /sec                   
           820.540      page-faults               #  290,465 K/sec                  
    10.279.655.892      cycles                    #    3,639 GHz                    
    16.038.558.971      instructions              #    1,56  insn per cycle         
     2.714.267.088      branches                  #  960,830 M/sec                  
        38.377.349      branch-misses             #    1,41% of all branches        

       2,825607071 seconds time elapsed

       1,824543000 seconds user
       1,000298000 seconds sys
```

# Exponential growing strategy (same as vec)
```
  Performance counter stats for './target/release/memcheck':

          2.678,32 msec task-clock                #    0,996 CPUs utilized          
                51      context-switches          #   19,042 /sec                   
                 3      cpu-migrations            #    1,120 /sec                   
           711.114      page-faults               #  265,507 K/sec                  
     9.706.936.972      cycles                    #    3,624 GHz                    
    15.635.366.640      instructions              #    1,61  insn per cycle         
     2.645.151.012      branches                  #  987,616 M/sec                  
        38.408.838      branch-misses             #    1,45% of all branches        

       2,687799246 seconds time elapsed

       1,835066000 seconds user
       0,843570000 seconds sys
```

ritchie46:
Good to go on my side. :+1:

joshuataylor:
This is amazing, when using latest arrow2 branch against Polars, with lazy IPC Stream against 241 Streaming Arrow files totalling 1gb dataset dropped over 10% to scan! :tada:

PR #1141, opened 2022-07-07
Moved `Bytes` to own crate

jorgecarleitao:
This PR proposes that we move the `Bytes` API to another crate so others can use it without having to copy paste. It also separates concerns between this crate (arrow) and a mechanism to use share foreign arrays with Rust still compatible with mutable APIs.

Proposed home: https://github.com/DataEngineeringLabs/foreign_vec


jorgecarleitao:
cc @jhorstmann @ritchie46 , since both also contributed to this component (either here or in arrow-rs).

PR #1143, opened 2022-07-07
fix casting dictionary keys

ritchie46:
Updating arrow2 had some failed test in polars. It turned out that casting dictionary keys was a no-op. This PR fixes that and adds a test.

### DIFF
--- a/src/compute/cast/dictionary_to.rs
+++ b/src/compute/cast/dictionary_to.rs
@@ -143,11 +143,9 @@ pub(super) fn dictionary_cast_dyn<K: DictionaryKey + num_traits::NumCast>(
 
             // Safety:
             // we return an error on overflow so the integers remain within bounds
-            unsafe {
-                match_integer_type!(to_keys_type, |$T| {
-                    key_cast!(keys, values, array, &to_key_type, $T, to_type.clone())
-                })
-            }
+            match_integer_type!(to_keys_type, |$T| {
+                key_cast!(keys, values, array, &to_key_type, $T, to_type.clone())
+            })
         }
         _ => unpack_dictionary::<K>(keys, values.as_ref(), to_type, options),
     }
--- a/src/compute/cast/dictionary_to.rs
+++ b/src/compute/cast/dictionary_to.rs
@@ -92,8 +92,7 @@ where
             Box::new(values.data_type().clone()),
             is_ordered,
         );
-        // Safety
-        // we errored if any cast overflowed
+        // Safety: this is safe because given a type `T` that fits in a `usize`, casting it to type `P` either overflows or also fits in a `usize`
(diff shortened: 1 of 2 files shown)


src/compute/cast/dictionary_to.rs
```diff
@@ -88,8 +89,9 @@ where
             Box::new(values.data_type().clone()),
             is_ordered,
         );
-        // some of the values may not fit in `usize` and thus this needs to be checked
-        DictionaryArray::try_new(data_type, casted_keys, values.clone())
+        // Safety
+        // we errored if any cast overflowed
+        unsafe { DictionaryArray::try_new_unchecked(data_type, casted_keys, values.clone()) }
```
ritchie46, line 94: The branch above checks the null count. This ensures us that we don't have any overflowing integers, so the invariants of the struct are held and we don't need to check it.

ritchie46: P.S. This function is not used. Do we want it for the public API or can we remove it?

src/compute/cast/dictionary_to.rs
```diff
@@ -7,15 +7,16 @@ use crate::{
...
+    ($keys:expr, $values:expr, $array:expr, $to_keys_type:expr, $to_type:ty, $to_datatype:expr) => {{
         let cast_keys = primitive_to_primitive::<_, $to_type>($keys, $to_keys_type);
 
         // Failure to cast keys (because they don't fit in the
         // target type) results in NULL values;
         if cast_keys.null_count() > $keys.null_count() {
             return Err(Error::Overflow);
         }
-        DictionaryArray::try_new($array.data_type().clone(), $keys.clone(), $values.clone())
+        DictionaryArray::try_new_unchecked($to_datatype, cast_keys, $values.clone())
```
jorgecarleitao, line 18:
```suggestion
        // Safety: this is safe because given a type `T` that fits in a `usize`, casting it to type `P` either overflows or also fits in a `usize`
        unsafe {
             DictionaryArray::try_new_unchecked($to_datatype, cast_keys, $values.clone())
        }
```

src/compute/cast/dictionary_to.rs
```diff
@@ -88,8 +89,9 @@ where
             Box::new(values.data_type().clone()),
             is_ordered,
         );
-        // some of the values may not fit in `usize` and thus this needs to be checked
-        DictionaryArray::try_new(data_type, casted_keys, values.clone())
+        // Safety
+        // we errored if any cast overflowed
```
jorgecarleitao, line 93:
```suggestion
        // Safety: this is safe because given a type `T` that fits in a `usize`, casting it to type `P` either overflows or also fits in a `usize`
```

jorgecarleitao (commented):
Looks awesome, thanks a lot, @ritchie46 !

Left minor suggestions to guide readability of `unsafe`

PR #1148, opened 2022-07-08
Use ahash by default

ritchie46:
[ahash](https://github.com/tkaitchuck/aHash) is a dos resistant hashing algorithm that is much faster than std's default. Given that there are users with millions of columns, not paying unneeded cryptographic hashing costs seems worth it to me.

Aside from the crate being pulled in, this should not have much effect on build times as the hasmaps/sets get compiled on instantiation.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,7 +64,7 @@ base64 = { version = "0.13.0", optional = true }
 futures = { version = "0.3", optional = true }
 
 # for faster hashing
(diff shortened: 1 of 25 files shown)


jorgecarleitao:
AFAIK `std` is already using `ahash`, so there isn't much to gain performance-wise. I think the benefit of `ahash` is a public API related to `entry`?

jorgecarleitao (approved):

PR #1149, opened 2022-07-08
fix ipc projection

ritchie46:
Depending of the ordering of the hashmap, the result of projecting columns could be different. This PR fixes that.

Not something I want to include in this PR, but I am interested why we want to sort the projections? I saw that we had an iterator that had that assumption, but it is my understanding that we also could keep a count on that iterator and stop when we reached that count. If so, I am happy to make a follow up PR that leaves projection as is.


ritchie46:
I could not create the ipc files so I did not create a test. The correctness is tested against polars tests, but I am happy to add one here as well.

PR #1151, opened 2022-07-11
Added example reading Avro produced by Kafka

jorgecarleitao:
Kafka uses the single object representation of Avro. This PR adds an example showing how to read such stream.

This emerged from a question by [scimas](https://users.rust-lang.org/u/scimas) at rust users: https://users.rust-lang.org/t/status-of-avro-in-rust/78086/6

### DIFF
--- a/examples/avro_kafka.rs
+++ b/examples/avro_kafka.rs
@@ -1,0 +1,54 @@
+use arrow2::{
+    datatypes::{DataType, Field},
+    error::Error,
+    io::avro,
+};
+
+fn read_schema_id<R: std::io::Read>(reader: &mut R) -> Result<i32, Error> {
+    let mut header = [0; 5];
+    reader.read_exact(&mut header)?;
+
+    if header[0] != 0 {
+        return Err(Error::ExternalFormat(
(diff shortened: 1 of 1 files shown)


examples/avro_kafka.rs
```diff
@@ -0,0 +1,54 @@
...
+fn read_schema_id<R: std::io::Read>(reader: &mut R) -> Result<i32, Error> {
+    let mut header = [0; 5];
+    reader.read_exact(&mut header)?;
+
+    if header[0] != 0 {
+        return Err(Error::ExternalFormat(
+            "Avro requires the first byte to be a zero".to_string(),
+        ));
+    }
+    Ok(i32::from_be_bytes(header[1..].try_into().unwrap()))
```
scimas, line 16: I see that this is already merged, but I would suggest changing this to u32 instead of i32.

I don't know where the specification for the schema_id is, but the API documentation for Confluent's schema registry [specifies](https://docs.confluent.io/platform/current/schema-registry/develop/api.html#get--subjects-(string-%20subject)-versions-(versionId-%20version)) that version (schema_id) should be in the range [1, 2^31 - 1].

PR #1152, opened 2022-07-11
chore(buffer): make constraint in buffer shorter

sundy-li:


jorgecarleitao:
This is a nice idea! I think we can make this even stronger - 

```
pub struct Buffer<T> {...}

impl<T: std::fmt::Debug> std::fmt::Debug for Buffer<T> {...}
```

i.e. no constraint.

Side: could you describe what is the goal? I am genuinely curious

sundy-li:
> Side: could you describe what is the goal? I am genuinely curious

I want to use `Buffer<u256>` and `Buffer<u512>` in databend.

PR #1153, opened 2022-07-11
Removed un-needed indirection

jorgecarleitao:
Essentially replaced `Box<InternalArrowArray>` by `InternalArrowArray`, since the box was not needed.


ritchie46:
Nice catch ❤️

PR #1157, opened 2022-07-12
Removed unneeded `Default` constraint

hohav:
I was trying to use `MutableListArray` with a type that doesn't implement `Default`, and ran into an error caused by this constraint. It looks unnecessary, and removing it doesn't seem to break anything.


jorgecarleitao:
Thanks a lot, @hohav !

PR #1161, opened 2022-07-13
Added constructor `new_from`

hohav:
Implements a more general `MutableFixedSizeListArray` constructor, `new_from`, to allow creating one with a non-nullable datatype (since `default_datatype` creates one that's nullable).

### DIFF
--- a/src/array/fixed_size_list/mutable.rs
+++ b/src/array/fixed_size_list/mutable.rs
@@ -48,7 +48,7 @@ impl<M: MutableArray> MutableFixedSizeListArray<M> {
     pub fn new_from(values: M, data_type: DataType, size: usize) -> Self {
         assert_eq!(values.len(), 0);
         match data_type {
-            DataType::FixedSizeList(..) => {},
+            DataType::FixedSizeList(..) => (),
             _ => panic!("data type must be FixedSizeList (got {:?})", data_type),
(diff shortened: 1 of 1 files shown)


src/array/fixed_size_list/mutable.rs
```diff
@@ -47,6 +41,11 @@ impl<M: MutableArray> MutableFixedSizeListArray<M> {
             Box::new(Field::new(name, values.data_type().clone(), nullable)),
             size,
         );
+        Self::new_from(values, data_type, size)
+    }
+
+    /// Creates a new [`MutableFixedSizeListArray`] from a [`MutableArray`], [`DataType`] and size.
+    pub fn new_from(values: M, data_type: DataType, size: usize) -> Self {
```
jorgecarleitao, line 48: I _think_ we need to check that this `data_type` is compatible with `MutableFixedSizeListArray` and `values`, or we end up with an incompatible data_type, right? Something like we do for `ListArray` [here](https://github.com/jorgecarleitao/arrow2/blob/26120da6811eb0b279ebba1dfea2992e66cb1cb2/src/array/list/mod.rs#L59)

jorgecarleitao (commented):
Thanks a lot!

I left a small comment.

PR #1163, opened 2022-07-13
release memory on predicate with 0% selectivity

ritchie46:
When we selected no data from an array, we sliced the buffer keeping all the data in memory. This ensures we are able to free the array data by creating a new empty array.


jorgecarleitao:
Oooof, this was a nasty one!

PR #1165, opened 2022-07-15
support convert i96 values into other time unit of arrow's timestamp

sundy-li:
Related to  https://github.com/datafuselabs/databend/issues/6627

### DIFF
--- a/src/io/parquet/read/deserialize/simple.rs
+++ b/src/io/parquet/read/deserialize/simple.rs
@@ -279,6 +279,37 @@ pub fn page_iter_to_arrays<'a, I: 'a + DataPages>(
     })
 }
 
+/// Unify the timestamp unit from parquet TimeUnit into arrow's TimeUnit
+/// Returns (a int64 factor, is_multiplier)
+fn unifiy_timestmap_unit(
+    logical_type: &Option<PrimitiveLogicalType>,
+    time_unit: TimeUnit,
+) -> (i64, bool) {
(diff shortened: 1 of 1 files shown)


jorgecarleitao:
Ooof, what a simplification!

jorgecarleitao:
Thanks a lot, @sundy-li !

src/io/parquet/read/deserialize/simple.rs
```diff
@@ -347,130 +353,38 @@ fn timestamp_dict<'a, K: DictionaryKey, I: 'a + DataPages>(
...
-                int96_to_i64_ns,
-            )));
-        } else {
-            return Err(Error::nyi("Can't decode int96 to timestamp other than ns"));
-        }
+                move |x| int96_to_i64_ns(x) * a,
+            ))),
+            (a, false) => Ok(dyn_iter(primitive::DictIter::<K, _, _, _, _>::new(
+                pages,
+                DataType::Timestamp(TimeUnit::Nanosecond, None),
```
sundy-li, line 366: @jorgecarleitao  Should we use `data_type` here? It may be incorrect to enforce the datatype to be `DataType::Timestamp(TimeUnit::Nanosecond, None)`.

jorgecarleitao: I agree.

I think that this comes from https://github.com/jorgecarleitao/arrow2/blob/main/src/io/parquet/read/schema/convert.rs#L181 where we infer int96 to be that, which is why here it works.

PR #1169, opened 2022-07-19
implement Default for Utf8Array

ritchie46:
Improves ergonomics when using `std::mem::take`.

### DIFF
--- a/src/array/utf8/mod.rs
+++ b/src/array/utf8/mod.rs
@@ -590,3 +590,14 @@ unsafe impl<O: Offset> GenericBinaryArray<O> for Utf8Array<O> {
         self.offsets()
     }
 }
+
+impl<O: Offset> Default for Utf8Array<O> {
+    fn default() -> Self {
+        let data_type = if O::IS_LARGE {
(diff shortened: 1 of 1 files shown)


src/array/utf8/mod.rs
```diff
@@ -590,3 +590,14 @@ unsafe impl<O: Offset> GenericBinaryArray<O> for Utf8Array<O> {
...
 }
+
+impl<O: Offset> Default for Utf8Array<O> {
+    fn default() -> Self {
+        let data_type = if O::IS_LARGE {
+            DataType::LargeUtf8
+        } else {
+            DataType::Utf8
+        };
+        Utf8Array::new(data_type, Default::default(), Default::default(), None)
```
jorgecarleitao, line 601:
```suggestion
        Utf8Array::new(data_type, vec![0].into(), Default::default(), None)
```

we require offsets to always contain the 0 (to avoid such check on every access to the offsets).

ritchie46: Good catch. I knew that.

PR #1176, opened 2022-07-22
use `FnOnce` for `PrimitiveArray::apply_validity`

ritchie46:
Allow passed closures to mutably borrow from their environment.


ritchie46:
Maybe we should have two versions so that callers can choose what they need. E.g. an `apply_validity` and an `apply_validity_mut`. :thinking:

PR #1190, opened 2022-07-28
Improved performance of reading Binary from parquet

ritchie46:
During a flamegraph of a parquet file I saw that `Binary<O>::push` spend a large amount reallocation. This PR  pre-allocates a the values buffer in parquet hinted by the capacity. This can prevent a few reallocations. The string/binary size is arbitrarily set to the size of a minimal string, e.g. 3 pointers. Could of course be something else that better fits real world data. I could do some runs to find a distribution?

### DIFF
--- a/src/io/parquet/read/deserialize/binary/utils.rs
+++ b/src/io/parquet/read/deserialize/binary/utils.rs
@@ -42,7 +42,7 @@ impl<O: Offset> Binary<O> {
         offsets.push(O::default());
         Self {
             offsets: Offsets(offsets),
-            values: vec![],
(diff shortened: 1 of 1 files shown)


jorgecarleitao:
Thanks, @ritchie46! Should we make it a power of 2, i.e. 16 or 32?

jorgecarleitao:
Thinking about it, `capacity` is not a multiple, so my comment is not very insightful.

ritchie46:
> Thinking about it, `capacity` is not a multiple, so my comment is not very insightful.

Yes, we could also round up to a power of two? Maybe the allocator does something like that already?

In any case, I have no strong opinion here.

PR #1197, opened 2022-07-31
Added support to mmap IPC format

jorgecarleitao:
This PR adds support for memory mapping (uncompressed) IPC files into arrays, thereby avoiding data reads.


src/ffi/mmap.rs
```diff
@@ -0,0 +1,154 @@
...
+        .iter()
+        .map(|maybe_buffer| match maybe_buffer {
+            Some(b) => *b as *const std::os::raw::c_void,
+            None => std::ptr::null(),
+        })
+        .collect::<Box<[_]>>();
+    let n_buffers = buffers.len() as i64;
+
+    let mut private_data = Box::new(PrivateData::<T> {
+        data: data.clone(),
```
ritchie46, line 106: What do we clone here? E.g.` what is `T`?

tests/it/io/ipc/read/mmap.rs
```diff
@@ -0,0 +1,40 @@
+use arrow2::error::Result;
+use arrow2::io::ipc::mmap::map_chunk;
+
+use super::super::common::read_gzip_json;
+
+#[derive(Clone)]
```
ritchie46, line 6: When we mmap we never get an owned `Vec`, only a `&[u8]` or a `&mut [u8]`. Do we create this `Vec` temporarily and simply never allow it to drop?

We might also store the `memmap2::Mmap` struct. This implements `AsRef<[u8]>` and will ensure the memory mapped region remains mapped until it is dropped.

src/ffi/mmap.rs
jorgecarleitao: `T` here is anything that implements `Clone + AsRef<[u8]>`, so, for example for `Mmap`, something like `Arc<Mmap<File>>`.

tests/it/io/ipc/read/mmap.rs
jorgecarleitao: yeah, but `&[u8]` is technically `&'a[u8]` and our arrays are owned (they do not carry any lifetime). So, we need to control lifetime somewhere else, which is why we need an owned version of the `AsRef<[u8]>`. Moving `memmap2::Mmap` to `Arc<_>` is the most natural choice.

src/ffi/mmap.rs
ritchie46: Right, I was confused as `Vec<u8>` also implements those. ^^

tests/it/io/ipc/read/mmap.rs
ritchie46: > Moving memmap2::Mmap to Arc<_> is the most natural choice.

Agreed.

PR #1204, opened 2022-08-03
Fix escaped like wildcards

### LINKED ISSUE #702: Fix LIKE escape issues
Due to using `regex` crate directly, it's sensitive to not escape `like` patterns`.
A solution to it can be using the `escape` function in `regex`: https://docs.rs/regex/1.5.4/regex/fn.escape.html

Next, there should be a possibility to escape the `%` and `_` characters.

Arrow-rs issues/PRs:

https://github.com/apache/arrow-rs/pull/1085
https://github.com/apache/arrow-rs/pull/1042
https://github.com/apache/arrow-rs/issues/415

daniel-martinez-maqueda-sap:
# Which issue does this PR close?

Closes #702 

# Rationale for this change
 
It is explained in the Issues linked above.

# What changes are included in this PR?

Added a new function that replaces the like wildcards '%' and '_' for the regex counterparts before executing them. It also takes into account that the wildcards can be escaped, in that case, it does remove the escape characters and leaves the wildcards so that they are matched against the raw character.

This is implemented iterating over all the characters of the pattern to figure out when it needs to be transformed or not.

# Are there any user-facing changes?
N/A

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ csv-core = { version = "0.1", optional = true }
 csv-async = { version = "^1.1", optional = true }
 
 regex = { version = "^1.3", optional = true }
+regex-syntax = { version = "^0.6", optional = true }
 streaming-iterator = { version = "0.1", optional = true }
 fallible-streaming-iterator = { version = "0.1", optional = true }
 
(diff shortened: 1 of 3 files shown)


jorgecarleitao:
Thanks a lot @daniel-martinez-maqueda-sap 🙇

jorgecarleitao:
Released as part of 0.13.1.

PR #1205, opened 2022-08-04
Improved API of `mmap`

ritchie46:
Two changes.

### 1. expose more information on the `FileMetaData`.
I wanted to know how many `Chunks` there are in the IPC file. This also exposes the `size` attribute as that can also be valuable information.

If they are an invariant for unsafe code, I can expose them via functions?

### 2. use Arc<T>

There was an assumption that `Arc<Mmap` implements `Clone + AsRef<[u8]`. It does not. Calling `as_ref` on an `Arc<T> ` gives us `T`, which was opaque to us.

Changed the function signature so that we can pass an `Arc<Mmap>`.


ritchie46:
I will update the tests tonight.

PR #1213, opened 2022-08-05
`FileSink` now closes the underlying writer.

samkaufman:
This PR modifies `FileSink` to, on close, close its inner `FileStreamer`'s owned writer. At the moment, the inner `AsyncWrite`'s `close` method is not called when the `FileSink` is closed.

This is a minor backwards-incompatible change. While it won't affect cases where the FileSink wraps an actual file, which is generally closed when dropped, it may affect behavior when wrapping an `AsyncWrite` which expects close to be called. (In my case, that was a handle to an S3 object.)

### DIFF
--- a/src/io/parquet/write/sink.rs
+++ b/src/io/parquet/write/sink.rs
@@ -1,7 +1,7 @@
 use ahash::AHashMap;
 use std::{pin::Pin, task::Poll};
 
-use futures::{future::BoxFuture, AsyncWrite, FutureExt, Sink, TryFutureExt};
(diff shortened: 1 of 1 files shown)


samkaufman:
The "Integration IPC / Flight / Test" failure seems to be a result of this recent commit: https://github.com/apache/arrow/commit/6d1bc624454b8a83e5c90410b51c2718059c394b.

jorgecarleitao:
Thanks a lot @samkaufman 🙇  Yes, ignore the integration test - I need to fix that as something upstream changed.

PR #1223, opened 2022-08-15
Mutable struct push unchecked

hohav:
Adds `push_unchecked` to `MutableStructArray`.

In my tests, this reduced total time to generate a representative StructArray by more than 1/3.

### DIFF
--- a/src/array/mod.rs
+++ b/src/array/mod.rs
@@ -9,7 +9,7 @@
 //! * [`Utf8Array`] and [`MutableUtf8Array`], an array of variable length utf8 values
 //! * [`BinaryArray`] and [`MutableBinaryArray`], an array of opaque variable length values
 //! * [`ListArray`] and [`MutableListArray`], an array of arrays (e.g. `[[1, 2], None, [], [None]]`)
-//! * [`StructArray`], an array of arrays identified by a string (e.g. `{"a": [1, 2], "b": [true, false]}`)
+//! * [`StructArray`] and [`MutableStructArray`], an array of arrays identified by a string (e.g. `{"a": [1, 2], "b": [true, false]}`)
 //! All immutable arrays implement the trait object [`Array`] and that can be downcasted
 //! to a concrete struct based on [`PhysicalType`](crate::datatypes::PhysicalType) available from [`Array::data_type`].
 //! All immutable arrays are backed by [`Buffer`](crate::buffer::Buffer) and thus cloning and slicing them is `O(1)`.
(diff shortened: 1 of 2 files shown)


jorgecarleitao:
What do you think about simply removing the check in `push`, or rename `push` to `try_push` and make the assert an error?

I am asking because `*_unchecked` is usually used in Rust to denote an operation that requires unsafe (e.g. [`get_unchecked`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.get_unchecked)).

hohav:
> What do you think about simply removing the check in `push`, or rename `push` to `try_push` and make the assert an error?
> 
> I am asking because `*_unchecked` is usually used in Rust to denote an operation that requires unsafe (e.g. [`get_unchecked`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.get_unchecked)).

Removing the check altogether would be fine for my purposes, but OTOH it's still an important invariant. My inclination is to make it "opt out", i.e. have the most "obvious" option (`push`) do the check, with some other less-obvious function to allow skipping it. I don't have a better name to suggest than `push_unchecked`; do you know of another naming scheme for this situation?

Re: `try_push`, I still prefer not return an Err for the reasons we discussed in #1196. That seems independent of when/where to do the validation check, which is more of a performance question.

jorgecarleitao:
You are right. What about `push_validity`, hinting that we are essentially pushing to the validity?

(note that the integration test is unrelated, it is something upstream)

hohav:
Even knowing how `MutableStructArray` works, I wouldn't be able to guess at the difference between `push` and `push_validity` from the names. What about `push_fast`? Or maybe `push_fast_n_furious`? :)

On the other hand I've come around somewhat to removing the check altogether. The invariant will be checked when converting to a `StructArray`, so we'd still fail reasonably early.

Yet another option would be `#[cfg(debug_assertions)]`, but I don't know if that approach is idiomatic. Anyway, I'm fine with whichever approach you prefer.

jorgecarleitao:
exactly, that was the reason why I was thinking about keeping `push` as not performing the check. 

Another option I can think about is:

```
fn push<F: Fn(&mut [Box<dyn Array>])>(&mut self, op: F) {
    op(&mut self.values);
    self.validity.push(true); // or something like this
    self.assert_lengths();
}
```

this allows the user to modify the values at will multiple times before the push

hohav:
I like removing the check more now, especially since it allows pushing to all the children at once before calling `push` N times which might be convenient.

I removed the check altogether and rebased. Should be good to go now, I hope.

jorgecarleitao:
Sorry for the back and forth, I just think it is easier for others to keep the API aligned with Rust (specially functions named `*_unchecked`, that triggers unsafety feelings.

Thanks again for the PR and for the discussion 🙇

PR #1234, opened 2022-08-20
Fixed incorrect command in doc for generating ORC files

poga:
While following `DEVELOPMENT.md` to setup a working environment for `arrow2`, I noticed that `cargo test --features full` failed because missing ORC files.

This PR replaced the incorrect command for "Generate ORC files" with the correct one.


jorgecarleitao:
Indeed, thanks a lot!

Issue #1177, opened 2022-07-22
Dataframe written to disk failed to read


mutecamel:
Hi there. I have crafted a [polars](https://github.com/pola-rs/polars) dataframe, and saved it to disk in parquet/feather format. But this file cannot be read back into memory using `arrow2`. It can be successfully read by `pyarrow` though.

```python
In [1]: import pandas as pd

In [2]: import polars as pl

In [3]: df = pd.read_parquet("err.parquet")

In [4]: df = pl.read_parquet("err.parquet")
---------------------------------------------------------------------------
ArrowErrorException                       Traceback (most recent call last)
Input In [4], in <cell line: 1>()
----> 1 df = pl.read_parquet("err.parquet")

File ~/.pyenv/versions/3.10.4/envs/.../lib/python3.10/site-packages/polars/io.py:862, in read_parquet(source, columns, n_rows, use_pyarrow, memory_map, storage_options, parallel, row_count_name, row_count_offset, low_memory, **kwargs)
    849         raise ImportError(
    850             "'pyarrow' is required when using 'read_parquet(..., use_pyarrow=True)'."
    851         )
    853     return from_arrow(  # type: ignore[return-value]
    854         pa.parquet.read_table(
    855             source_prep,
   (...)
    859         )
    860     )
--> 862 return DataFrame._read_parquet(
    863     source_prep,
    864     columns=columns,
    865     n_rows=n_rows,
    866     parallel=parallel,
    867     row_count_name=row_count_name,
    868     row_count_offset=row_count_offset,
    869     low_memory=low_memory,
    870 )

File ~/.pyenv/versions/3.10.4/envs/.../lib/python3.10/site-packages/polars/internals/frame.py:684, in DataFrame._read_parquet(cls, file, columns, n_rows, parallel, row_count_name, row_count_offset, low_memory)
    682 projection, columns = handle_projection_columns(columns)
    683 self = cls.__new__(cls)
--> 684 self._df = PyDataFrame.read_parquet(
    685     file,
    686     columns,
    687     projection,
    688     n_rows,
    689     parallel,
    690     _prepare_row_count_args(row_count_name, row_count_offset),
    691     low_memory=low_memory,
    692 )
    693 return self

ArrowErrorException: OutOfSpec("The dictionary key must be smaller")

In [5]: df = pl.read_parquet("err.parquet", use_pyarrow=True)  # no exception
```

[...]

PR #1235, opened 2022-08-21
Fix check_indexes for empty values in dictory

DCjanus:
fix: https://github.com/jorgecarleitao/arrow2/issues/1177


DCjanus:
I'm not familiar with this repo, please let me know if there is anything need to be revised.

And feel free to push commits to this branch if you don't want to type words.

jorgecarleitao:
Hi, thanks for the PR; very idiomatic :)

I think that we need to be careful here: the change of the invariant results in some functions now requiring `unsafe` (or adding extra safeguards). For example, a user could use `set_validity` and `with_validity` to "unmask" some indexes, which would result in out of bound accesses in safe Rust.

In other words, because the invariant is now on the combination (indices, validity), changes to either must now be checked (or marked `unsafe` accordingly).

PR #1238, opened 2022-08-28
Added documentation on parquet compatibility under `TimeUnit`.

TurnOfACard:
`TimeUnit::Second` will not store a logical type when converted to parquet format.

### DIFF
--- a/src/io/parquet/write/mod.rs
+++ b/src/io/parquet/write/mod.rs
@@ -1,4 +1,17 @@
 //! APIs to write to Parquet format.
+//!
+//! # Arrow/Parquet Interoperability
+//! As of [parquet-format v2.9](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md)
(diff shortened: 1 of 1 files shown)


jorgecarleitao:
Thanks! Would it be possible to move this to `io::parquet`? People can use this crate without parquet :)

Basically, this is an assumption that we do when we write to to parquet, so somewhere around `io::parquet::write`?

TurnOfACard:
> Thanks! Would it be possible to move this to `io::parquet`? People can use this crate without parquet :)
> 
> Basically, this is an assumption that we do when we write to to parquet, so somewhere around `io::parquet::write`?

A good point; I have moved it to `io::parquet::write` and also enumerated other types I did not originally think of as lacking logical types according to `io::parquet::write::schema::to_parquet_type`.

jorgecarleitao:
Thanks @TurnOfACard !

PR #1253, opened 2022-09-19
Fixed error reading unbounded Avro list

### LINKED ISSUE #1252: Potential bug in reading lists from avro?
Line of code: https://github.com/jorgecarleitao/arrow2/blob/c615095dbfb930a247c23423e36118843f71e1d2/src/io/avro/read/deserialize.rs#L136

Feel free to close this issue if my understanding is incorrect here. Referencing the Java Avro implementation, looks like array values are stored/read as blocks and can be read consecutively until a 0 is encountered. (https://github.com/apache/avro/blob/42822886c28ea74a744abb7e7a80a942c540faa5/lang/java/avro/src/main/java/org/apache/avro/io/Decoder.java#L203)

In the Avro deserialisation logic for reading in lists, looks like we are making a call to `try_push_valid` for each block in a list item, rather than just for the whole item (outside of the `loop`). Intuitively, it looks like this would be incorrect behavior since blocks are an Avro implementation detail and our validity bitmap is tracking Arrow types (e.g. a list item). 

I wasn't able to easily create a list in Avro that is stored as multiple blocks, so as to validate my assumptions, but wanted to open this issue in case it is an obvious bug. Thank you

jorgecarleitao:
Fixed #1252 - Thanks to @shaeqahmed for the report!

### DIFF
--- a/src/io/avro/read/deserialize.rs
+++ b/src/io/avro/read/deserialize.rs
@@ -122,7 +122,9 @@ fn deserialize_value<'a>(
                 .as_mut_any()
                 .downcast_mut::<DynMutableListArray<i32>>()
                 .unwrap();
+            // Arrays are encoded as a series of blocks.
             loop {
+                // Each block consists of a long count value, followed by that many array items.
                 let len = util::zigzag_i64(&mut block)?;
                 let len = if len < 0 {
                     // Avro spec: If a block's count is negative, its absolute value is used,
@@ -134,16 +136,18 @@ fn deserialize_value<'a>(
                     len
                 };
 
+                // A block with count zero indicates the end of the array.
                 if len == 0 {
                     break;
                 }
 
+                // Each item is encoded per the array’s item schema.
                 let values = array.mut_values();
                 for _ in 0..len {
                     block = deserialize_item(values, is_nullable, avro_inner, block)?;
                 }
-                array.try_push_valid()?;
             }
+            array.try_push_valid()?;
         }
         DataType::Struct(inner_fields) => {
             let fields = match avro_field {
--- a/src/io/avro/read/deserialize.rs
+++ b/src/io/avro/read/deserialize.rs
@@ -123,7 +123,16 @@ fn deserialize_value<'a>(
                 .downcast_mut::<DynMutableListArray<i32>>()
                 .unwrap();
             loop {
-                let len = util::zigzag_i64(&mut block)? as usize;
+                let len = util::zigzag_i64(&mut block)?;
+                let len = if len < 0 {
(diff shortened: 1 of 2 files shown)


src/io/avro/read/deserialize.rs
```diff
@@ -123,7 +123,16 @@ fn deserialize_value<'a>(
...
+                let len = util::zigzag_i64(&mut block)?;
+                let len = if len < 0 {
+                    // Avro spec: If a block's count is negative, its absolute value is used,
+                    // and the count is followed immediately by a long block size indicating the number of bytes in the block. This block size permits fast skipping through data, e.g., when projecting a record to a subset of its fields.
+                    let _ = util::zigzag_i64(&mut block)?;
+
+                    -len
+                } else {
+                    len
+                };
```
shaeqahmed, line 135: This PR looks correct to me, but it actually fixes a different bug than the one I was referring to in #1252. The fix here addresses the handling the case when an avro list item block encodes an optional `byte_size` for data skipping as `<-1*block_count><byte_size>...`.

The issue I was referring to, however, is the call to  `array.try_push_valid()?;` on line 145 (136 original), which is called for *each* block deserialised. Shouldn't this call be moved right outside of the `loop {...}` onto line 147, so that we only push valid *once* per list item? 

Thinking out loud, for an empty list, looks like the `if len == 0 { break; }` causes this loop to short circuit currently meaning we won't push a valid. I wonder if this is also an unintended bug, of having this call to try_push_valid in the block unrolling loop as opposed to outside.

To clarify, I am concerned about the handling of the following two cases:

- `List [1,2,3,4]`, encoded as two "blocks"
  - Currently we make **2 calls** to `try_push_valid` whereas I would **expect 1**

- `List []`, encoded as 0 blocks, indicated by a prefixed byte of `len == 0`
  - Currently we make **0 calls** to `try_push_valid` whereas I would **expect 1**, as you mentioned before that empty lists/structs are still "valid" in Arrow so as to maintain O(1) validity checks.

Again, I am not well versed in the Avro format, so feel free to correct me if I have an incorrect understanding here. I came across these potential bugs while reading the deserialisation code thoroughly, as I was considering taking a stab at implementing Map type support for Avro, which will be very similar to `List`. Thank you

shaeqahmed (changes requested):

src/io/avro/read/deserialize.rs
jorgecarleitao: Agree. Really good call.

Did another push with an extra fix

shaeqahmed: Awesome, LGTM

PR #1267, opened 2022-10-04
Bump json_deserializer to 0.4.2

cjermain:
This PR bumps `json_deserializer` to v0.4.2 to address the fix in jorgecarleitao/json-deserializer#14. I tested this locally against features in pola-rs/polars#3413 and confirmed the fixes are working correctly. Thanks for pushing out the release!


jorgecarleitao:
Thanks! 

Note that this is not necessary - cargo picks the latest version within the constraint (patch in this case); running `cargo update` should suffice

cjermain:
Good to know, it sounds like leaving this as `0.4` would have been sufficient to pick up the patches. Thanks!

PR #1271, opened 2022-10-10
Upgrade to ahash v0.8

benesch:
Hey @jorgecarleitao! Just looking to avoid a duplicate dependency downstream, and this looked to drop in easily.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ hash_hasher = "^2.0.3"
 # For SIMD utf8 validation
 simdutf8 = "0.1.3"
 # faster hashing
-ahash = { version = "0.7" }
+ahash = { version = "0.8" }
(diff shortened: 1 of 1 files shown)


jorgecarleitao:
This broke main and I missed it somehow when approving :/

jorgecarleitao:
I just reverted this until we can fix it - see #1274

benesch:
Thanks for being so on top of it! #1274 seems to have passed tests, unlike this PR! Anything you need me to do?

jorgecarleitao:
We need to fix the #1274 - it is failing to compile on wasm32

benesch:
> We need to fix the #1274 - it is failing to compile on wasm32

Ah, thanks! I believe I've fixed it in #1297.

Issue #1178, opened 2022-07-22
io::json enhancements


AnIrishDuck:
Hey!

We'd like to add some functionality to `io::json`, and I wanted to check if it would make sense as a contribution to the project.

In particular, we'd like to add support for the `records`-oriented output of the `to_json` and `from_json` Pandas methods. We probably also want to add support for the `split` orientation in the near future, but we'd like to start with `records`.

This is the rough path I've worked out to do so:

- Add a new `json::write::RecordSerializer` alongside `json::write::Serializer`. Instead of outputting a batch of arrays, this takes a `Chunk` and outputs record blobs in the transposed format.
- Create an `infer_records_schema` method in `json::read::infer_schema`. This would infer a `Schema` from the first row of inputted `json` in Pandas `records` format.
- Add a `json::read::deserialize_from_records` that takes the above `Schema` and outputs the correct `Chunk`. This is the hardest part to untangle so far. My current rough plan of attack is to modify the `deserialize_xxx` methods to `deserialize_into_xxx`, and take a `&mut` mutable array as a parameter instead of outputting finalized arrays. We can then "extend" the right data into each array per record, avoiding copies as we go. This is starting to feel like a pretty invasive change, in contrast to the previous two.

I'm still figuring the last part out. I don't feel fully stuck, but I'm worried about the scope of the changes. Here's my questions:

- Does the overall goal here fit with the direction of this project? Is it functionality you'd like to see in this crate?
- Does the above plan make sense? Are there tactical errors I'm making due to my lack of experience with the code?
- Should this even be in the `io::json` module? I see other modules for other types of JSON in there (`ndjson`, `json_integration`), but I'm not sure what the cutoff is for a new `io` module. This feels closer to the original `io::json` than either of those two formats?

Finally: thanks for working so hard on this crate. It's been hugely useful for us at Wallaroo.

jorgecarleitao:
Thank you for this issue and for the through explanation. 🙇 I think this makes a lot of sense indeed! 

Answering the questions:

1. Yes - interoperability with standard formats fits exactly into `io`. Record-based is quite a standard way of representing data in json, so this fits quite well. 👍
2. Yes - 
    2.1. serialize: the `io::csv` and `io::avro` has similar ideas that we can piggy back, since they are also row-based formats (and thus required a transposition). The relevant aspect there is that a serializer of a column is a `StreamingIterator` that attaches itself to an array (or its `.iter` and serializes each of its items to the corresponding format).
[...]

PR #1275, opened 2022-10-13
Added support for JSON ser/de records layout

AnIrishDuck:
Prior discussion in #1178

Serialization was easy: individual streaming iterators already produce a record at a time, so we just need to salami slice them to transpose the results.

Deserialization was more complex:
- Define `Preallocate` for arrays that have a backing that can be opportunistically resized.
    - Add expand to `MutableListArray`. This is the workhorse that makes later arbitrarily-nested deserialization possible.
    - Declare an `impl MutableArray` for `Box<dyn MutableArray>`. This is mostly trivial delegation, so it may be worth investigating the [delegate crate](https://docs.rs/delegate/latest/delegate/) for boilerplate reduction. We need this to support arbitrary recursion at the type level. This allows us to define a `MutableListArray<O, Box<dyn MutableArray>>`.
    - Finally, implement schema inference and deserialization. Here be dragons:
        - Schema inference basically recurses down to the first element for each list, and assumes that element and array shape for all future records. There are obvious pitfalls here.
        - For deserialization, we want to avoid an expensive transpose. So we instead allocate mutable arrays for each record in the schema, and expand any nested list arrays recursively after filling the recursive sub-level. This requires converting a bunch of deserialize_xxx methods into deserialize_xxx_into methods that extend an existing array instead of returning a new array. We can generally use adapters like fill_array_from to avoid code duplication in the original case where we just need a plain list returned.

It's worth noting that this approach may be useful in general to avoid some of the allocations that the json deserializer currently performs (creating `Vec` of row references).

### DIFF
--- a/src/array/binary/mutable.rs
+++ b/src/array/binary/mutable.rs
@@ -1,7 +1,9 @@
 use std::{iter::FromIterator, sync::Arc};
 
 use crate::{
-    array::{specification::check_offsets, Array, MutableArray, Offset, TryExtend, TryPush},
+    array::{
+        specification::check_offsets, Array, MutableArray, Offset, Preallocate, TryExtend, TryPush,
+    },
     bitmap::MutableBitmap,
     datatypes::DataType,
     error::{Error, Result},
@@ -183,6 +185,12 @@ impl<O: Offset> MutableBinaryArray<O> {
     /// returns its offsets.
     pub fn offsets(&self) -> &Vec<O> {
         &self.offsets
+    }
+}
+
+impl<O: Offset> Preallocate for MutableBinaryArray<O> {
+    fn with_capacity(capacity: usize) -> Self {
+        MutableBinaryArray::with_capacity(capacity)
     }
 }
 
--- a/src/array/binary/mutable.rs
+++ b/src/array/binary/mutable.rs
@@ -1,9 +1,7 @@
 use std::{iter::FromIterator, sync::Arc};
 
 use crate::{
-    array::{
-        specification::check_offsets, Array, Container, MutableArray, Offset, TryExtend, TryPush,
-    },
+    array::{specification::check_offsets, Array, MutableArray, Offset, TryExtend, TryPush},
     bitmap::MutableBitmap,
     datatypes::DataType,
     error::{Error, Result},
@@ -185,12 +183,6 @@ impl<O: Offset> MutableBinaryArray<O> {
     /// returns its offsets.
     pub fn offsets(&self) -> &Vec<O> {
         &self.offsets
-    }
-}
-
-impl<O: Offset> Container for MutableBinaryArray<O> {
-    fn with_capacity(capacity: usize) -> Self {
-        MutableBinaryArray::with_capacity(capacity)
     }
 }
 
--- a/src/array/binary/mutable.rs
+++ b/src/array/binary/mutable.rs
@@ -2,7 +2,7 @@ use std::{iter::FromIterator, sync::Arc};
 
 use crate::{
     array::{
-        specification::check_offsets, Array, MutableArray, Offset, Preallocate, TryExtend, TryPush,
+        specification::check_offsets, Array, Container, MutableArray, Offset, TryExtend, TryPush,
     },
     bitmap::MutableBitmap,
     datatypes::DataType,
@@ -188,7 +188,7 @@ impl<O: Offset> MutableBinaryArray<O> {
     }
 }
 
-impl<O: Offset> Preallocate for MutableBinaryArray<O> {
+impl<O: Offset> Container for MutableBinaryArray<O> {
     fn with_capacity(capacity: usize) -> Self {
         MutableBinaryArray::with_capacity(capacity)
     }
--- a/src/array/list/mutable.rs
+++ b/src/array/list/mutable.rs
@@ -1,10 +1,14 @@
 use std::sync::Arc;
 
 use crate::{
-    array::{Array, MutableArray, Offset, Preallocate, TryExtend, TryPush},
+    array::{
+        specification::try_check_offsets, Array, MutableArray, Offset, Preallocate, TryExtend,
+        TryPush,
+    },
     bitmap::MutableBitmap,
     datatypes::{DataType, Field},
     error::{Error, Result},
+    trusted_len::TrustedLen,
 };
 
 use super::ListArray;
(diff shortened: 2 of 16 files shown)


AnIrishDuck:
Thoughts on using [`match-downcast`](https://crates.io/crates/match-downcast) to get rid of the if-else ugliness? Clippy just reminded me that it is indeed, quite gross.

ritchie46:
> Thoughts on using [`match-downcast`](https://crates.io/crates/match-downcast) to get rid of the if-else ugliness? Clippy just reminded me that it is indeed, quite gross.

I don't think that's worth another crate pulled in.

src/array/mod.rs
```diff
@@ -170,6 +180,49 @@ pub trait MutableArray: std::fmt::Debug + Send + Sync {
     fn shrink_to_fit(&mut self);
 }
 
+impl MutableArray for Box<dyn MutableArray> {
```
jorgecarleitao, line 183: this is neat!

src/io/json/read/deserialize.rs
```diff
@@ -243,6 +252,137 @@ fn deserialize_list<'a, O: Offset, A: Borrow<Value<'a>>>(
...
+        false
+    }
+}
+
+/// Deserialize `rows` by extending them into the given `target`
+fn deserialize_into<'a, A: Borrow<Value<'a>>>(target: &mut Box<dyn MutableArray>, rows: &[A]) {
+    // It'd be nice to have something like pattern matching for downcasting from Any
+    // I'm not aware of anything like that, which leads to this ... ugliness
+    #[allow(clippy::if_same_then_else)]
+    if let Some(list_array) = target
```
jorgecarleitao, line 346: we have been using a different approach for these:

```
match self.data_type().to_logical_type() {
    ...
}
```

see e.g. https://github.com/jorgecarleitao/arrow2/blob/main/src/array/equal/mod.rs#L208

could we use this pattern instead?

tests/it/io/json/read.rs
```diff
@@ -33,3 +33,138 @@ fn read_json() -> Result<()> {
...
+    );
+    a.try_extend(a_iter).unwrap();
+    let a_expected: FixedSizeListArray = a.into();
+
+    let json = json_deserializer::parse(data)?;
+
+    let schema = Schema {
+        fields: vec![Field::new("a", a_expected.data_type().clone(), true)],
+        metadata: Metadata::default(),
+    };
```
jorgecarleitao, line 155:
```suggestion
    let schema: Schema = vec![Field::new("a", a_expected.data_type().clone(), true)].into();
```

tests/it/io/json/read.rs
```diff
@@ -33,3 +33,138 @@ fn read_json() -> Result<()> {
...
+
+    for (f, arr) in schema.fields.iter().zip(actual.arrays().iter()) {
+        let (expected, actual) = if f.name == "a" {
+            (&a_expected, arr.as_ref())
+        } else {
+            panic!("unexpected field found: {}", f.name);
+        };
+
+        // No idea why assert_eq! doesn't work here, but this does.
+        assert_eq!(format!("{:?}", expected), format!("{:?}", actual));
```
jorgecarleitao, line 166: I suspect that this is because line 144 uses `false` on the nullability of the inner field, but the result is a nullability true

tests/it/io/json/write.rs
```diff
@@ -211,6 +211,82 @@ fn nested_list() -> Result<()> {
...
+
+    let c2 = Utf8Array::<i32>::from(&vec![Some("foo"), Some("bar"), None]);
+
+    let schema = Schema {
+        fields: vec![
+            Field::new("c1", c1.data_type().clone(), true),
+            Field::new("c2", c2.data_type().clone(), true),
+        ],
+        metadata: Metadata::default(),
+    };
```
jorgecarleitao, line 244:
```suggestion
    let schema: Schema = vec![
            Field::new("c1", c1.data_type().clone(), true),
            Field::new("c2", c2.data_type().clone(), true),
        ].into();
```

src/array/list/mutable.rs
```diff
@@ -152,6 +156,75 @@ impl<O: Offset, M: MutableArray> MutableListArray<O, M> {
...
+    /// Assumes the expansion begins at the highest previous offset, or zero if
+    /// this [MutableListArray] is currently empty.
+    ///
+    /// Panics if:
+    /// - the new offsets are not in monotonic increasing order.
+    /// - any new offset is not in bounds of the backing array.
+    /// - the passed iterator has no upper bound.
+    pub fn extend_offsets<II>(&mut self, expansion: II)
+    where
+        II: IntoIterator<Item = Option<O>> + TrustedLen,
```
jorgecarleitao, line 169:
```suggestion
        II: TrustedLen<Item = Option<O>>,
```

src/array/list/mutable.rs
```diff
@@ -152,6 +156,75 @@ impl<O: Offset, M: MutableArray> MutableListArray<O, M> {
...
 
+    /// Expand this array, using elements from the underlying backing array.
+    /// Assumes the expansion begins at the highest previous offset, or zero if
+    /// this [MutableListArray] is currently empty.
+    ///
+    /// Panics if:
+    /// - the new offsets are not in monotonic increasing order.
+    /// - any new offset is not in bounds of the backing array.
+    /// - the passed iterator has no upper bound.
+    pub fn extend_offsets<II>(&mut self, expansion: II)
```
jorgecarleitao, line 167:
```suggestion
    pub(crate) fn extend_offsets<II>(&mut self, expansion: II)
```

let's keep this private for now ^^

src/array/list/mutable.rs
```diff
@@ -152,6 +156,75 @@ impl<O: Offset, M: MutableArray> MutableListArray<O, M> {
...
+    /// # Safety
+    ///
+    /// Assumes that `offsets` are in order, and do not overrun the underlying
+    /// `values` backing array.
+    ///
+    /// Also assumes the expansion begins at the highest previous offset, or
+    /// zero if the array is currently empty.
+    ///
+    /// Panics if the passed iterator has no upper bound.
+    pub unsafe fn unsafe_extend_offsets<II>(&mut self, expansion: II)
```
jorgecarleitao, line 200:
```suggestion
    pub(crate) unsafe fn unsafe_extend_offsets<II>(&mut self, expansion: II)
```

jorgecarleitao (commented):
This looks ready great

My only major comment is to keep things private - i.e. keep all changes needed for the json private to this crate, since they are only used in the context of the crate.

Great PR!!

src/io/json/read/deserialize.rs
AnIrishDuck: My concern there is we'd still need to do downcasting in each match arm, and then unwrapping. Example:

```
match self.data_type().to_logical_type() {
  // ...
  DataType::Float32 => 
    deserialize_into(target.as_mut_any().downcast_mut::<MutablePrimitiveArray<f32>>().unwrap(), rows),
```

Is there any way to avoid that `.unwrap()` in each arm? It should always succeed, but it feels kinda ugly. Not sure if I'm missing a better way.

src/array/binary/mutable.rs
```diff
@@ -186,6 +188,12 @@ impl<O: Offset> MutableBinaryArray<O> {
     }
 }
 
+impl<O: Offset> Container for MutableBinaryArray<O> {
```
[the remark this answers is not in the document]
AnIrishDuck: Yup, I can move all these `impl`s into there.

src/io/json/read/deserialize.rs
jorgecarleitao: afaik there is no better way since the invariant "`DataType` is `DataType::Int32` implies `&dyn Array` is `&PrimitiveArray<i32>` and `&dyn MutableArray` is `&MutablePrimitiveArray<i32>`" is a runtime invariant. This enforcement is `safe`, which means that in theory we could even use `.unwrap_unchecked` (but it is not worth the risk imo).

src/array/mod.rs
```diff
@@ -113,6 +114,15 @@ pub trait Array: Send + Sync + dyn_clone::DynClone + 'static {
 
 dyn_clone::clone_trait_object!(Array);
 
+/// A trait describing an array with a backing store that can be preallocated to
+/// a given size.
+pub trait Container {
```
jorgecarleitao, line 119:
```suggestion
pub(crate) trait Container {
```

tests/it/io/json/read.rs
AnIrishDuck: Yup, this was 100% it. Kinda an interesting philosophical question on whether they are actually equal... value-wise, they are identical, even though one _could_ hold nullable values. While I'm not 100% sure I agree with the conclusion, I understand and respect the reasoning.

AnIrishDuck: The last thought I'll leave here is that the error when they are not equal is pretty confusing:

```
thread 'io::json::read::read_json_records' panicked at 'assertion failed: `(left == right)`
  left: `ListArray[[[1.1, 2, 3], [2, 3], [4, 5, 6]], [[3, 2, 1], [3, 2], [6, 5, 4]]]`,
 right: `ListArray[[[1.1, 2, 3], [2, 3], [4, 5, 6]], [[3, 2, 1], [3, 2], [6, 5, 4]]]`', tests/it/io/json/read.rs:114:9
```

Not sure if the solution is to somehow mark that one is nullable and one is not.

AnIrishDuck: Yup, that makes sense. Especially with deeply nested types.

jorgecarleitao:
Thank you so much! Awesome feature and PR! 🙇

PR #1278, opened 2022-10-21
Added `TryExtendFromSelf`

### LINKED ISSUE #1268: Additional operations on `MutableListArray`
It would be nice if there was a way to append one `MutableListArray` into another, which seems hard to do with the current API. This would ideally be both possible and efficient.

It would probably also be nice to be able to iterate over the values in the `MutableListArray`.

jorgecarleitao:
This PR adds a trait implemented for most mutable arrays that tells how a mutable array extends from itself, `fn try_extend_from_self(&mut self, other: &Self) -> Result<()>`

This also adds some functions that were missing in mutable counterparts to make it more idiomatic to write them (such as `Clone` and `from`).

Closes #1268


TethysSvensson:
This looks good! :+1:

TethysSvensson:
Can you also implement it for `MutableDictionaryArray` ?

jorgecarleitao:
> Can you also implement it for `MutableDictionaryArray` ?

I tried, but it is not so easy. I will split it into another issue/PR since it requires some trait thinking.

PR #1279, opened 2022-10-21
Add avro record names when converting arrow schema to avro

### LINKED ISSUE #1269: Avro schema: Invalid record names
Currently, when mapping an Arrow schema to Avro schema, a struct type is mapped to an Avro record with empty string ("") as the Avro record name.

See https://github.com/jorgecarleitao/arrow2/blob/main/src/io/avro/write/schema.rs#L54

This is incorrect. Record names in avro cannot be empty and cannot be repeated across different records.

So, currently avro files with written by arrow2 cannot be deserialized.

Would have to modify to use a real record name, probably the field name.

Samrose-Ahmed:
This PR fixes avro serialization by adding record names when converting Arrow schema to Avro schema.

Closes #1269

### DIFF
--- a/src/io/avro/write/schema.rs
+++ b/src/io/avro/write/schema.rs
@@ -8,10 +8,11 @@ use crate::error::{Error, Result};
 
 /// Converts a [`Schema`] to an Avro [`Record`].
 pub fn to_record(schema: &Schema) -> Result<Record> {
+    let mut name_counter: i32 = 0;
     let fields = schema
         .fields
         .iter()
-        .map(field_to_field)
+        .map(|f| field_to_field(f, &mut name_counter))
         .collect::<Result<_>>()?;
     Ok(Record {
         name: "".to_string(),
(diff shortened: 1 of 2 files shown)


src/io/avro/write/schema.rs
```diff
@@ -22,20 +23,25 @@ pub fn to_record(schema: &Schema) -> Result<Record> {
...
     } else {
-        _type_to_schema(data_type)?
+        _type_to_schema(data_type, name_counter)?
     })
 }
 
-fn _type_to_schema(data_type: &DataType) -> Result<AvroSchema> {
+fn _get_field_name(name_counter: &mut i32) -> String {
+    *name_counter += 1;
+    "r".to_owned() + &name_counter.to_string()
```
jorgecarleitao, line 41:
```suggestion
    f"r{name_counter}"
```

I think this is 1 allocation vs 3 allocations on the current situation (1 for the `r`, 1 for the `name_counter.to_string()`, one for the concatenation of the two)

src/io/avro/write/schema.rs
```diff
@@ -8,10 +8,11 @@ use crate::error::{Error, Result};
 
 /// Converts a [`Schema`] to an Avro [`Record`].
 pub fn to_record(schema: &Schema) -> Result<Record> {
+    let mut name_counter: i32 = 0;
     let fields = schema
         .fields
         .iter()
-        .map(field_to_field)
+        .map(|f| field_to_field(&f, &mut name_counter))
```
jorgecarleitao, line 15: won't this be unstable on addition of a new field? Could the counter be reset between fields, so that a field's name is independent of previous fields in the schema?

src/io/avro/write/schema.rs
```diff
@@ -48,13 +54,13 @@ fn _type_to_schema(data_type: &DataType) -> Result<AvroSchema> {
...
-            type_to_schema(&inner.data_type, inner.is_nullable)?,
+            type_to_schema(&inner.data_type, inner.is_nullable, name_counter)?,
         )),
         DataType::Struct(fields) => AvroSchema::Record(Record::new(
-            "",
+            _get_field_name(name_counter),
             fields
                 .iter()
-                .map(field_to_field)
+                .map(|f| field_to_field(&f, name_counter))
```
jorgecarleitao, line 63: an alternative approach is to make the items this iteration as `struct_name.item1`, `struct_name.item2`, etc, but I imagine that `struct_name` would still need to be unique across the schema, right?

jorgecarleitao (commented):
Great PR, thank you! 🙇 

Left some minor comments but otherwise good to go!

src/io/avro/write/schema.rs
Samrose-Ahmed: Yes the names have to be globally unique

src/io/avro/write/schema.rs
Samrose-Ahmed: Yes, perhaps you could, it wouldn't be a reset since needs to be globally unique but you could maybe increment the counter by some large number (would be weird). But imo it shouldn't be a goal to rely on the record names when converting an Arrow schema to an Avro schema.

jorgecarleitao:
Could you rebase and run `clippy`? I think this is ready to merge :)

Samrose-Ahmed:
Thanks I've run clippy and fmt and updated.

PR #1281, opened 2022-10-24
Added support to cast `FixedSizeListArray - ListArray`

AnIrishDuck:
This deliberately ignores the cases from List => FixedSizeList where the underlying offsets are not directly compatible (e.g. null entries or "sparse" lists).

"Repacking" the list properly would be ideal, but much more work. We have several use cases where we know that the list is in a compatible format.


jorgecarleitao:
Looks great! Thank you 🙇

PR #1284, opened 2022-10-29
Improved ZipValidity iterators

ritchie46:
Due to slicing it is pretty common to have an array with a validity that has a null count of 0. 

Given that we have cached the null count, we can use that information to determine if we need to take `Optional` branch in `ZipValidity`. This PR ensures that we can take the `Required` branch in case the `null_count/unset_bits` of a validity is 0.

### DIFF
--- a/src/bitmap/utils/zip_validity.rs
+++ b/src/bitmap/utils/zip_validity.rs
@@ -101,10 +101,13 @@ where
     V: Iterator<Item = bool>,
 {
     /// Returns a new [`ZipValidity`]
-    pub fn new(values: I, validity: Option<V>) -> Self {
+    pub fn new(values: I, validity: Option<V>, null_count: Option<usize>) -> Self {
         match validity {
-            Some(validity) => Self::Optional(ZipValidityIter::new(values, validity)),
-            None => Self::Required(values),
+            // only if we have a validity and there are nulls we will iterate them
+            Some(validity) if null_count != Some(0) => {
+                Self::Optional(ZipValidityIter::new(values, validity))
+            }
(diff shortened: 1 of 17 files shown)


src/bitmap/utils/zip_validity.rs
```diff
@@ -101,10 +101,13 @@ where
     V: Iterator<Item = bool>,
 {
     /// Returns a new [`ZipValidity`]
-    pub fn new(values: I, validity: Option<V>) -> Self {
+    pub fn new(values: I, validity: Option<V>, null_count: Option<usize>) -> Self {
```
jorgecarleitao, line 104: instead of passing the option here, could we pass

```rust
array.validity().and_then(|validity| (validity.unset_bits() > 0).then(|| validity.iter()))
```

to `Zip::new` instead? This keeps `Zip::new` with a single responsibility.

jorgecarleitao (commented):
Thanks!

This hints that we should check unsets on slices so that we do not keep the ref count on empty validities.

Regardless, this is a good optimization. I left a comment with a small proposal

src/bitmap/utils/zip_validity.rs
ritchie46: Yeap, much cleaner as well. That I didn't think of that one.

ritchie46:
I have added a `ZipValidity::new_with_validity` so that the

`array.validity().and_then(|validity| (validity.unset_bits() > 0).then(|| validity.iter()))`
 
branch can be done in a single place and is unlikely to be forgotten if we have designated constructor.

PR #1286, opened 2022-10-31
Fixed descending ordering when specify nulls first

sandflee:
`databend` is a database using arrow2 to sort the data, when running query like `select * from xx order by xx desc nulls first`,  null is the last value not first, seems a logic bug in `build_comparator_impl`

### DIFF
--- a/src/compute/merge_sort/mod.rs
+++ b/src/compute/merge_sort/mod.rs
@@ -509,8 +509,14 @@ pub fn build_comparator_impl<'a>(
             let descending = pairs[c].1.descending;
             let null_first = pairs[c].1.nulls_first;
             let (l_is_valid, r_is_valid, value_comparator) = &data[c];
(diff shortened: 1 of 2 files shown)


sundy-li:
Need a test case to cover

sandflee:
> Need a test case to cover

I will try to add a test, this may take a time

RinChanNOWWW:
> `databend` is a database using arrow2 to sort the data, when running query like `select * from xx order by xx desc nulls first`, null is the last value not first, seems a logic bug in `build_comparator_impl`

BTW, part of the bug will be solved after the row format is introduced into databend's sort.

sandflee:
> Need a test case to cover

add a ut in the latest commit

Xuanwo:
Https://github.com/jorgecarleitao/arrow2/pull/1299 could address the Clippy error.

This PR is ready for ready now.

jorgecarleitao:
Thanks a lot @sandflee, @sundy-li and @Xuanwo !

PR #1287, opened 2022-10-31
Added comparable row-oriented representation of a collection of [`Array`].

### LINKED ISSUE #1280: Is there a module like arrow-rs's `row` in arrow2?
https://docs.rs/arrow/latest/arrow/row/index.html

Description: A comparable row-oriented representation of a collection of [Array](https://docs.rs/arrow/latest/arrow/array/trait.Array.html)

RinChanNOWWW:
Implemtation of [a comparable row format](https://docs.rs/arrow/latest/arrow/row/index.html) in arrow2.

The `row` module is used to convert columns into rows by `SortOption`s. After converting columns to rows, we can compare two 'row's directly without `build_compare`. This is very useful for `order by` execution in query engines and can help to improve the performance.

A `Rows` is fix-sized, and use `Box<[u8]>`  to store data. So there won't be too much heap allocation like `Vec<u8>`.

Struture:

- fixed: encode fix-sized types.
- variale: encode variable-sized types.
- dictionary: encode dictionary type.
- interner: dictionary type sort helper.

closes #1280


src/lib.rs
```diff
@@ -12,8 +12,6 @@
 #![allow(clippy::type_complexity)]
 #![cfg_attr(docsrs, feature(doc_cfg))]
 #![cfg_attr(feature = "simd", feature(portable_simd))]
-#![feature(build_hasher_simple_hash_one)]
```
ritchie46: This has performance benefits.

We could add this behind a nightly feature (and maybe activate that feature in a build.rs file if a nightly compiler is detected).

We can write a custom trait to deal with the dispatch:

```rust
use std::hash::{BuildHasher, Hash};

pub trait HashSingle: BuildHasher {
    #[inline]
    fn hash_single<T: Hash>(&self, x: T) -> u64
    where
        Self: Sized,
    {
        #[cfg(feature = "nightly")]
        {
            self.hash_one(x)
        }
        #[cfg(not(feature = "nightly"))]
        {
            use std::hash::Hasher;
            let mut hasher = self.build_hasher();
            x.hash(&mut hasher);
            hasher.finish()
        }
    }
}

impl<T: BuildHasher> HashSingle for T {}

```

src/compute/sort/row/mod.rs
```diff
@@ -0,0 +1,1041 @@
...
+    fn next_back(&mut self) -> Option<Self::Item> {
+        if self.end == self.start {
+            return None;
+        }
+        let row = self.rows.row(self.end);
+        self.end -= 1;
+        Some(row)
+    }
+}
+
```
jorgecarleitao, line 334:
```suggestion

unsafe impl<'a> crate::trusted_len::TrustedLen for RowsIter<'a> {}
```

since this iterator is trusted len.

src/compute/sort/row/mod.rs
```diff
@@ -0,0 +1,1041 @@
...
+/// A comparable representation of a row
+///
+/// Two [`Row`] can be compared if they both belong to [`Rows`] returned by calls to
+/// [`RowConverter::convert_columns`] on the same [`RowConverter`]
+///
+/// Otherwise any ordering established by comparing the [`Row`] is arbitrary
+#[derive(Debug, Copy, Clone)]
+pub struct Row<'a> {
+    data: &'a [u8],
+    _fields: &'a Arc<[SortField]>, // currently not used.
```
jorgecarleitao, line 344:
```suggestion
```

not used :)

src/compute/sort/row/mod.rs
```diff
@@ -0,0 +1,1041 @@
...
+                .downcast_ref::<DictionaryArray<$T>>()
+                .unwrap();
+            encode_dictionary(out, column, dictionary.unwrap(), opts);
+        }),
+        t => unimplemented!("not yet implemented: {:?}", t),
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
jorgecarleitao, line 611: move to `tests/` (all tests to public APIs go there so that we verify that the functions and structs are available from the crate instead of `super`)

jorgecarleitao (approved):
This looks really great. Thanks a lot, @RinChanNOWWW !

I left minor comments that I do not think need an action - they are improvements to the API and performance optimizations that I thought about while going through it.

I will wait for the CI to be green and we can merge it 👍

src/compute/sort/row/mod.rs
```diff
@@ -0,0 +1,1041 @@
...
+            })
+            .collect::<Result<Vec<_>>>()?;
+
+        let mut rows = new_empty_rows(columns, &dictionaries, Arc::clone(&self.fields))?;
+
+        for ((column, field), dictionary) in
+            columns.iter().zip(self.fields.iter()).zip(dictionaries)
+        {
+            // We encode a column at a time to minimise dispatch overheads
+            encode_column(&mut rows, column, field.options, dictionary.as_deref())
```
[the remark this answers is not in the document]
jorgecarleitao: This is almost parallelizable - it is changing `rows`. However, there is still an optimization since modifying `rows` is `O(1)` but encoding is `O(C)`. Will continue to think about this.

RinChanNOWWW:
I'm adding the row encoding for List types. And then I will apply the advices above.

src/compute/sort/row/mod.rs
RinChanNOWWW: How about some tests using private methods of a module?

RinChanNOWWW: Should we spilit the two types of tests?

src/compute/sort/row/mod.rs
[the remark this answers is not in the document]
RinChanNOWWW: I'm puzzled about how to convert this into a parallel one?

src/compute/sort/row/mod.rs
jorgecarleitao: exactly - tests to public functionality go to public, rest stays here :)

src/compute/sort/row/mod.rs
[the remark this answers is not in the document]
jorgecarleitao: I agree.

The idea would be that we can build `Vec<u8>` per column in parallel and only write them to the rows afterwards. However, I realized later that "encoding" is very simple and low overhead, so I am not sure it is worth.

Cargo.toml
```diff
@@ -32,6 +32,8 @@ hash_hasher = "^2.0.3"
 simdutf8 = "0.1.3"
 # faster hashing
 ahash = { version = "0.7" }
+# A Rust port of SwissTable 
+hashbrown = { version = "0.12", default-features = false }
```
jorgecarleitao, line 36:
```suggestion
hashbrown = { version = "0.12", default-features = false, optional = true }
```

together with `compute_sort = ["compute_take", "hashbrown"]`

so that we do not drag this dependency when we do not need it.

jorgecarleitao (approved):

src/compute/sort/row/interner.rs
```diff
@@ -0,0 +1,430 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
alamb, line 1: FWIW this is quite similar (possibly copied directly from ) arrow-rs: https://github.com/apache/arrow-rs/blob/65d5576/arrow/src/row/interner.rs

Which is fine, I just wanted to give @tustvold credit where credit is due

RinChanNOWWW: Yes, the basic idea is the same as arrow-rs, and I adapted the codes to fit arrow2 data structures. Thanks for arrow-rs and @tustvold.

tustvold: FWIW I think it is pretty cool that the flow of technology is now from arrow to arrow2, it's a nice validation of the work I put into porting the ideas behind arrow2 across.Perhaps eventually we can combine efforts :smile:

build.rs
```diff
@@ -0,0 +1,7 @@
+use rustc_version::{version_meta, Channel};
+
+fn main() {
+    if version_meta().unwrap().channel == Channel::Nightly {
+        println!("cargo:rustc-cfg=nightly_build");
```
Xuanwo, line 5: I prefer to introduce a new feature called `nightly` instead of adding a new dep in `build-dependencies`.

And I think the current implementation is wrong.

```diff
- #![cfg_attr(feature = "nightly_build", feature(build_hasher_simple_hash_one))]
+ #![cfg_attr(nightly_build, feature(build_hasher_simple_hash_one))]
```

Xuanwo: We can take `hashbrown`'s implementation for a look: https://github.com/rust-lang/hashbrown/pull/292

src/compute/sort/row/interner.rs
```diff
@@ -0,0 +1,431 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use std::hash::{BuildHasher, Hash};
+use std::num::NonZeroU32;
+use std::ops::Index;
+
+use hashbrown::hash_map::RawEntryMut;
+use hashbrown::HashMap;
```
Xuanwo, line 23: To my best knowledge, rust std's `HashMap` uses the same implementation as `hashbrown`. Is there any reason to hashbrown directly?

build.rs
RinChanNOWWW: > I prefer to introduce a new feature called nightly instead of adding a new dep in build-dependencies.

But users need to add this to the features list. `build.rs` can help to select automatically.

src/compute/sort/row/interner.rs
RinChanNOWWW: > Is there any reason to hashbrown directly?

No, it just follows arrow-rs.

build.rs
Xuanwo: > But users need to add this to the features list. `build.rs` can help to select automatically.

I prefer adding a new feature instead. Leave this for @jorgecarleitao to decide.

src/compute/sort/row/interner.rs
Xuanwo: How about using std's `HashMap` instead? So we don't need to introduce `hashbrown`.

tustvold: At least historically the raw entry API, which this requires, isn't exposed in std

Xuanwo: > At least historically the raw entry API, which this requires to store the values separately from the HashMap, isn't exposed in std

Thanks for explanation!

Is the raw entry API we use here something different from https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html?

Xuanwo: By the way, we have https://doc.rust-lang.org/std/collections/hash_map/enum.RawEntryMut.html under nightly feature `hash_raw_entry`.

src/compute/sort/row/mod.rs
```diff
@@ -0,0 +1,1041 @@
...
+pub struct RowsIter<'a> {
+    rows: &'a Rows,
+    start: usize,
+    end: usize,
+}
+
+impl<'a> Iterator for RowsIter<'a> {
+    type Item = Row<'a>;
+
+    fn next(&mut self) -> Option<Self::Item> {
```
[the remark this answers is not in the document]
Xuanwo: Lesson learned.

build.rs
jorgecarleitao: I would also prefer a specific feature, but both are mergable :) - we use this for `simd` also where users activate simd if they are in nightly. Maybe the nightly feature can activate both simd and this path? I can work out the details once we merge this :)

src/compute/sort/row/interner.rs
jorgecarleitao: > FWIW I think it is pretty cool that the flow of technology is now from arrow to arrow2

I believe it has always been both ways - which is part of the beauty of open source - people can do (almost) whatever they want without having to be worried about who "owns" what.

Xuanwo:
For the latest CI failing in `Check and test / Clippy (pull_request)`: https://github.com/jorgecarleitao/arrow2/actions/runs/3390608894/jobs/5636710157

```shell
error: the borrowed expression implements the required traits
   --> src/io/parquet/write/pages.rs:301:48
    |
301 |         let boolean = BooleanArray::from_slice(&[false, false, true, true]).boxed();
    |                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: change this to: `[false, false, true, true]`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow
```

Welcome to rust 1.65! We can fix them by `cargo clippy --fix`.

RinChanNOWWW:
62 files have clippy problems under rust 1.65 (83 files under rust 1.67). We can fix them in another pr.

jorgecarleitao:
Thank you @tustvold for the contribution (eagerly await the blog post!), @RinChanNOWWW for the port, and @Xuanwo and @ritchie46 for reviewing it! 🙇

src/compute/sort/row/interner.rs
ritchie46: Awesome work on thus @tustvold!

alamb:
> Thank you @tustvold for the contribution (eagerly await the blog post!),

In case anyone missed it, here is the blog post about the format

https://arrow.apache.org/blog/2022/11/07/multi-column-sorts-in-arrow-rust-part-1/
https://arrow.apache.org/blog/2022/11/07/multi-column-sorts-in-arrow-rust-part-2/

RinChanNOWWW:
image

> and is then followed by the unpadded length of this final block as a single byte in place of a continuation token

The demo in https://arrow.apache.org/blog/2022/11/07/multi-column-sorts-in-arrow-rust-part-2/ seems be wrong.

alamb:
Thanks for the report @RinChanNOWWW  -- there is a PR to correct this https://github.com/apache/arrow-site/pull/270

PR #1290, opened 2022-11-04
Fixed link from the API to the guide

datapythonista:
Not sure exactly how things are set up, but https://jorgecarleitao.github.io/arrow2/ seems to be an old version of the guide, with the link to the IO docs broken. I see that in the README that https://jorgecarleitao.github.io/arrow2/main/guide/ seems to be the latest version of the guide, so using it in the API docs.

### DIFF
--- a/src/doc/lib.md
+++ b/src/doc/lib.md
@@ -5,7 +5,7 @@ This is a library for efficient in-memory data operations with
 It is a re-write from the bottom up of the official `arrow` crate with soundness
 and type safety in mind.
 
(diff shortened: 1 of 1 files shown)


Xuanwo:
`gh-pages` seems deprecated for a while. We can migrate to GitHub-actions instead: https://github.blog/2022-08-10-github-pages-now-uses-actions-by-default/

jorgecarleitao:
Obrigado, @datapythonista 🙇

PR #1299, opened 2022-11-13
Fix deprecated warning from chrono

Xuanwo:
Signed-off-by: Xuanwo <github@xuanwo.io>

### DIFF
--- a/src/array/list/mutable.rs
+++ b/src/array/list/mutable.rs
@@ -181,6 +181,7 @@ impl<O: Offset, M: MutableArray> MutableListArray<O, M> {
     /// - the new offsets are not in monotonic increasing order.
     /// - any new offset is not in bounds of the backing array.
     /// - the passed iterator has no upper bound.
+    #[allow(dead_code)]
     pub(crate) fn extend_offsets<II>(&mut self, expansion: II)
     where
         II: TrustedLen<Item = Option<O>>,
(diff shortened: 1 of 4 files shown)


src/array/list/mutable.rs
```diff
@@ -181,6 +181,7 @@ impl<O: Offset, M: MutableArray> MutableListArray<O, M> {
     /// - the new offsets are not in monotonic increasing order.
     /// - any new offset is not in bounds of the backing array.
     /// - the passed iterator has no upper bound.
+    #[allow(dead_code)]
```
Xuanwo, line 184: I'm not sure do we need `extend_offsets` and `unsafe_extend_offsets` anymore.

Cc @jorgecarleitao for comments.

jorgecarleitao: I believe that this is only used when json feature is active.

Xuanwo: Since clippy didn't test with the `json` feature enabled, It's OK to add a `#[allow(dead_code)]` here?

benesch:
Random thought: you may want to consider checking Cargo.lock into version control! Even though standard practice is to avoid checking in Cargo.lock for library crates, I've found it helpful in my library-only projects, as it ensures that new releases of dependencies can't break CI in my repository.

Xuanwo:
> it ensures that new releases of dependencies can't break CI in my repository.

Interesting idea!

I will leave this for @jorgecarleitao to decide, although I prefer to break CI early. Checkout `Cargo.lock` in the repo will not change the fact that we need to resolve the deprecate warnings by hand, it only changes the time of finding them.

jorgecarleitao:
I also think it is beneficial to have the CI issues flagged sooner than later.

jorgecarleitao:
Thanks a lot for the PR, @Xuanwo !

PR #1300, opened 2022-11-17
Added support for Extension types in ffi

jondo2010:
Some nestings of Extension and Struct types need additional support in the ffi Schema handling.


jondo2010:
Clippy errors should be fixed now

PR #1301, opened 2022-11-17
Added `reserve` to pushable containers in parquet extend_from_decoder

ritchie46:
This ensure we reserve the amount of slots we are going to push into the container up front.

### DIFF
--- a/src/io/parquet/read/deserialize/binary/utils.rs
+++ b/src/io/parquet/read/deserialize/binary/utils.rs
@@ -25,6 +25,9 @@ impl<O: Offset> Offsets<O> {
 }
 
 impl<O: Offset> Pushable<O> for Offsets<O> {
+    fn reserve(&mut self, additional: usize) {
+        self.0.reserve(additional)
+    }
     #[inline]
     fn len(&self) -> usize {
         self.0.len() - 1
@@ -88,6 +91,11 @@ impl<O: Offset> Binary<O> {
 }
 
 impl<'a, O: Offset> Pushable<&'a [u8]> for Binary<O> {
+    fn reserve(&mut self, additional: usize) {
+        let avg_len = self.values.len() / std::cmp::max(self.last_offset.to_usize(), 1);
+        self.values.reserve(additional * avg_len);
+        self.offsets.reserve(additional);
+    }
     #[inline]
     fn len(&self) -> usize {
         self.len()
--- a/src/io/parquet/read/deserialize/binary/utils.rs
+++ b/src/io/parquet/read/deserialize/binary/utils.rs
@@ -91,6 +91,7 @@ impl<O: Offset> Binary<O> {
 }
 
 impl<'a, O: Offset> Pushable<&'a [u8]> for Binary<O> {
+    #[inline]
     fn reserve(&mut self, additional: usize) {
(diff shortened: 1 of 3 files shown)


src/io/parquet/read/deserialize/binary/utils.rs
```diff
@@ -25,6 +25,9 @@ impl<O: Offset> Offsets<O> {
 }
 
 impl<O: Offset> Pushable<O> for Offsets<O> {
+    fn reserve(&mut self, additional: usize) {
```
jorgecarleitao, line 28:
```suggestion
    #[inline]
    fn reserve(&mut self, additional: usize) {
```

some reserves are used inside loops

src/io/parquet/read/deserialize/binary/utils.rs
```diff
@@ -88,6 +91,11 @@ impl<O: Offset> Binary<O> {
 }
 
 impl<'a, O: Offset> Pushable<&'a [u8]> for Binary<O> {
+    fn reserve(&mut self, additional: usize) {
```
jorgecarleitao, line 94:
```suggestion
    #[inline]
    fn reserve(&mut self, additional: usize) {
```

src/io/parquet/read/deserialize/fixed_size_binary/utils.rs
```diff
@@ -30,6 +30,9 @@ impl FixedSizeBinary {
 }
 
 impl<'a> Pushable<&'a [u8]> for FixedSizeBinary {
+    fn reserve(&mut self, additional: usize) {
```
jorgecarleitao, line 33:
```suggestion
    #[inline]
    fn reserve(&mut self, additional: usize) {
```

src/io/parquet/read/deserialize/utils.rs
```diff
@@ -30,14 +30,17 @@ pub fn not_implemented(page: &DataPage) -> Error {
...
-    //fn reserve(&mut self, additional: usize);
+    fn reserve(&mut self, additional: usize);
     fn push(&mut self, value: T);
     fn len(&self) -> usize;
     fn push_null(&mut self);
     fn extend_constant(&mut self, additional: usize, value: T);
 }
 
 impl Pushable<bool> for MutableBitmap {
+    fn reserve(&mut self, additional: usize) {
```
jorgecarleitao, line 41:
```suggestion
    #[inline]
    fn reserve(&mut self, additional: usize) {
```

src/io/parquet/read/deserialize/utils.rs
```diff
@@ -60,6 +63,9 @@ impl Pushable<bool> for MutableBitmap {
 }
 
 impl<A: Copy + Default> Pushable<A> for Vec<A> {
+    fn reserve(&mut self, additional: usize) {
```
jorgecarleitao, line 66:
```suggestion
    #[inline]
    fn reserve(&mut self, additional: usize) {
```

jorgecarleitao (approved):
Thanks @ritchie46 !

ritchie46:
I have been thinking a bit about this one. How often do we call the function that does the reserve? If it only writes to that buffer in one function call, I think it makes sense to reserve up front. If we call this function more often (with the same `Pushable`) we will trigger a realloc every call because we don't have an exponential growing strategy.

PR #1303, opened 2022-11-23
feat(parquet): introduce data_pagesize_limit to write parquet pages

### LINKED ISSUE #1291: Add page_size opt to control array_to_pages
seems the `2^31 - 2^25` is a fixed value now, which may be to large.

```
 // maximum page size is 2^31 e.g. i32::MAX
    // we split at 2^31 - 2^25 to err on the safe side
    // we also check for an array.len > 3 to prevent infinite recursion
    // still have to figure out how to deal with values that are i32::MAX size, such as very large
    // strings or a list column with many elements
    if (estimated_bytes_size(array)) >= (2u32.pow(31) - 2u32.pow(25)) as usize && array.len() > 3 {
        let split_at = array.len() / 2;
        let left = array.slice(0, split_at);
        let right = array.slice(split_at, array.len() - split_at);

        Ok(DynIter::new(
            array_to_pages(&*left, type_.clone(), nested, options, encoding)?
                .chain(array_to_pages(&*right, type_, nested, options, encoding)?),
        ))
    }
```

`parquet.page.size:` The page size is for compression. When reading, each page can be decompressed independently. A block is composed of pages. The page is the smallest unit that must be read fully to access a single record. If this value is too small, the compression will deteriorate. Default size is 1048576 bytes (= 1 * 1024 * 1024).

sundy-li:
arrow2 writes single page by default which will hurt the read performance a lot (up to 4x slower than the original file).

This pr introduces an option named `data_pagesize_limit`  to split large Array into small pages.

fixes #1291

### DIFF
--- a/src/io/parquet/write/mod.rs
+++ b/src/io/parquet/write/mod.rs
@@ -56,7 +56,7 @@ pub struct WriteOptions {
     pub version: Version,
     /// The compression to apply to every page
     pub compression: CompressionOptions,
-    /// The size to flush a page, defaults to 1024 * 1024 if zero
+    /// The size to flush a page, defaults to 1024 * 1024 if None
     pub data_pagesize_limit: Option<usize>,
 }
 
@@ -172,8 +172,8 @@ pub fn array_to_pages(
                 const DEFAULT_PAGE_SIZE: usize = 1024 * 1024;
                 let page_size = options.data_pagesize_limit.unwrap_or(DEFAULT_PAGE_SIZE);
                 let bytes_per_row =
-                    (((array_byte_size as f64) / (array.len() as f64)) as usize).max(1);
-                let rows_per_page = (page_size / bytes_per_row).max(1);
+                    ((array_byte_size as f64) / (array.len() as f64)) as usize;
+                let rows_per_page = page_size / (bytes_per_row + 1);
 
                 let vs: Vec<Result<EncodedPage>> = (0..array.len())
                     .step_by(rows_per_page)
--- a/src/io/parquet/write/mod.rs
+++ b/src/io/parquet/write/mod.rs
@@ -143,8 +143,9 @@ pub fn array_to_pages(
     // we also check for an array.len > 3 to prevent infinite recursion
     // still have to figure out how to deal with values that are i32::MAX size, such as very large
     // strings or a list column with many elements
-
-    if (estimated_bytes_size(array)) >= (2u32.pow(31) - 2u32.pow(25)) as usize && array.len() > 3 {
+    
+    let array_byte_size = estimated_bytes_size(array.as_ref());
+    if array_byte_size >= (2u32.pow(31) - 2u32.pow(25)) as usize && array.len() > 3 {
         let split_at = array.len() / 2;
         let left = array.slice(0, split_at);
         let right = array.slice(split_at, array.len() - split_at);
@@ -169,27 +170,21 @@ pub fn array_to_pages(
             _ => {
                 const DEFAULT_PAGE_SIZE: usize = 1024 * 1024;
                 let page_size = options.data_pagesize_limit.unwrap_or(DEFAULT_PAGE_SIZE);
-
-                let bytes_per_row = estimated_bytes_size(array.slice(0, 1).as_ref()).max(1);
-                let row_per_page = page_size / bytes_per_row + 1;
-
-                if array.len() <= row_per_page {
-                    array_to_page(array, type_, nested, options, encoding)
-                        .map(|page| DynIter::new(std::iter::once(Ok(page))))
-                } else {
-                    let first = array.slice(0, row_per_page);
-                    let other = array.slice(row_per_page, array.len() - row_per_page);
-                    Ok(DynIter::new(
-                        array_to_pages(first.as_ref(), type_.clone(), nested, options, encoding)?
-                            .chain(array_to_pages(
-                                other.as_ref(),
-                                type_,
-                                nested,
-                                options,
-                                encoding,
-                            )?),
-                    ))
-                }
+                let bytes_per_row = ((array_byte_size as f64) / (array.len() as f64)) as usize;
+                let row_per_page = (page_size / bytes_per_row).max(1);
+                
+                let iter = (0..array.len()).step_by(row_per_page).map(move |offset| {
+                    let length = if offset + row_per_page > array.len() {
+                        array.len() - offset
+                    } else {
+                        row_per_page
+                    };
+                    
+                    let sub_array = array.slice(offset, length);
+                    array_to_page(sub_array.as_ref(), type_, nested, options, encoding)
+                });
+                
+                Ok(DynIter::new(iter))
             }
         }
     }
--- a/src/io/parquet/write/mod.rs
+++ b/src/io/parquet/write/mod.rs
@@ -131,6 +131,7 @@ pub fn can_encode(data_type: &DataType, encoding: Encoding) -> bool {
 }
 
 /// Returns an iterator of [`EncodedPage`].
+#[allow(clippy::needless_collect)]
 pub fn array_to_pages(
     array: &dyn Array,
     type_: ParquetPrimitiveType,
@@ -144,7 +145,7 @@ pub fn array_to_pages(
     // still have to figure out how to deal with values that are i32::MAX size, such as very large
     // strings or a list column with many elements
 
-    let array_byte_size = estimated_bytes_size(array.as_ref());
+    let array_byte_size = estimated_bytes_size(array);
     if array_byte_size >= (2u32.pow(31) - 2u32.pow(25)) as usize && array.len() > 3 {
         let split_at = array.len() / 2;
         let left = array.slice(0, split_at);
--- a/src/io/parquet/write/mod.rs
+++ b/src/io/parquet/write/mod.rs
@@ -56,6 +56,8 @@ pub struct WriteOptions {
     pub version: Version,
     /// The compression to apply to every page
     pub compression: CompressionOptions,
+    /// The size to flush a page, defaults to 1024 * 1024 if zero
+    pub data_pagesize_limit: Option<usize>,
 }
 
 use crate::compute::aggregate::estimated_bytes_size;
@@ -141,6 +143,7 @@ pub fn array_to_pages(
     // we also check for an array.len > 3 to prevent infinite recursion
     // still have to figure out how to deal with values that are i32::MAX size, such as very large
     // strings or a list column with many elements
+
     if (estimated_bytes_size(array)) >= (2u32.pow(31) - 2u32.pow(25)) as usize && array.len() > 3 {
         let split_at = array.len() / 2;
         let left = array.slice(0, split_at);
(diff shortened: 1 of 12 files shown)


sundy-li:
Read write benchmark gists:  https://gist.github.com/sundy-li/4984ec7cfeade556d60306a3a218ec8a


We use TPCH's  lineitem table, it's original file is generated by dremio, it's well paged for column `l_quantity`

```
column ["l_quantity"]: num_of_pages 460, num_of_rows 5192615, size 10454182, size(uncompressed) 41555178
```

We use [parquet_write.rs](https://gist.github.com/sundy-li/4984ec7cfeade556d60306a3a218ec8a#file-parquet_write-rs) to generate the output file by input file and  [parquet_read.rs](https://gist.github.com/sundy-li/4984ec7cfeade556d60306a3a218ec8a#file-parquet_read-rs) to test the performance read on parallel.


Before (single large page per rowgroup):

```
/tmp/input.parquet, parallel 1 read took: 21 ms
/tmp/input.parquet, parallel 2 read took: 21 ms
/tmp/input.parquet, parallel 4 read took: 23 ms
/tmp/input.parquet, parallel 8 read took: 37 ms
/tmp/input.parquet, parallel 16 read took: 53 ms
/tmp/output.parquet, parallel 1 read took: 37 ms
/tmp/output.parquet, parallel 2 read took: 42 ms
/tmp/output.parquet, parallel 4 read took: 65 ms
/tmp/output.parquet, parallel 8 read took: 124 ms
/tmp/output.parquet, parallel 16 read took: 284 ms
```

As you can see, it's 5X worse in reading on 16 parallelism.

After this pr:

```
/tmp/input.parquet, parallel 1 read took: 21 ms
/tmp/input.parquet, parallel 2 read took: 22 ms
/tmp/input.parquet, parallel 4 read took: 24 ms
/tmp/input.parquet, parallel 8 read took: 29 ms
/tmp/input.parquet, parallel 16 read took: 52 ms
/tmp/output.parquet, parallel 1 read took: 23 ms
/tmp/output.parquet, parallel 2 read took: 25 ms
/tmp/output.parquet, parallel 4 read took: 27 ms
/tmp/output.parquet, parallel 8 read took: 31 ms
/tmp/output.parquet, parallel 16 read took: 48 ms
```

src/io/parquet/write/mod.rs
```diff
@@ -163,8 +166,31 @@ pub fn array_to_pages(
                     )
                 })
             }
-            _ => array_to_page(array, type_, nested, options, encoding)
-                .map(|page| DynIter::new(std::iter::once(Ok(page)))),
+            _ => {
+                const DEFAULT_PAGE_SIZE: usize = 1024 * 1024;
+                let page_size = options.data_pagesize_limit.unwrap_or(DEFAULT_PAGE_SIZE);
+
+                let bytes_per_row = estimated_bytes_size(array.slice(0, 1).as_ref()).max(1);
```
jorgecarleitao, line 173:
```
let bytes_per_row = ((estimated_bytes_size(array.as_ref()) as f64) / array.len()).max(1); // needs some rounding to integer
```

likely produces more reliable estimates for variable sized arrays, as the first element may be an outlier.

jorgecarleitao:
Out of curiosity, what is the difference in file size? Usually smaller pages => larger files.

sundy-li:
> Out of curiosity, what is the difference in file size? Usually smaller pages => larger files.

It's a large file, 255MB snappy encoded.

```
column ["l_orderkey"]: num_of_pages 460, num_of_rows 5192615, size 8113390, size(uncompressed) 41555178
column ["l_partkey"]: num_of_pages 460, num_of_rows 5192615, size 25860456, size(uncompressed) 41555178
column ["l_suppkey"]: num_of_pages 460, num_of_rows 5192615, size 23360484, size(uncompressed) 41555178
column ["l_linenumber"]: num_of_pages 460, num_of_rows 5192615, size 4690406, size(uncompressed) 41555177
column ["l_quantity"]: num_of_pages 460, num_of_rows 5192615, size 10454182, size(uncompressed) 41555178
column ["l_extendedprice"]: num_of_pages 460, num_of_rows 5192615, size 29932431, size(uncompressed) 41555178
column ["l_discount"]: num_of_pages 460, num_of_rows 5192615, size 10101375, size(uncompressed) 41555178
column ["l_tax"]: num_of_pages 460, num_of_rows 5192615, size 9653229, size(uncompressed) 41555178
column ["l_returnflag"]: num_of_pages 289, num_of_rows 5192615, size 4907634, size(uncompressed) 25972033
column ["l_linestatus"]: num_of_pages 289, num_of_rows 5192615, size 3020518, size(uncompressed) 25972033
column ["l_shipdate"]: num_of_pages 231, num_of_rows 5192615, size 15957395, size(uncompressed) 20777621
column ["l_commitdate"]: num_of_pages 231, num_of_rows 5192615, size 15813887, size(uncompressed) 20777621
column ["l_receiptdate"]: num_of_pages 231, num_of_rows 5192615, size 16016467, size(uncompressed) 20777621
column ["l_shipinstruct"]: num_of_pages 914, num_of_rows 5192615, size 11190998, size(uncompressed) 83103193
column ["l_shipmode"]: num_of_pages 477, num_of_rows 5192615, size 11659513, size(uncompressed) 43039249
column ["l_comment"]: num_of_pages 1731, num_of_rows 5192615, size 67467279, size(uncompressed) 158417914
```

src/io/parquet/write/mod.rs
```diff
@@ -163,8 +166,31 @@ pub fn array_to_pages(
...
+                let row_per_page = page_size / bytes_per_row + 1;
+
+                if array.len() <= row_per_page {
+                    array_to_page(array, type_, nested, options, encoding)
+                        .map(|page| DynIter::new(std::iter::once(Ok(page))))
+                } else {
+                    let first = array.slice(0, row_per_page);
+                    let other = array.slice(row_per_page, array.len() - row_per_page);
+                    Ok(DynIter::new(
+                        array_to_pages(first.as_ref(), type_.clone(), nested, options, encoding)?
```
sundy-li, line 183: here may reach stackoverflow of recursion.

But if I change to 
```
 _ => {
                const DEFAULT_PAGE_SIZE: usize = 1024 * 1024;
                let page_size = options.data_pagesize_limit.unwrap_or(DEFAULT_PAGE_SIZE);
                let bytes_per_row = ((array_byte_size as f64) / (array.len() as f64)) as usize;
                let row_per_page = (page_size / bytes_per_row).max(1);
                
                let iter = (0..array.len()).step_by(row_per_page).map(move |offset| {
                    let length = if offset + row_per_page > array.len() {
                        array.len() - offset
                    } else {
                        row_per_page
                    };
                    
                    let sub_array = array.slice(offset, length);
                    array_to_page(sub_array.as_ref(), type_, nested, options, encoding)
                });
                
                Ok(DynIter::new(iter))
            }
```

may have lifetime error.

```
 Ok(DynIter::new(iter))
    |                 ^^^^^^^^^^^^^^^^^^^^^^ returning this value requires that `'3` must outlive `'static`
```

@jorgecarleitao  any advice about this?

sundy-li: I choose to collect the iter.

alamb:
There is a similar limit in arrow-rs:

https://docs.rs/parquet/27.0.0/parquet/file/properties/struct.WriterProperties.html#method.data_pagesize_limit


Depending on your usecase you may also find limiting the row count useful as some data compresses so well it fits on quite small pages: https://docs.rs/parquet/27.0.0/parquet/file/properties/struct.WriterProperties.html#method.data_page_row_count_limit

sundy-li:
> There is a similar limit in arrow-rs

Thank you @alamb,  I can't figure out a better name so I keep it compatible with arrow-rs.  The row count limit is less useful than the bytes limit, I'll try to add it in the next pr.

I found this best practice to optimize parquet files:  https://docs.dremio.com/software/data-formats/parquet-files/

The page-size defaults to ~100KB seem to be really reasonable. (I have test from 8192byte to 1MB today and found it has the best performance in this default config)

src/io/parquet/write/mod.rs
```diff
@@ -56,6 +56,8 @@ pub struct WriteOptions {
     pub version: Version,
     /// The compression to apply to every page
     pub compression: CompressionOptions,
+    /// The size to flush a page, defaults to 1024 * 1024 if zero
+    pub data_pagesize_limit: Option<usize>,
```
ritchie46, line 60: This should be `"The size to flush a page, defaults to 1024 * 1024 if None"`. Otherwise it could be interpreted as `Some(0)`.

src/io/parquet/write/mod.rs
```diff
@@ -163,8 +168,29 @@ pub fn array_to_pages(
...
                 })
             }
-            _ => array_to_page(array, type_, nested, options, encoding)
-                .map(|page| DynIter::new(std::iter::once(Ok(page)))),
+            _ => {
+                const DEFAULT_PAGE_SIZE: usize = 1024 * 1024;
+                let page_size = options.data_pagesize_limit.unwrap_or(DEFAULT_PAGE_SIZE);
+                let bytes_per_row =
+                    (((array_byte_size as f64) / (array.len() as f64)) as usize).max(1);
+                let rows_per_page = (page_size / bytes_per_row).max(1);
```
ritchie46, line 176: Maybe we should divide by `bytes_per_row + 1`? Are we certain that we never hit this branch with an empty array?

src/io/parquet/write/mod.rs
```diff
@@ -163,8 +168,29 @@ pub fn array_to_pages(
...
+                        let length = if offset + rows_per_page > array.len() {
+                            array.len() - offset
+                        } else {
+                            rows_per_page
+                        };
+
+                        let sub_array = array.slice(offset, length);
+                        array_to_page(sub_array.as_ref(), type_.clone(), nested, options, encoding)
+                    })
+                    .collect();
```
ritchie46, line 190: I don't think we need to collect here. We could `Box` the iterator as the `DynIter` is defined as:

```rust
pub struct DynIter<'a, V> {
    iter: Box<dyn Iterator<Item = V> + 'a + Send + Sync>,
}
```

src/io/parquet/write/mod.rs
sundy-li: `bytes_per_row + 1` may make `rows_per_page` less than the expect value if bytes_per_row is small.

As here `bytes_per_row` is a float value GTE than 1 and `rows_per_page` is GTE than 1, so if the array is empty, the iterator will be empty.

src/io/parquet/write/mod.rs
sundy-li: I tried to use
```
let iter = (0..array.len())
                    .step_by(rows_per_page)
                    .map(|offset| {
                        let length = if offset + rows_per_page > array.len() {
                            array.len() - offset
                        } else {
                            rows_per_page
                        };

                        let sub_array = array.slice(offset, length);
                        array_to_page(sub_array.as_ref(), type_.clone(), nested, options, encoding)
                    });

                Ok(DynIter::new(Box::new(iter)))
```

But got
```
   --> src/io/parquet/write/mod.rs:191:17
    |
138 |     nested: &[Nested],
    |     ------ has type `&[pages::Nested<'3>]`
...
191 |                 Ok(DynIter::new(Box::new(iter)))
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ returning this value requires that `'3` must outlive `'static`

```

Any advices?

ritchie46: And if we `move` the borrowed items into the closure?

sundy-li: Same error:

```
   Compiling arrow2 v0.14.2 (/home/sundy/work/arrow2)
error: lifetime may not live long enough
   --> src/io/parquet/write/mod.rs:191:17
    |
136 |     array: &dyn Array,
    |            - let's call the lifetime of this reference `'1`
...
191 |                 Ok(DynIter::new(Box::new(iter)))
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ returning this value requires that `'1` must outlive `'static`

error: lifetime may not live long enough
   --> src/io/parquet/write/mod.rs:191:17
    |
138 |     nested: &[Nested],
    |             - let's call the lifetime of this reference `'2`
...
191 |                 Ok(DynIter::new(Box::new(iter)))
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ returning this value requires that `'2` must outlive `'static`

error: lifetime may not live long enough
   --> src/io/parquet/write/mod.rs:191:17
    |
138 |     nested: &[Nested],
    |     ------ has type `&[pages::Nested<'3>]`
...
191 |                 Ok(DynIter::new(Box::new(iter)))
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ returning this value requires that `'3` must outlive `'static`

error: could not compile `arrow2` due to 3 previous errors
```

ritchie46: I checked locally and I think making that work is not worth it. We need to Box the `dyn Array` and clone the `&[Nested]` slice to do so. I think collecting the iterator is better.

jorgecarleitao:
Thanks a lot @sundy-li for the PR, and everyone else for the ideas, suggestions and reviews 🙇

PR #1307, opened 2022-11-28
perf: elide unneeded offset checks in reading parquet binary/utf8 columns

### LINKED ISSUE #1306: Reconsider forbid unsafe
I think we should reconsider `forbid_unsafe` on the IO module. Currently we are having redundant checks in the parquet reader that are perfectly sound to elide.

I saw that in my local benchmarks 10% of samples were in `check_offsets` when converting a `Binary<O>` to a `Utf8Array`. I believe that these offsets never have to be checked as they are created by us and are not read from disk. I think we only should check utf8 in that case.

We could probably create a safe wrapper around a `from` implementation, but that still requires `unsafe`.

ritchie46:
This would close #1306 and uses a `unsafe` conversion which is actually sound because we have carefully constructed the `Offsets` and uphold its invariant. This saves 5% runtime in my local benchmark of reading utf8 columns. I expect the maximum performance increase is ~10% when we deal with one character strings.

### DIFF
--- a/src/array/binary/mod.rs
+++ b/src/array/binary/mod.rs
@@ -24,6 +24,8 @@ mod from;
 mod mutable_values;
 pub use mutable_values::*;
 mod mutable;
+use crate::array::offsets::ValidOffsets;
+use crate::array::specification::check_offsets_minimal;
 pub use mutable::*;
 
 /// A [`BinaryArray`] is Arrow's semantically equivalent of an immutable `Vec<Option<Vec<u8>>>`.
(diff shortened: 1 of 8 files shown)


jorgecarleitao:
I was thinking about this and I think that what we are looking for here is to introduce `struct Offsets(Vec<O>)` and `struct OffsetsBuffer(Buffer<O>)` that upheld the invariants. We then change all arrays that contain offsets to use these structures accordingly, and stop validating offsets on the arrays themselves (they are validated by `OffsetsBuffer` and `Offsets`).

In other words, we elevate the importance of `ValidOffsets` in this PR to a core struct of the arrays (and introduce the corresponding mutable version).

What do you think about this approach? I think in json we have this same situation atm.

I can work on this.

ritchie46:
> struct Offsets(Vec<O>) and struct OffsetsBuffer(Buffer<O>) that upheld the invariants. 

Yeap, I have been thinking in the same direction, so must be a good idea. ^^

jorgecarleitao:
Draft PR for that: https://github.com/jorgecarleitao/arrow2/pull/1316

PR #1311, opened 2022-12-01
Added MutableMapArray

### LINKED ISSUE #1310: Feature: Add MutableMapArray
`MapArray` is currently left out of the mutable builder party.

jondo2010:
There's still a few things to do on this (2 todo!()'s still in the `MutableArray` impl), but wanted to get early feedback.

Closes #1310

### DIFF
--- a/src/array/map/mutable.rs
+++ b/src/array/map/mutable.rs
@@ -1,0 +1,176 @@
+use std::sync::Arc;
+
+use crate::{
+    array::{Array, MutableArray, MutableStructArray, StructArray},
+    bitmap::MutableBitmap,
+    datatypes::DataType,
+    error::{Error, Result},
+    types::Index,
+};
+
+use super::MapArray;
+
+/// The mutable version lf [`MapArray`].
+#[derive(Debug)]
+pub struct MutableMapArray {
+    data_type: DataType,
+    offsets: Vec<i32>,
+    field: MutableStructArray,
+    validity: Option<MutableBitmap>,
+}
+
+impl From<MutableMapArray> for MapArray {
+    fn from(other: MutableMapArray) -> Self {
+        let validity = if other.validity.as_ref().map(|x| x.unset_bits()).unwrap_or(0) > 0 {
+            other.validity.map(|x| x.into())
+        } else {
+            None
+        };
+
+        let field: StructArray = other.field.into();
+
+        MapArray::from_data(
+            other.data_type,
+            other.offsets.into(),
+            Box::new(field),
+            validity,
+        )
+    }
+}
+
+impl MutableMapArray {
+    /// Creates a new empty [`MutableMapArray`].
+    /// # Errors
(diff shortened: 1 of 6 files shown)


src/array/map/mutable.rs
```diff
@@ -0,0 +1,192 @@
...
+    array::{Array, MutableArray, MutableStructArray, StructArray},
+    bitmap::MutableBitmap,
+    datatypes::DataType,
+    error::{Error, Result},
+    types::Index,
+};
+
+use super::MapArray;
+
+/// The mutable version lf [`MapArray`].
```
jorgecarleitao, line 13:
```suggestion
/// The mutable version of [`MapArray`].
```

src/array/map/mutable.rs
```diff
@@ -0,0 +1,192 @@
...
+            validity.extend_constant(len, true);
+            validity.set(len - 1, false);
+        }
+        self.validity = Some(validity);
+    }
+
+    fn take_into(&mut self) -> MapArray {
+        MapArray::from_data(
+            self.data_type.clone(),
+            std::mem::take(&mut self.offsets).into(),
```
jorgecarleitao, line 137: if we take here, the offsets end up being empty, which I believe violates an invariant of `self`

src/array/struct_/mutable.rs
```diff
@@ -38,44 +39,77 @@ impl MutableStructArray {
...
-    /// # Panics
-    /// This function panics iff:
+    /// Fallibly create a [`MutableStructArray`] out of low-level APIs.
+    /// # Errors
+    /// This function returns an error if:
     /// * `data_type` is not [`DataType::Struct`]
     /// * The inner types of `data_type` are not equal to those of `values`
     /// * `validity` is not `None` and its length is different from the `values`'s length
-    pub fn from_data(
+    pub fn try_from_data(
```
jorgecarleitao, line 48: could we rename it to `try_new` instead? It is more aligned with the rest of the containers

jorgecarleitao (commented):
Thanks a lot for the contribution! This looks great!

I left minor comments, but otherwise I think we just need a bit more tests covering the different APIs.

One imo useful test here is:

```rust
let mut mutable = MutableMapArray::...
let a = mutable.take_into();
let b: MapArray = mutable.into()

assert_eq!(b.len(), 0)
```

src/array/map/mutable.rs
jondo2010: Ah yeah, the API on `MutableArray` is unclear here. Shouldn't `as_box` and `as_arc` take `self` instead of `&mut self`?

jorgecarleitao: Yes, but that would require `MutableArray` to not be object safe, and we need that for using it as a trait object.

jorgecarleitao:
Hey! Could you rebase so we can merge it in? :)

PR #1312, opened 2022-12-01
Fixed failing to accept dictionary full of nulls

ritchie46:
This fixes https://github.com/pola-rs/polars/issues/5655

### DIFF
--- a/src/array/dictionary/mod.rs
+++ b/src/array/dictionary/mod.rs
@@ -119,7 +119,9 @@ impl<K: DictionaryKey> DictionaryArray<K> {
     ) -> Result<Self, Error> {
         check_data_type(K::KEY_TYPE, &data_type, values.data_type())?;
 
-        check_indexes(keys.values(), values.len())?;
+        if keys.null_count() != keys.len() {
+            check_indexes(keys.values(), values.len())?;
+        }
 
         Ok(Self {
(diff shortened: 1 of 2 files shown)


src/array/dictionary/mod.rs
```diff
@@ -119,7 +119,9 @@ impl<K: DictionaryKey> DictionaryArray<K> {
     ) -> Result<Self, Error> {
         check_data_type(K::KEY_TYPE, &data_type, values.data_type())?;
 
-        check_indexes(keys.values(), values.len())?;
+        if keys.null_count() != keys.len() {
+            check_indexes(keys.values(), values.len())?;
```
jorgecarleitao, line 123: shouldln't we check that the valid keys are in bound?

ritchie46: Oh, my bad. This should be `keys.null_count() == keys.len()`, meaning that all key values are masked out. 

This is for instance the keys when we create `DictionaryArray([None, None])`. This keys will be `[0, 0]`, but the length of values will be `0`, so it would not suffice the index check.

jorgecarleitao: Agree - I mis-read the condition on the if. Thanks @ritchie46 ! 🙇

PR #1313, opened 2022-12-01
Improved performance of check_indexes

ritchie46:
This continues on the fix in #1312 to prevent conflicts.

Our generic `check_indexes` did a check on `usize` conversion which can be elided depending on the key type. This PR does the check on the type instead of on the value.

By doing this we can auto-vectorize check indexes and improve the performance when reading indexes from IO.


jorgecarleitao:
Thanks a lot, @ritchie46 ! 🙇

Issue #1317, opened 2022-12-05
Unnecessarily slow writes while writing arrow IPC


cyr:
As per the parquet2 issue, writing arrow IPC files do some flushing while writing, which won't let a BufWriter to fullly utilize its buffers. PR incoming!

PR #1318, opened 2022-12-05
Removed flushing during arrow IPC writing to improve performance when using a buffered writer

cyr:
As per https://github.com/jorgecarleitao/arrow2/issues/1317

This actually boost performance by a whole heap, getting close to 30% higher speeds when writing with a 1mb buffered writer during my own testing.


jorgecarleitao:
Thanks, @cyr , awesome improvement!

PR #1320, opened 2022-12-06
feat: add typed dictionary_iter

ritchie46:
Iterating over a `DictionaryArray` is very expensive as it allocates an opaque heap value of each iteration.

This adds a generic typed iterator that only compiles upon instantiation and thus will not add compiler bloat for all types. I want to follow up with a PR that implements writing `DictionaryArrays` to json. That was the reason I stumbled upon this.

The iterator impementation will `panic` iff the `values` contain null values. Otherwise it can become pretty awkward with nested arrays (which is something we could follow up on if needed)>

The existing iterators are flawed IMO. They will happily iterate over a `values` array with null values and return the masked out values. Shall we also add a panic check there? That's much better than return the wrong results.

### DIFF
--- a/tests/it/bitmap/mutable.rs
+++ b/tests/it/bitmap/mutable.rs
@@ -368,7 +368,11 @@ fn extend_bitmap() {
     assert_eq!(b.len(), 5 + 4);
 }
 
+// TODO! undo miri ignore once issue is fixed in miri
+// this test was a memory hog and lead to OOM in CI
+// given enough memory it was able to pass succesfully on a local
 #[test]
+#[cfg_attr(miri, ignore)]
 fn extend_constant1() {
     use std::iter::FromIterator;
     for i in 0..64 {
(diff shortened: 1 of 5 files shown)


ritchie46:
@jorgecarleitao could we update rustc in CI? I needed GATS here.

tests/it/bitmap/mutable.rs
```diff
@@ -368,7 +368,11 @@ fn extend_bitmap() {
     assert_eq!(b.len(), 5 + 4);
 }
 
+// TODO! undo miri ignore once issue is fixed in miri
+// this test was a memory hog and lead to OOM in CI
+// given enough memory it was able to pass succesfully on a local
 #[test]
```
ritchie46, line 374: MIRI used a lot of memory on this one and caused OOM. 

I think we can try to turn it on later once the issue is fixed in miri.

ritchie46 (commented):
@jorgecarleitao this is ready for review.

PR #1321, opened 2022-12-07
Bumped json-deserializer version

universalmind303:


jorgecarleitao:
Thanks @universalmind303 !

PR #1326, opened 2022-12-11
Fixed writing nested/sliced arrays to parquet

### LINKED ISSUE #1323: Parquet writes all values of sliced arrays?
During parquet writes we write smaller pages because of the `i32::limit` and because of performance improvements when writing smaller pages during reading.

Nested structures such as lists and utf8 are then sliced by their offsets, but the whole values are then send to the pages writer. I haven't yet confirmed this, but I believe this is what happens and is the reason for:

1: invalid parquet files
2: extreme memory usage
3: extreme file sizes

all reported in: https://github.com/pola-rs/polars/issues/4393

ritchie46:
fixes #1323 

We wrote invalid arrays to parquet because we did not take the offset of the list arrays into account. This got worse as we now slice leaf arrays when we write to pages and the offsets don't belong to the pages written anymore.

Because we don't use offsets, but dremel during reading/writing of the parquet this still read values, but just way too many.

This PR fixes the huges memory usage/ huge files and incorrect files as observed in #1323.

### DIFF
--- a/src/io/parquet/write/binary/nested.rs
+++ b/src/io/parquet/write/binary/nested.rs
@@ -4,7 +4,7 @@ use parquet2::{encoding::Encoding, page::DataPage};
 use super::super::{nested, utils, WriteOptions};
 use super::basic::{build_statistics, encode_plain};
 use crate::io::parquet::read::schema::is_nullable;
-use crate::io::parquet::write::Nested;
+use crate::io::parquet::write::{slice_nested_leaf, Nested};
 use crate::{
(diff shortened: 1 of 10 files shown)


ritchie46:
O(n) is restored: 
image

jorgecarleitao:
Oof, thank you so much, @ritchie46 !

ritchie46:
@jorgecarleitao This should be good to go. The coverage failure is unrelated.

PR #1330, opened 2022-12-13
fix invalid parquet read

### LINKED ISSUE #1329: Regression parquet read Utf8 column.
See issue downstream: https://github.com/pola-rs/polars/issues/5795

I did a bisect and it was introduced in #1316. Decoding we hit is `OptionalDictionary`.

ritchie46:
fixes #1329 

The length of the `offsets` container return `n - 1`. This is unintuitive and lead to a bug where the `extend_constant` added a single value too little. I made the distinction between `len` (the real number of indexes) and `len_proxy` the number of values an array with this many offsets would have.


jorgecarleitao:
Agree that it is a bit confusing - I was also not sure about what to use. This addresses it indeed ^^

PR #1334, opened 2022-12-17
Fix nested extension types

jondo2010:
This PR fixes various issues with Extension types at various nesting levels. Rebased onto `main`.

### DIFF
--- a/src/array/growable/structure.rs
+++ b/src/array/growable/structure.rs
@@ -3,7 +3,6 @@ use std::sync::Arc;
 use crate::{
     array::{Array, StructArray},
     bitmap::MutableBitmap,
-    datatypes::DataType,
 };
 
 use super::{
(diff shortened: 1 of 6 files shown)


src/array/list/mod.rs
```diff
@@ -54,7 +54,7 @@ impl<O: Offset> ListArray<O> {
 
         let child_data_type = Self::try_get_child(&data_type)?.data_type();
         let values_data_type = values.data_type();
-        if child_data_type != values_data_type {
+        if child_data_type.to_logical_type() != values_data_type.to_logical_type() {
```
jorgecarleitao, line 57: don't we want this to remain? I.e. if the datatype is `List<Extension(Int32)>` but the array is `List<Int32>`, we want to error, right?

jorgecarleitao (approved):
Thank you for this PR! I have a minor comment, but otherwise we can merge it.

src/array/list/mod.rs
jondo2010: Yep, I missed the actual root-cause of the problem, which was in `GrowableStruct`.

jondo2010:
Seems like a codecov submission failure. @jorgecarleitao, can you re-trigger the build on this please?

PR #1337, opened 2022-12-18
Reduced re-alloc in parquet

### LINKED ISSUE #1324: Arrow2 read parquet file did not reuse the page decoder buffer to array
Let's look at these codes:

```
  State::Required(page) => {
                values.extend(
                    page.values
                        .by_ref()
                        .map(decode)
                        .map(self.op)
                        .take(remaining),
                );
            }
```

It had extra `memcpy` in `values.extend` and decode`, I think maybe we could optimize it by using Buffer clone.

The first motivation is to move 
```
#[derive(Debug, Clone)]
pub struct DataPage {
    pub(super) header: DataPageHeader,
    pub(super) buffer: Vec<u8>,
    ...
}
```

to

```
#[derive(Debug, Clone)]
pub struct DataPage {
    pub(super) header: DataPageHeader,
    pub(super) buffer: Buffer<u8>,
    ...
}
```

@jorgecarleitao what do you think about this?

I found arrow-rs had addressed this improvement in https://github.com/apache/arrow-rs/blob/master/parquet/src/arrow/array_reader/byte_array.rs#L115-L138

jorgecarleitao:
Closes #1324

### DIFF
--- a/src/io/parquet/read/deserialize/binary/basic.rs
+++ b/src/io/parquet/read/deserialize/binary/basic.rs
@@ -339,11 +339,37 @@ impl<'a, O: Offset> utils::Decoder<'a> for BinaryDecoder<O> {
         }
     }
 
-    fn with_capacity(&self, capacity: usize) -> Self::DecodedState {
-        (
-            Binary::<O>::with_capacity(capacity),
-            MutableBitmap::with_capacity(capacity),
-        )
+    fn with_capacity(&self, capacity: usize, page: &Self::State) -> Self::DecodedState {
+        match page {
+            State::Optional(_, values) => (
+                // this is an upper bound - we may not consume the whole page.
+                // we do not know how many nulls are there, so we do not know how many
+                // valid items are there to discount over the sequence.
+                Binary::<O>::with_capacity(capacity, values.num_bytes()),
+                MutableBitmap::with_capacity(capacity),
+            ),
+            State::Required(values) => (
+                // this is an upper bound - we may not consume the whole page.
+                Binary::<O>::with_capacity(capacity, values.values.num_bytes()),
+                MutableBitmap::new(),
+            ),
+            State::FilteredRequiredDictionary(_)
+            | State::FilteredRequired(_)
+            | State::RequiredDictionary(_)
+            | State::Delta(_)
+            | State::FilteredDelta(_) => (
+                Binary::<O>::with_capacity(capacity, 0),
+                MutableBitmap::new(),
+            ),
+            State::FilteredOptionalDictionary(_, _)
+            | State::FilteredOptional(_, _)
+            | State::OptionalDictionary(_, _)
+            | State::OptionalDelta(_, _)
+            | State::FilteredOptionalDelta(_, _) => (
+                Binary::<O>::with_capacity(capacity, 0),
+                MutableBitmap::with_capacity(capacity),
+            ),
+        }
     }
 
     fn extend_from_state(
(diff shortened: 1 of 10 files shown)


ritchie46:
I noticed a significant performance regression on reading utf8 data. I think it is related to now defaulting to a `values_capacity` of `0` instead of `24 * capacity` as a reasonable default.

src/io/parquet/read/deserialize/binary/basic.rs
```diff
@@ -339,11 +339,37 @@ impl<'a, O: Offset> utils::Decoder<'a> for BinaryDecoder<O> {
...
+            | State::FilteredDelta(_) => (
+                Binary::<O>::with_capacity(capacity, 0),
+                MutableBitmap::new(),
+            ),
+            State::FilteredOptionalDictionary(_, _)
+            | State::FilteredOptional(_, _)
+            | State::OptionalDictionary(_, _)
+            | State::OptionalDelta(_, _)
+            | State::FilteredOptionalDelta(_, _) => (
+                Binary::<O>::with_capacity(capacity, 0),
```
ritchie46, line 369: This value capacity of `0` is very costly in most cases.

src/io/parquet/read/deserialize/binary/basic.rs
```diff
@@ -339,11 +339,37 @@ impl<'a, O: Offset> utils::Decoder<'a> for BinaryDecoder<O> {
...
+                // this is an upper bound - we may not consume the whole page.
+                Binary::<O>::with_capacity(capacity, values.values.num_bytes()),
+                MutableBitmap::new(),
+            ),
+            State::FilteredRequiredDictionary(_)
+            | State::FilteredRequired(_)
+            | State::RequiredDictionary(_)
+            | State::Delta(_)
+            | State::FilteredDelta(_) => (
+                Binary::<O>::with_capacity(capacity, 0),
```
ritchie46, line 361: This value capacity of `0` is very costly in most cases.

src/io/parquet/read/deserialize/binary/dictionary.rs
```diff
@@ -59,7 +59,7 @@ fn read_dict<O: Offset>(data_type: DataType, dict: &DictPage) -> Box<dyn Array>
 
     let values = SizedBinaryIter::new(&dict.buffer, dict.num_values);
 
-    let mut data = Binary::<O>::with_capacity(dict.num_values);
+    let mut data = Binary::<O>::with_capacity(dict.num_values, 0);
```
ritchie46, line 62: This value capacity of `0` is very costly in most cases.

src/io/parquet/read/deserialize/binary/utils.rs
```diff
@@ -36,10 +36,10 @@ impl<O: Offset> Pushable<usize> for Offsets<O> {
 
 impl<O: Offset> Binary<O> {
     #[inline]
-    pub fn with_capacity(capacity: usize) -> Self {
+    pub fn with_capacity(capacity: usize, values_capacity: usize) -> Self {
```
ritchie46, line 39: Maybe we should have this signature:

```rust
pub fn with_capacity(capacity: usize, values_capacity: Option<usize>) -> Self {
```

and then 

```rust
let values_capacity = values_capacity.unwrap_or(capacity.min(100) * 24);
```

sundy-li: If the values_capacity is zero by default, maybe we could use a `need_estimated` to reserve the space.

We use this in [databend](https://github.com/datafuselabs/databend/blob/8be55847084afb38f9c92d28ee10eb2f52bed90c/src/query/expression/src/types/string.rs#L318-L328)

jorgecarleitao:
Sorry for the delay on this one - if anyone would like to take an extra pass go ahead, otherwise I will merge it in.

sundy-li (approved):

PR #1342, opened 2022-12-22
Fixed csv infer_schema on empty fields

tripokey:
The default `infer` function provided by `src/io/csv/utils.rs` infers an empty field as `DataType::Utf8`. If the same column contains non empty fields containing valid `DataType::Float64` data the column in `infer_schema` will then contain both DataTypes. When `merge_schema` is called it will decide that `DataType::Utf8` has precedence over `DataType::Float64` and set the column type to `DataType::Float64`.

If we instead decline to infer anything for a field without any data we will in the end get `DataType::Float64` if the column contained other fields with valid f64 data and if we only had empty data for the column it will still default to `DataType::Utf8`.

### DIFF
--- a/src/io/csv/read/infer_schema.rs
+++ b/src/io/csv/read/infer_schema.rs
@@ -46,7 +46,9 @@ pub fn infer_schema<R: Read + Seek, F: Fn(&[u8]) -> DataType>(
 
         for (i, column) in column_types.iter_mut().enumerate() {
             if let Some(string) = record.get(i) {
-                column.insert(infer(string));
+                if !string.is_empty() {
+                    column.insert(infer(string));
+                }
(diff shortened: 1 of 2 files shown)


jorgecarleitao:
Thanks a lot for the PR! Could you add a test in `tests/it/io/csv` demoing this, so we get a regression test?

tripokey:
Sure 🙂, I'm on vacation now but I will look into it when I'm back.

tripokey:
I added the test `infer_ints_with_empty_fields()` which verifies that type inference still works when there is an empty field present.

I apologize for the delay...

jorgecarleitao:
Thanks a lot, merged!

PR #1345, opened 2022-12-24
Dependencies update

aldanor:
There's a few dependencies that are quite outdated (including some hefty ones, like zstd). The only one left outdated is odbc-api since I'm unsure about that.


jorgecarleitao:
This seems to stop the CI from working for wasm32. Any idea why?

aldanor:
@jorgecarleitao I think I know and I'll fix - but could you please force-rerun the CI jobs first? (I don't have permissions)

PR #1349, opened 2023-01-04
add toolchain.toml

ritchie46:
This ensures that everyone uses the same rustc as we have defined in CI.


jorgecarleitao:
Good idea! Could you rebase against #1353 ? I fixed the clippy issues raised after this ^^

ritchie46:
Ah, you were faster than me. :)

PR #1350, opened 2023-01-04
allow empty structs

ritchie46:
Empty structs are very useful as they map directly to empty objects/ dictionaries in host languages. See: https://github.com/pola-rs/polars/issues/6001

I believe the specification allows for empty structs as pyarrow also is able to generate them:

```python
>>> import pyarrow as pa
>>> pa.struct([])
StructType(struct<>)
```


ritchie46:
Will think this further first.

PR #1351, opened 2023-01-04
Fix incorrect downcast in `estimated_size_bytes`

jleibs:
Looks like this might have been a copy-and-paste error?

Added a unit-test for fixed_size_list as well.


ritchie46:
Thanks you @jleibs :+1:

jorgecarleitao:
Thanks a lot, @jleibs !

Issue #1359, opened 2023-01-14
growable::union is broken for dense unions


jleibs:
The derived array ends up repeating values from the first union in every extension.

Minimal unit test demonstrating the problem here: https://github.com/rerun-io/arrow2/blob/b082fd9b9bdd2a54ed68a6dc5570e457e6c24b00/tests/it/array/growable/union.rs#L73
```
---- array::growable::union::complex_dense stdout ----
thread 'array::growable::union::complex_dense' panicked at 'assertion failed: `(left == right)`
  left: `UnionArray[1, 2, c, d, e, 6]`,
 right: `UnionArray[1, 2, c, c, d, 1]`', tests/it/array/growable/union.rs:116:5
```

The crux of the problem occurs here:
 - https://github.com/jorgecarleitao/arrow2/blob/211be21991c29c235253a19413f43a76a367955f/src/array/growable/union.rs#L76
 
 The new offsets that are being extended here refer to positions in the original values-array, but need to account for the fact that these will be in a newly-extended values array and need to be adjusted to reflect their new position accordingly.
 
 I'll put up a PR to resolve this in the next day or two, but wanted an issue to refer to when I submit that PR.

PR #1360, opened 2023-01-15
Correctly update child-offsets in `GrowableUnion`

jleibs:
This resolves: https://github.com/jorgecarleitao/arrow2/issues/1359

As mentioned in the ticket, union offsets weren't being correctly updated to reflect positions in the new child fields. This adds a new method to the `Growable` interface: `len()`.  I believe I've got the right interpretation for the different growable implementations but would very much appreciate a sanity check there.

### DIFF
--- a/src/array/growable/mod.rs
+++ b/src/array/growable/mod.rs
@@ -43,6 +43,10 @@ pub trait Growable<'a> {
     /// Extends this [`Growable`] with null elements, disregarding the bound arrays
     fn extend_validity(&mut self, additional: usize);
 
+    /// Get the offset of the next value to be added to this [`Growable`].
+    /// Used to remap offsets when building up a [`GrowableUnion`].
+    fn next_offset(&self) -> usize;
+
     /// Converts this [`Growable`] to an [`Arc<dyn Array>`], thereby finishing the mutation.
     /// Self will be empty after such operation.
     fn as_arc(&mut self) -> Arc<dyn Array> {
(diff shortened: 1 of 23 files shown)


src/array/growable/mod.rs
```diff
@@ -43,6 +43,10 @@ pub trait Growable<'a> {
     /// Extends this [`Growable`] with null elements, disregarding the bound arrays
     fn extend_validity(&mut self, additional: usize);
 
+    /// Get the offset of the next value to be added to this [`Growable`].
+    /// Used to remap offsets when building up a [`GrowableUnion`].
+    fn next_offset(&self) -> usize;
```
jorgecarleitao, line 48: I wonder - isn't this always equal to the logical `length` of the current stuff being built? If yes, maybe we could rename it to `len` (and therefore make it so that it is not really specific to build a `GrowableUnion` but useful to more use-cases?)

jorgecarleitao (approved):
I went through this and makes a lot of sense. Thanks a lot for the PR (and for the well framed issue!)

I left a minor suggestion, but really good regardless!

src/array/growable/mod.rs
jleibs: I think so -- I was less confident that was going to be the fact when I started the implementation, but agree that seems to be the case at this point. Will rename to `len()`

jleibs: The added benefit of naming this length is logically testing it in the other unit-tests independently of union-specific offsets so we can get code-coverage on it without needing to make unions of each sub-type.

PR #1367, opened 2023-01-18
Added `Utf8Array::apply_validity`

Arty-Maly:
I added an `apply_validity` function to the Utf8Array. Basically I just copied the functionality from PrimitiveArray. If there is a better way of doing this Im open to suggestions. 

I also was not sure where to put the test for it, so I added it in the same file under a test module.

### DIFF
--- a/src/array/utf8/mod.rs
+++ b/src/array/utf8/mod.rs
@@ -515,6 +515,17 @@ impl<O: Offset> Utf8Array<O> {
     {
         MutableUtf8Array::<O>::try_from_trusted_len_iter(iter).map(|x| x.into())
     }
+
+    /// Applies a function `f` to the validity of this array.
+    ///
+    /// This is an API to leverage clone-on-write
+    /// # Panics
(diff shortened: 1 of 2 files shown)


Arty-Maly:
@jorgecarleitao Let me know if I can improve anything.

src/array/utf8/mod.rs
```diff
@@ -579,3 +590,25 @@ impl<O: Offset> Default for Utf8Array<O> {
         Utf8Array::new(data_type, Default::default(), Default::default(), None)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_apply_validity() {
```
jorgecarleitao, line 599: The tests are in here: https://github.com/jorgecarleitao/arrow2/blob/main/tests/it/array/utf8/mod.rs, so that we also test that the functions are available outside the crate (and to avoid re-compiling the crate whenever we play with tests). Could you move this one there?

Arty-Maly: Yep, done

PR #1369, opened 2023-01-20
feat: allow memory mapping a slice into a primitive array.

ritchie46:
I often need to an arrow kernel without owning the data. This allows you to create a `PrimitiveArray` of the slice data, use the needed kernel and drop the array immediately all while the slice is in scope.

## Example usage in the wild

```rust
    pub unsafe fn take_unchecked_from_slice(
        &self,
        idx: &[IdxSize],
    ) -> PolarsResult<Series> {
        let idx = IdxCa::borrowed_from_slice("", idx);
        self.take_unchecked(&idx)
    }
```

### DIFF
--- a/src/array/primitive/mod.rs
+++ b/src/array/primitive/mod.rs
@@ -19,7 +19,6 @@ mod from_natural;
 mod iterator;
 pub use iterator::*;
 mod mutable;
-use crate::ffi::mmap::mmap_slice;
 pub use mutable::*;
 
 /// A [`PrimitiveArray`] is Arrow's semantically equivalent of an immutable `Vec<Option<T>>` where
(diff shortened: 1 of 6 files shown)


jorgecarleitao:
Interesting idea! What do you think of a slightly different API: https://github.com/jorgecarleitao/arrow2/compare/added_mmap_slice?expand=1

Essentially:
* Not expose it in PrimitiveArray, but on `arrow2::ffi::mmap::slice`, to make it obvious that we are mmapping
* Avoid the extra clone by calling `try_from_ffi` directly
* Avoid the transmute explicitely via `cast_slice`
* Re-organize the modules so that the IPC-specific mmap is in another module

ritchie46:
Oh,  I didn't see you already had a mockup PR. :see_no_evil: 

Let me know what you want to do. I can rebase to previous commit and merge yours?

jorgecarleitao:
> Oh, I didn't see you already had a mockup PR. 🙈
> 
> Let me know what you want to do. I can rebase to previous commit and merge yours?

I mocked after looking at yours - was trying to understand the gist of this PR and modifying it as I was reading it. ^^

I think probably merging mine on this one is easiest, but I have no preference.

ritchie46:
Alright, I just PRed your mockup, I will close this one.

PR #1374, opened 2023-01-26
Added `set_len` method to Buffer

haixuanTao:
Adding `set_len` enables to change the length value of the buffer.

This is required if `get_mut` changes the length of the buffer.

I think it will be great to add generic vector method such as `extend_from_slice`, `copy_from_slice`, with the according length change. I can add it to this pull request if interested.

### DIFF
--- a/src/buffer/immutable.rs
+++ b/src/buffer/immutable.rs
@@ -178,6 +178,13 @@ impl<T> Buffer<T> {
         self.offset
     }
 
+    /// # Safety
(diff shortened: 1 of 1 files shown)


jorgecarleitao:
Thank you for the PR! 🙇  Indeed this is needed in some cases.

jorgecarleitao:
Note that `Buffer` is primarily an immutable container (think `Arc<Vec<>>`), this is why we do not expose mutable methods to it (other than returning the mutable vector ^^)

PR #1395, opened 2023-02-12
Fixed broken guide link

kjschiroo:
It looks like there is a typo in the csv read documentation causing the existing link on the README.md page to result in a 404. Rather than change the README.md link from `csv_read` to `csv_reader` this PR changes the filename from `csv_reader` to `csv_read` and updates the only other existing reference to the file. This path was taken to make the filename align with the wider pattern within the guide while also addressing the broken link.


jorgecarleitao:
Thank you @kjschiroo ! 🙇

PR #1406, opened 2023-02-14
Added apply_validity and set_validity to mutable utf8 array

Arty-Maly:
This is similar to my previous pr for `Utf8Array` 

I am adding two methods: 
`set_validity` and `apply_validity` for `MutableUtf8Array`

This was pretty straightforward to add so I preemptively created a pr without double checking if it is something that would be accepted.

@jorgecarleitao Let me know if this looks good.


jorgecarleitao:
Looks perfect. Thank you!

PR #1414, opened 2023-02-22
Fixed statistics writing flag and correct null_count in dictionaries

ritchie46:
This ensures the parquet writer respects the `write_statistics==false` option. It also corrects the `null_count`. The reported null count was that of the dictionary values instead of the dictionary array.

### DIFF
--- a/tests/it/io/parquet/mod.rs
+++ b/tests/it/io/parquet/mod.rs
@@ -591,7 +591,7 @@ pub fn pyarrow_nullable_statistics(column: &str) -> Statistics {
 
             Statistics {
                 distinct_count: UInt64Array::from([None]).boxed(),
-                null_count: UInt64Array::from([Some(0)]).boxed(),
+                null_count: UInt64Array::from([Some(1)]).boxed(),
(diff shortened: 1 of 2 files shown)


ritchie46:
@jorgecarleitao can I easily add a test for this in the form of:

```rust
#[test]
fn int64_required_v1() -> Result<()> {
    round_trip(
        "int64",
        "required",
        Version::V1,
        CompressionOptions::Uncompressed,
        vec![Encoding::Plain],
    )
}
```

tests/it/io/parquet/mod.rs
```diff
@@ -591,7 +591,7 @@ pub fn pyarrow_nullable_statistics(column: &str) -> Statistics {
 
             Statistics {
                 distinct_count: UInt64Array::from([None]).boxed(),
-                null_count: UInt64Array::from([Some(0)]).boxed(),
+                null_count: UInt64Array::from([Some(1)]).boxed(),
```
ritchie46, line 594: This test result was invalid. I debugged and clearly found null_count of 1.

PR #1419, opened 2023-02-24
Added support for MapArray read and write to parquet

b41sh:
implement Map data type read and write for parquet


jorgecarleitao (commented):
Thanks for this, looks great! Could you add a test (e.g. roundtrip)?

sundy-li (approved):

PR #1435, opened 2023-03-16
Added impl_mutable_array_mut_validity macro for mutable arrays

Arty-Maly:
I am adding `set_validity`, `apply_validity`, and `with_validity` to mutable binary arrays. 

I am using a macro so it can be reused for other mutable arrays.

The macro for immutable arrays is called: `impl_mut_validity` so I tried sticking with that convention and called it `impl_mutable_array_mut_validity` But if there is a better name to describe this macro I will take it.

### DIFF
--- a/src/array/binary/mutable.rs
+++ b/src/array/binary/mutable.rs
@@ -160,6 +160,8 @@ impl<O: Offset> MutableBinaryArray<O> {
             validity.shrink_to_fit()
         }
     }
+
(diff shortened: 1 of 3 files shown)


Arty-Maly:
@jorgecarleitao I seem to be not able to figure out this IPC test failure, I even reverted my changes and ran it and it still failed. Can you give me some direction to debug this?

PR #1436, opened 2023-03-17
Added memmap `&[u8]` as `BooleanArray`

ritchie46:
Same a memory mapping a slice returns a `PrimitiveArray` with a lifetime that is tight to the slice, this adds the functionality for a `BooleanArray`


ritchie46:
I don't understand the integration test failure. It seems unrelated.

Issue #1429, opened 2023-03-07
[Proposal] Combination of arrow-rs and arrow2, deprecation of arrow2 repository


alamb:
This proposal is a response to the seeming desire to unify Rust Arrow development communities going  forward and my summary of the discussion on https://github.com/apache/arrow-rs/issues/1176 (specifically https://github.com/apache/arrow-rs/issues/1176#issuecomment-1430883886) . There is a summary with some diagrams that may be helpful in  https://docs.google.com/presentation/d/1cqQEpC-kJES2Mng152r_qZyaOqHjtb5YFuseSTWyulU/edit

# Current state
1. Both arrow-rs and arrow2 offer Rust implementations of apache Arrow
2. arrow-rs is under active development, is governed by the Apache Software Foundation (ASF), and has adequate resources for maintenance
3. arrow2 state is (maybe?) struggling to find sufficient maintenance capacity
4. It is challenging for users to build applications that use mix and match libraries that use arrow-rs (e.g. [datafusion](https://docs.rs/datafusion/latest/datafusion/index.html)) and arrow2 (e.g. [pola.rs](http://pola.rs/)). 

# Proposal 1 End State (alternate options are listed below)
1. arrow-rs continues to be maintained and released under the ASF guidelines
5. The arrow2 codebase is moved into the existing apache repo https://github.com/apache/arrow-experimental-rs-arrow2 and deprecated after some period (e.g. 6 months)
6. All relevant / needed functionality is ported from arrow2 to arrow-rs during the deprecation period

*This will necessitate users of arrow2 to either change their code, or find alternate maintainers with sufficient capacity for arrow2 if they want a maintained version of arrow going forward.*

## Prior to the Deprecation Period
1. The arrow2 code goes through the ASF IP clearance process
2. arrow2 code is brought into the existing experimental repo (https://github.com/apache/arrow-experimental-rs-arrow2) with a deprecation note, https://github.com/jorgecarleitao/arrow2 is marked read only, directing users to the experimental repo.
3. During the deprecation period, the community (whoever needs) continue to maintain the arrow-experimental-rs-arrow2 repo, we can release non-official ASF releases, etc

## After the Depreciation Period
1. https://github.com/apache/arrow-experimental-rs-arrow2 is marked as read only

# Proposed Milestones

## Milestone 1: Ergonomic (and Zero runtime cost) conversion
[...]

PR #1437, opened 2023-03-17
Added buffer interoperability with arrow-rs (#1429)

tustvold:
As part of #1429 we want to provide an interoperability story between arrow2 and arrow-rs.

The original [proposal](https://docs.google.com/presentation/d/1cqQEpC-kJES2Mng152r_qZyaOqHjtb5YFuseSTWyulU/edit) involved porting arrow-rs and arrow2 to have a common base array representation. This was to preserve the original spirit of @jorgecarleitao 's proposal in https://github.com/apache/arrow-rs/issues/1176#issuecomment-1430883886. However, doing this in an incremental fashion whilst not introducing performance regressions or major breaking changes is complicated and extremely time consuming.

Taking a step-back, all we really want is a reasonably fast way to convert between array representations, to facilitate interoperability and potentially incremental migration of codebases. Whilst perhaps less "pure", simply providing a safe API to convert between `ArrayData` and `Box<dyn arrow2::Array>` is likely sufficient.

The major things this would change are:

* Conversion is no longer just a memmove, but likely on a similar order of magnitude
* The converted arrays would not be convertible back to `Vec` as they would be opaque allocations

However, it would allow us to provide an interoperability story in a matter of days instead of weeks/months.

In this vein, this PR adds zero-copy conversion between the buffer representations, as this is all that is really necessary to permit this. The rest of the conversion logic is fairly mechanical, I already have it mostly implemented but wanted to get feedback first.

### DIFF
--- a/src/buffer/mod.rs
+++ b/src/buffer/mod.rs
@@ -5,7 +5,42 @@ mod iterator;
 
 use crate::ffi::InternalArrowArray;
 
-pub(crate) type Bytes<T> = foreign_vec::ForeignVec<InternalArrowArray, T>;
+pub(crate) enum BytesAllocator {
+    InternalArrowArray(InternalArrowArray),
+
+    #[cfg(feature = "arrow")]
+    Arrow(arrow_buffer::Buffer),
+}
+
+pub(crate) type Bytes<T> = foreign_vec::ForeignVec<BytesAllocator, T>;
+
+#[cfg(feature = "arrow")]
+pub(crate) fn to_buffer<T: crate::types::NativeType>(
+    value: std::sync::Arc<Bytes<T>>,
+) -> arrow_buffer::Buffer {
+    // This should never panic as ForeignVec pointer must be non-null
+    let ptr = std::ptr::NonNull::new(value.as_ptr() as _).unwrap();
+    let len = value.len() * std::mem::size_of::<T>();
+    // Safety: allocation is guaranteed to be valid for `len` bytes
+    unsafe { arrow_buffer::Buffer::from_custom_allocation(ptr, len, value) }
+}
+
+#[cfg(feature = "arrow")]
+pub(crate) fn to_bytes<T: crate::types::NativeType>(value: arrow_buffer::Buffer) -> Bytes<T> {
+    let ptr = value.as_ptr();
+    let align = ptr.align_offset(std::mem::align_of::<T>());
+    assert_eq!(align, 0, "not aligned");
+    let len = value.len() / std::mem::size_of::<T>();
+
+    // Valid as `NativeType: Pod` and checked alignment above
+    let ptr = value.as_ptr() as *const T;
+
+    let owner = crate::buffer::BytesAllocator::Arrow(value);
+
+    // Safety: slice is valid for len elements of T
+    unsafe { Bytes::from_foreign(ptr, len, owner) }
+}
+
 pub(super) use iterator::IntoIter;
 
 pub use immutable::Buffer;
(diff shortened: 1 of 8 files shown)


src/buffer/mod.rs
```diff
@@ -5,7 +5,42 @@ mod iterator;
...
-pub(crate) type Bytes<T> = foreign_vec::ForeignVec<InternalArrowArray, T>;
+pub(crate) enum BytesAllocator {
+    InternalArrowArray(InternalArrowArray),
+
+    #[cfg(feature = "arrow")]
+    Arrow(arrow_buffer::Buffer),
+}
+
+pub(crate) type Bytes<T> = foreign_vec::ForeignVec<BytesAllocator, T>;
+
```
tustvold, line 16: This is the real meat of the conversion

src/types/native.rs
```diff
@@ -14,6 +15,7 @@ pub trait NativeType:
     + Send
     + Sync
     + Sized
+    + RefUnwindSafe
```
tustvold, line 18: This is necessary to ensure that `arrow2::Bytes<T>` is `RefUnwindSafe` which is important to `arrow::Buffer`

tustvold:
> Is there anything about this API that would preclude an (eventual) unification of the underlying buffer types

No, although if this approach is given the green light, it is unclear that such a unification would be worth the fairly significant effort, I certainly would not be intending to undertake it.

tustvold:
Integration test failure does not appear to be related to this PR

ritchie46:
I do think it is a regression if we cannot get back to `Vec` anymore, In polars we convert back sometimes. Could we make this a feature gate? I could feature gate that behavior out in polars as well.


> Whilst perhaps less "pure", simply providing a safe API to convert between ArrayData and Box<dyn arrow2::Array> is likely sufficient.

Couldn't we already do this with arrow FFI spec? What are the pro's and cons against this route? As we would still need to compile both libraries if we convert between the two.

tustvold:
> if we cannot get back to Vec anymore

It's only you can't go back to Vec from an array created initially by the other library, i.e. the conversion loses this ability. Arrow-rs arrays created from vec can still be converted back, and the same for arrow2

> ffi

The conversion is safe and ergonomic, ffi is neither 😅

alamb:
> It's only you can't go back to Vec from an array created initially by the other library and then converted, i.e. the conversion loses the ability to go back to a vec. Arrow-rs arrays created from vec can still be converted back, and the same for arrow2

Maybe we can add a test demonstrating going back/forth to vec (and when it doesn't work) as a way to document the limitiation?

ritchie46:
> It's only you can't go back to Vec from an array created initially by the other library and then converted, i.e. the conversion loses the ability to go back to a vec. Arrow-rs arrays created from vec can still be converted back, and the same for arrow2

Right, I misunderstood that part. In that case this looks great! :+1:

alamb (approved):
This PR makes sense to me -- thank you @tustvold 

However, I am not super familiar with the arrow2 codebase so I defer to @ritchie46  / @jorgecarleitao  / @sundy-li for final approval

jorgecarleitao (approved):
I have reviewed and this is very ingenious and well implemented. Thank you @tustvold !

My only minor concern is that because arrow-buffer bumps major version every 2 weeks, we need to update this repo every 2 weeks, but this is only a procedural issue as the crate is not changing much.

Thank you again, @tustvold 🙇

alamb:
> My only minor concern is that because arrow-buffer bumps major version every 2 weeks, we need to update this repo every 2 weeks, but this is only a procedural issue as the crate is not changing much.


We might be able to publish new versions of arrow2 with minor (e.g. `0.16.1`) with just version updates if that turns out to be an issue. I think bumping dependents is perfectly semantically compatible

PR #1442, opened 2023-03-22
Added interoperability with arrow-schema

tustvold:
Part of https://github.com/jorgecarleitao/arrow2/issues/1429

Adds conversions to/from `arrow-schema`

### DIFF
--- a/src/datatypes/mod.rs
+++ b/src/datatypes/mod.rs
@@ -160,6 +160,126 @@ pub enum DataType {
     Extension(String, Box<DataType>, Option<String>),
 }
 
+#[cfg(feature = "arrow")]
+impl From<DataType> for arrow_schema::DataType {
+    fn from(value: DataType) -> Self {
+        match value {
+            DataType::Null => Self::Null,
+            DataType::Boolean => Self::Boolean,
+            DataType::Int8 => Self::Int8,
+            DataType::Int16 => Self::Int16,
+            DataType::Int32 => Self::Int32,
+            DataType::Int64 => Self::Int64,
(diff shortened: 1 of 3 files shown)


src/datatypes/mod.rs
```diff
@@ -160,6 +160,126 @@ pub enum DataType {
...
+                Self::Union(fields, ids, mode.into())
+            }
+            DataType::Map(f, ordered) => Self::Map(Box::new((*f).into()), ordered),
+            DataType::Dictionary(key, value, _) => Self::Dictionary(
+                Box::new(DataType::from(key).into()),
+                Box::new((*value).into()),
+            ),
+            DataType::Decimal(precision, scale) => Self::Decimal128(precision as _, scale as _),
+            DataType::Decimal256(precision, scale) => Self::Decimal256(precision as _, scale as _),
+            DataType::Extension(_, d, _) => (*d).into(),
```
tustvold, line 215: I'm curious why arrow2 chose to represent extension types as an explicit data type, as opposed to just field metadata?

jorgecarleitao: It is so that the `Array` has the logical type on it. It allows the user of a `Box<dyn Array>` to use `.data_type()` to perform the downcast and have the necessary information to build the extension.

For example, polars uses it to store arbitrary Python objects on the type. In theory this could be kept in the Field's metadata.

Pyarrow does the same: https://arrow.apache.org/docs/python/generated/pyarrow.ExtensionType.html

jorgecarleitao:
Thank you @tustvold

PR #1454, opened 2023-04-05
Fixed bug in mmap of empty FixedSizeBinaryArray

### LINKED ISSUE #1453: Possible bug in mmap of complex types
I'm using columns with a schema of List(Struct(Dictionary(...))) to represent
indexed database tables with compact keys.

I'm getting the error:

```
Error: OutOfSpec("buffer's length is too small in mmap")
```
When I use the mmap chunk reader, but the schema works fine with the regular reader.

I've managed to cut my failing example down to the following code, but I'll concede that
I may have cut too far!

This shows a working and failing example.

```
use arrow2::{datatypes::{Schema, DataType, Field, IntegerType}, array::{DictionaryArray, FixedSizeBinaryArray, Int32Array, ListArray, StructArray}, chunk::Chunk, offset::OffsetsBuffer};

type BDE = Box<dyn std::error::Error + Send + Sync + 'static>;

fn main() -> Result<(), BDE> {
    let result = vec![];
    let mut writer = arrow2::io::ipc::write::FileWriter::try_new(
        result,
        schema(),
        None,
        Default::default(),
    )?;

    let keys = Int32Array::new_empty(DataType::Int32);
    let values = FixedSizeBinaryArray::new_empty(DataType::FixedSizeBinary(20));

    let accounts_arrays = vec![
        DictionaryArray::<i32>::try_from_keys(keys, values.boxed())?.boxed(),
    ];

    let chunk = Chunk::try_new(vec![
        ListArray::try_new(
            DataType::List(Box::new(Field::new(
                "list",
                DataType::Struct(accounts_fields()),
                false,
            ))),
            OffsetsBuffer::<i32>::try_from(vec![0, 0])?,
            StructArray::try_new(
                DataType::Struct(accounts_fields()),
                accounts_arrays,
                None,
            )?
            .boxed(),
            None,
        )?
        .boxed(),
    ])?;

    writer.write(&chunk, None)?;
    writer.finish()?;
    let bytes = writer.into_inner();

    // we first read the files' metadata
    let mut reader = std::io::Cursor::new(&bytes);
    let metadata = arrow2::io::ipc::read::read_file_metadata(&mut reader)?;

    if true {
        let data = std::sync::Arc::new(bytes);

        let dictionaries =
        unsafe { arrow2::mmap::mmap_dictionaries_unchecked(&metadata, data.clone())? };
    
        let _fails = unsafe { arrow2::mmap::mmap_unchecked(&metadata, &dictionaries, data, 0)? };
    } else {
        let mut reader = arrow2::io::ipc::read::FileReader::new(reader, metadata, None, None);
        let _works = reader.next().ok_or("no chunks")??;
    }
    Ok(())
}

fn schema() -> Schema {
    use DataType::*;
    Schema::from(vec![
        Field::new(
            "accounts",
            List(Box::new(Field::new(
                "list",
                Struct(accounts_fields()),
                false,
            ))),
            false,
        ),
    ])
}

fn accounts_fields() -> Vec<Field> {
    use DataType::*;
    use IntegerType::Int32;
    vec![
        // Address: 32 bits
        Field::new(
            "address",
            Dictionary(Int32, Box::new(FixedSizeBinary(20)), false),
            false,
        ),
    ]
}
```

andy-thomason:
This PR fixes #1453 

When mmap'ing a `FixedSizeBinaryArray`, the length in bytes was miscalculated.

This was only used for a check, but to avoid UB, this should be rigourous.

The problem manifested as an `OutOfSpec` error mapping zero length `FixedSizeBinaryArray`

This fix uses the `DataType` to determine the row length.

### DIFF
--- a/src/io/ipc/read/error.rs
+++ b/src/io/ipc/read/error.rs
@@ -95,7 +95,7 @@ pub enum OutOfSpecKind {
         /// The requested dictionary id
         requested_id: i64,
     },
-    /// FixedLengthBinaryArray has invalid datatype.
+    /// FixedSizeBinaryArray has invalid datatype.
     InvalidDataType,
 }
 
--- a/src/io/ipc/read/error.rs
+++ b/src/io/ipc/read/error.rs
@@ -95,6 +95,8 @@ pub enum OutOfSpecKind {
         /// The requested dictionary id
         requested_id: i64,
     },
+    /// FixedLengthBinaryArray has invalid datatype.
+    InvalidDataType,
 }
 
 impl From<OutOfSpecKind> for Error {
--- a/src/mmap/array.rs
+++ b/src/mmap/array.rs
@@ -141,7 +141,8 @@ fn mmap_fixed_size_binary<T: AsRef<[u8]>>(
     let data_ref = data.as_ref().as_ref();
 
     let validity = get_validity(data_ref, block_offset, buffers, null_count)?.map(|x| x.as_ptr());
-    let values = get_buffer::<u8>(data_ref, block_offset, buffers, num_rows * bytes_per_row)?.as_ptr();
+    let values =
+        get_buffer::<u8>(data_ref, block_offset, buffers, num_rows * bytes_per_row)?.as_ptr();
 
     Ok(unsafe {
         create_array(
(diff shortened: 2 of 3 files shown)


jorgecarleitao:
Thank you @andy-thomason for this PR and well spotted! 🙇

PR #1460, opened 2023-04-10
chore: bump arrow-rs

sundy-li:


alamb:
Thank you @sundy-li 

The integration IPC test failure on this PR https://github.com/jorgecarleitao/arrow2/actions/runs/4528720649/jobs/7975761365


Appears to also be failing on the main branch (aka not caused by this PR)


Most recent CI run on master seems to fail the same way
https://github.com/jorgecarleitao/arrow2/actions/runs/4528720649/jobs/7975761365

jorgecarleitao:
Thank you! was this also bumped on #1446 by @tustvold and thus this PR is superseeded?

PR #1463, opened 2023-04-11
Close Underlying Writer when IPC `StreamSink` is Closed

### LINKED ISSUE #1462: IPC `StreamSink` Doesn't Propagate Close to the Underlying Writer
Hi,

I came across this issue while working with a writer that writes data to the network. When I call close on the IPC SinkStream, I discovered that the underlying writer wasn't closed, which prevented the underlying writer from flushing its data to the network and cleaning itself up. 

If I'm misunderstanding how to use the SinkStream, let me know, and I can correct my usage. If this is an actual issue, I'm happy to contribute a fix PR.

Thanks!

theadd336:
- Closes #1462
- Calls flush and close on the underlying writer when poll_close is called.

### DIFF
--- a/src/io/ipc/write/stream_async.rs
+++ b/src/io/ipc/write/stream_async.rs
@@ -170,6 +170,7 @@ where
                     this.task = Some(
                         async move {
                             write_continuation(&mut writer, 0).await?;
+                            writer.flush().await?;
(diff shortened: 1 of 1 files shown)


ritchie46:
Thanks @theadd336!

PR #1467, opened 2023-04-16
Fixed timestamp to datetime conversion for values with fractional seconds before the epoch

alexander-beedie:
The `timestamp_XX_to_datetime` functions all panic with _"invalid or out-of-range datetime"_ on conversions of values with fractional seconds before the epoch (where the timestamp becomes negative).

For these timestamps (pre-epoch, fractional secs) the  `(v % <scale_factor>) as u32` formulation is not valid, as
* the result is a _negative_ number
* ... which causes integer wraparound via `as u32`
* ... which then leads to a panic, as we can't init with billions of microseconds (for example).

This PR allows pre-epoch conversions with fractional seconds to be handled correctly, and adds some new tests to provide the missing coverage.


ritchie46:
Thanks @alexander-beedie. Great fix. :+1:

PR #1491, opened 2023-05-24
Implement into_inner for Utf8Array and BinaryArray for reusing Buffer allocations

hzuo:
Many kernels operate directly on the underlying buffer, creating a new allocation for a new buffer with updated values, and re-use the other components of the array like the offsets, bitmap, etc.

In many cases, the the output of the kernel is transient, and the allocation for the buffer can be reused - as we run the kernel over a big set of arrays, we can reuse the same scratch buffer (similar to the scratch buffers used in the io modules).

Exposing `into_inner` allows applications to re-use the scratch allocation instead of calling out to the allocator for every array.

`PrimitiveArray`s and `BooleanArray`s already have an `into_inner` which deconstruct themselves into their constituent components - here we just bring Utf8Arrays and BinaryArrays in line with the other array types.

### DIFF
--- a/src/array/binary/mod.rs
+++ b/src/array/binary/mod.rs
@@ -221,6 +221,17 @@ impl<O: Offset> BinaryArray<O> {
     impl_mut_validity!();
     impl_into_array!();
 
+    #[must_use]
+    pub fn into_inner(self) -> (DataType, OffsetsBuffer<O>, Buffer<u8>, Option<Bitmap>) {
+        let Self {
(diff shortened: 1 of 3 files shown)


sundy-li (approved):

sundy-li:
LGTM.
BTW, The `Buffer` is copy free, so it's free to call `array.values().clone()`

ritchie46:
Yeap, seems good. :+1:

PR #1494, opened 2023-05-30
chore: feature gate decimal arithmetic

ritchie46:


ritchie46:
failing test is unrelated. Is due to new chrono release. Will fix that in another PR.

PR #1498, opened 2023-06-01
Fix missing `UInt64` case for page filtering

poshcoe:
Previously when attempting to deserialize `UInt64` fields for page filtering, the `read_filtered_pages` function would incorrectly raise a `NotYetImplemented` error.

The cause was a missing match case for `PrimitiveType:UInt64`.


ritchie46:
Thanks!

PR #1499, opened 2023-06-02
Relax the checks for the presence of ArrowSchema.name as this field is optional

Qqwy:
As per the spec, the [ArrowSchema.name](https://arrow.apache.org/docs/format/CDataInterface.html#c.ArrowSchema.name) field is optional.
SQLite is an example of a producer which does not set this field.

This PR removes the checks on this field, returning an empty string if the field is NULL for backwards compatibility.

### DIFF
--- a/src/ffi/schema.rs
+++ b/src/ffi/schema.rs
@@ -154,15 +154,18 @@ impl ArrowSchema {
     }
 
     /// returns the name of this schema.
+    ///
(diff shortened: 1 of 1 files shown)


sundy-li (approved):

Qqwy:
@sundy-li The formatter CI pass should now also succeed.

PR #1500, opened 2023-06-02
Allows to use wrappers other than Box to build ArrowArrayStreamReader.

Qqwy:
Before, only Box<ArrowArrayStream> was supported by ArrowArrayStreamReader.

This PR enables support for ArrowArrayStreams which are:
- stack-allocated rather than boxed (plain `ArrowArrayStream`);
- not owned (`&mut ArrowArrayStream`);
- wrapped in a `Box<ArrowArrayStream, CustomAllocator>`;
- wrapped in a Box-like type other than `std::alloc::Box`;
- wrapped in an `Arc<Mutex<>>`
- etc.

This is especially important when we receive a `*mut ArrowArrayStream` from C (or anywhere else outside of Rust) as it will need to be cleaned up in the correct way, so a normal `Box` does not cut it.

### DIFF
--- a/src/ffi/stream.rs
+++ b/src/ffi/stream.rs
@@ -1,5 +1,5 @@
-use std::ops::DerefMut;
 use std::ffi::{CStr, CString};
+use std::ops::DerefMut;
 
 use crate::{array::Array, datatypes::Field, error::Error};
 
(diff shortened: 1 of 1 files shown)


sundy-li (approved):

Qqwy:
@sundy-li The formatter CI pass should now also succeed. :blush:

PR #1501, opened 2023-06-04
ArrowArrayStreamReader::try_new(): Safeguard against released streams

Qqwy:
Before, passing an already-released stream to `ArrowArrayStreamReader::try_new()` would result in a call to `.get_schema()` (and later on when the reader is used possibly to `.get_next()`), which goes against the spec, as [those callbacks should not be called on released streams](https://arrow.apache.org/docs/format/CStreamInterface.html#the-arrowarraystream-structure).

This PR adds an extra check to the start of `try_new()` to safeguard against already-released streams.


ritchie46 (approved):
Thanks you @Qqwy

Issue #1502, opened 2023-06-05
MutableListArray from an existing MutableArray


cjermain:
Is there a way to directly nest a non-empty `MutableArray` in a `MutableListArray`?

In the majority of `MutableArray` implementations there is a `from_trusted_len_iter` method that allows the array to be constructed directly from an iterator. In the case of a `MutableStructArray`, the `new` method takes a vector of the inner `MutableArray`s, so you can directly construct it. Is there an analogy for `MutableListArray`?

It looks like the `new_from` has an assert that prevents non-empty `MutableArray`s from being passed through:
https://github.com/jorgecarleitao/arrow2/blob/fb5e4d591c7149df590a330365fae55d2370962f/src/array/list/mutable.rs#L118

Along with the `MutableArray` I also have all of the lengths in a `Vec<Option<usize>>`. I had been hoping to use the following, but am failing on the assert during runtime:
```rust
let mut array = MutableListArray::<i32, _>::new_with_capacity(inner_array, lengths.len());
array.try_extend_from_lengths(lengths).unwrap();
```

I'm trying to accomplish the same thing in this function, without the pre-allocation of an empty inner `MutableArray` ahead of time:
https://github.com/jorgecarleitao/arrow2/blob/fb5e4d591c7149df590a330365fae55d2370962f/src/io/json/read/deserialize.rs#L254-L277

cjermain:
@jorgecarleitao, what do you think?

PR #1503, opened 2023-06-08
Adding new constructor for MutableListArray

cjermain:
This PR addresses #1502 by introducing a new constructor for `MutableListArray` that allows for building an object based on existing values from a `MutableArray`, `Offsets`, and `MutableBitmap`. I'm interested in feedback on this relative to the question posed in #1502, and happy to add tests if this seems like the right approach.

### DIFF
--- a/src/array/list/mutable.rs
+++ b/src/array/list/mutable.rs
@@ -149,6 +149,7 @@ impl<O: Offset, M: MutableArray> MutableListArray<O, M> {
         offsets: Offsets<O>,
         validity: Option<MutableBitmap>,
     ) -> Self {
+        assert_eq!(values.len(), offsets.last().to_usize());
         let data_type = values.data_type().clone();
         Self {
             data_type,
--- a/src/array/list/mutable.rs
+++ b/src/array/list/mutable.rs
@@ -142,6 +142,22 @@ impl<O: Offset, M: MutableArray> MutableListArray<O, M> {
         Self::new_from(values, data_type, capacity)
     }
 
+    /// Creates a new [`MutableListArray`] from a [`MutableArray`], [`Offsets`] and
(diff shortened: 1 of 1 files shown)


src/array/list/mutable.rs
```diff
@@ -142,6 +142,22 @@ impl<O: Offset, M: MutableArray> MutableListArray<O, M> {
...
     }
 
+    /// Creates a new [`MutableListArray`] from a [`MutableArray`], [`Offsets`] and
+    /// [`MutableBitmap`].
+    pub fn new_from_mutable(
+        values: M,
+        offsets: Offsets<O>,
+        validity: Option<MutableBitmap>,
+    ) -> Self {
+        let data_type = values.data_type().clone();
```
sundy-li, line 152: LGTM, but better  check that the `values.len()` equals the last element of offsets.

src/array/list/mutable.rs
```diff
@@ -142,6 +142,23 @@ impl<O: Offset, M: MutableArray> MutableListArray<O, M> {
...
 
+    /// Creates a new [`MutableListArray`] from a [`MutableArray`], [`Offsets`] and
+    /// [`MutableBitmap`].
+    pub fn new_from_mutable(
+        values: M,
+        offsets: Offsets<O>,
+        validity: Option<MutableBitmap>,
+    ) -> Self {
+        assert_eq!(values.len(), offsets.last().to_usize());
+        let data_type = values.data_type().clone();
```
cjermain, line 153: @sundy-li, thanks for the quick response. Unfortunately I missed on the dtype -- it should be the following:
```rust
        let data_type = ListArray::<O>::default_datatype(values.data_type().clone());
```

PR #1506, opened 2023-06-12
feat: add from_inner

ritchie46:
This adds a complement of `into_inner`, `from_inner`. Allowing us to get data IN and OUT of arrow arrays without destructors being run, cached null counts being recomputed or checks being done in hot loops.


src/buffer/mod.rs
```diff
@@ -4,15 +4,65 @@ mod immutable;
...
 pub(crate) enum BytesAllocator {
     InternalArrowArray(InternalArrowArray),
 
     #[cfg(feature = "arrow")]
     Arrow(arrow_buffer::Buffer),
 }
+pub(crate) type BytesInner<T> = foreign_vec::ForeignVec<BytesAllocator, T>;
 
-pub(crate) type Bytes<T> = foreign_vec::ForeignVec<BytesAllocator, T>;
+/// Bytes representation.
```
ritchie46, line 17: Makes the struct public, so that it can be passed as argument, but keeps internals hidden.

src/ffi/array.rs
```diff
@@ -258,10 +258,11 @@ unsafe fn create_bitmap(
 
     let len: usize = array.length.try_into().expect("length to fit in `usize`");
     let offset: usize = array.offset.try_into().expect("Offset to fit in `usize`");
+    let null_count: usize = array.null_count();
     let bytes_len = bytes_for(offset + len);
     let bytes = Bytes::from_foreign(ptr, bytes_len, BytesAllocator::InternalArrowArray(owner));
 
-    Ok(Bitmap::from_bytes(bytes, offset + len).sliced(offset, len))
```
ritchie46, line 264: This saves 2 bitcounts. First on initialization. Seconds on slice.

PR #1508, opened 2023-06-15
Convert from_type to parent logical type before attempting a cast.

GPSnoopy:
This is part of larger PR for `polars` https://github.com/pola-rs/polars/issues/9373.

Though it does seem to be the right thing when doing a cast and dealing with extension types.


GPSnoopy:
Closing PR as per https://github.com/pola-rs/polars/issues/9373#issuecomment-1602265718

PR #1510, opened 2023-06-21
Add datetime with timezone support to Json serializer

noselasd:
Hello,
We were trying to serialize data to Json via polars, but was hit with the missing support for DateTime types with timezone
support. I'm not too well versed in Rust, but I made an attempt at adding support for it.

It's inspired by how the CSV serializer does this. The Json serializer doesn't propagate errors back in `new_serializer()` as the CSV serializer does, I havn't attempted rewriting the interface, so this panics if invalid timezones are encontered. It should still be an improvement over the `todo!("still have to implement timezone")` in `src/io/json/write/serialize.rs`

`DateTime` serializing is done differently from serializing `NaiveDateTime`, which uses a space instead of a T as the date and time separator.
However the default chrono `to_string` on timezone aware `DateTIme` produces a space before the tz offset (e.g. 1996-12-19 16:39:57 +08:00) which Javascript engines won't parse as a Date(), so the RFC3339 format is used instead. `chrono-tz` timezones are also serialized with a fixed timezone offset format [+-hh:mm] so Javascrpt engines can make use of it.

I do have some future plans as well, including adding a options for custom time format string as is done in the CSV parser, some knobs to tune how `NaiveDateTime` is serialized, but this is perhaps a start.

### DIFF
--- a/src/io/json/write/serialize.rs
+++ b/src/io/json/write/serialize.rs
@@ -11,8 +11,9 @@ use crate::offset::Offset;
 use crate::temporal_conversions::parse_offset_tz;
 use crate::temporal_conversions::{
     date32_to_date, date64_to_date, duration_ms_to_duration, duration_ns_to_duration,
-    duration_s_to_duration, duration_us_to_duration, timestamp_ms_to_datetime,
-    timestamp_ns_to_datetime, timestamp_s_to_datetime, timestamp_us_to_datetime,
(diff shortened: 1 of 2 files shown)


ritchie46:
@MarcoGorelli could you take a look at this one?

ritchie46 (approved):

ritchie46:
Nevermind, looks good!

PR #1511, opened 2023-06-22
Fix list array parsing in pandas record json

AnIrishDuck:
The double-list construction was a mistake that prevents parsing of several valid types of pandas record json constructs.

Fix that by removing the nested lists and add a test verifying that it works.

### DIFF
--- a/src/io/json/read/deserialize.rs
+++ b/src/io/json/read/deserialize.rs
@@ -625,14 +625,11 @@ fn allocate_array(f: &Field) -> Box<dyn MutableArray> {
             f.data_type().clone(),
             *size,
         )),
-        DataType::List(inner) => match inner.data_type() {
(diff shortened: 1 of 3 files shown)


ritchie46:
Thanks for the PR @AnIrishDuck. What was the behavior we had first that was flawed? For my understanding of the changes here.

AnIrishDuck:
When trying to decode pandas-record json with list arrays, it would expect them to be nested within another list array. In other words, deserialization of `ListArray`s for pandas record JSON was pretty much generally broken.

We missed this on implementation because we were using `FixedSizeList` in most places.

AnIrishDuck:
To elaborate (after looking at the tests), we only tested that nested list arrays worked, which hid the underlying issue (as the outer level of nesting that `infer_schema` would generate would still be properly handled by `allocate_array`)

ritchie46:
Right, Thanks!

PR #1515, opened 2023-06-27
Dont use offset as timezone

MarcoGorelli:
this came up here https://github.com/pola-rs/polars/issues/9586#issuecomment-1609854914

Any chance "UTC" could be used as the time zone, instead of "+00:00"? The latter isn't in the IANA timezone database and can't be parsed by `chrono-tz`


MarcoGorelli:
closing as I don't think this would be the correct fix for polars anyway - sorry for the noise

Issue #1524, opened 2023-07-28
Allowing null buffers


anjakefala:
Hello! =)

In the [spec for the C Data Interface](https://arrow.apache.org/docs/format/CDataInterface.html#c.ArrowArray.buffers) there are allowable scenarios in `Arrow-C++` where null buffers get created:
> for the null bitmap buffer, if [ArrowArray.null_count](https://arrow.apache.org/docs/format/CDataInterface.html#c.ArrowArray.null_count) is 0;
> for any buffer, if the size in bytes of the corresponding buffer would be 0.

The latter case has been bumped into by a couple of people independently:

While writing a numerically typed empty table to a Feather file v1 with memory map on:

```
import polars as pl
import pyarrow as pa

t = pa.table([pa.array([], type=pa.uint64())], names=['a'])

pa.feather.write_feather(t, 'out.ipc', version=1)

tbl = pa.feather.read_table('out.ipc', memory_map=True)
tbl['a'].chunks[0].buffers()                                                                                                       
[None, <pyarrow.Buffer address=0x0 size=0 is_cpu=True is_mutable=False>]  <-- Note the 0x0 address
```

Writing and reading a table through IPC:

```
import pyarrow as pa

def pass_table_through_ipc_and_back(tbl: pa.Table) -> pa.Table:
    sink = pa.BufferOutputStream()

    with pa.ipc.new_file(sink, tbl.schema) as writer:
        writer.write_table(tbl)

    buf = sink.getvalue()

    with pa.ipc.open_file(buf) as reader:
        tbl_out = reader.read_all()

    return tbl_out

tbl = pa.Table.from_pylist([{'a':[]}], schema=pa.schema([pa.field('a', pa.large_list(pa.int64()))]))
tbl_out = pass_table_through_ipc_and_back(tbl)

print(tbl['a'].chunks[0].buffers())
#  [
#    None, 
#    <pyarrow.Buffer address=0x54868020080 size=16 is_cpu=True is_mutable=True>, 
#    None, 
#    <pyarrow.Buffer address=0x1040cd100 size=0 is_cpu=True is_mutable=True>
#  ]
print(tbl_out['a'].chunks[0].buffers())
#  [
#    None, 
#    <pyarrow.Buffer address=0x54868050178 size=16 is_cpu=True is_mutable=False>, 
#    None, 
#    <pyarrow.Buffer address=0x0 size=0 is_cpu=True is_mutable=False>  <---- Note the 0x0 address
[...]

PR #1528, opened 2023-08-06
feat: better error message when reader feather v1

ritchie46:
#1524


src/ffi/array.rs
```diff
@@ -235,9 +235,14 @@ unsafe fn create_buffer<T: NativeType>(
     owner: InternalArrowArray,
     index: usize,
 ) -> Result<Buffer<T>> {
+    let len = buffer_len(array, data_type, index)?;
+
+    if len == 0 {
```
ritchie46, line 240: c++ arrow can create null pointers when arrays are empty. This ensures we don't have to deal with that.

Issue #1527, opened 2023-08-04
Reading Parquet int96 timestamps into arrow timestamps overflows


jaychia:
I am running into an issue where reading Parquet int96 timestamps into arrow2 `timestamp[ns]` arrays can potentially overflow silently, providing wrong results.

This issue was also noted in pyarrow/arrow-cpp [ARROW-12096](https://issues.apache.org/jira/browse/ARROW-12096).

---

Here is a quick example:

First, write a Parquet file with int96 timestamps, where some timestamps are out of range for the `timestamp[ns]` type:

```python
import pyarrow as pa
import pyarrow.parquet as papq
import datetime

# Use PyArrow to write Parquet files with int96 timestamps
table = pa.Table.from_pydict({
    "timestamps": pa.array([
        datetime.datetime(1000, 1, 1),
        datetime.datetime(2000, 1, 1),
        datetime.datetime(3000, 1, 1),
    ], pa.timestamp("ms"))
})
papq.write_table(table, "timestamps.parquet", use_deprecated_int96_timestamps=True, store_schema=False)
```

Reading this file in a unit test results in an overflow panic:

```rust
#[test]
fn read_int96_timestamps() -> Result<()> {
    let timestamp_data = &[
        0x50, 0x41, 0x52, 0x31, 0x15, 0x04, 0x15, 0x48, 0x15, 0x3c, 0x4c, 0x15, 0x06, 0x15, 0x00,
        0x12, 0x00, 0x00, 0x24, 0x00, 0x00, 0x0d, 0x01, 0x08, 0x9f, 0xd5, 0x1f, 0x0d, 0x0a, 0x44,
        0x00, 0x00, 0x59, 0x68, 0x25, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x14,
        0xfb, 0x2a, 0x00, 0x15, 0x00, 0x15, 0x14, 0x15, 0x18, 0x2c, 0x15, 0x06, 0x15, 0x10, 0x15,
        0x06, 0x15, 0x06, 0x1c, 0x00, 0x00, 0x00, 0x0a, 0x24, 0x02, 0x00, 0x00, 0x00, 0x06, 0x01,
        0x02, 0x03, 0x24, 0x00, 0x26, 0x9e, 0x01, 0x1c, 0x15, 0x06, 0x19, 0x35, 0x10, 0x00, 0x06,
[...]

PR #1532, opened 2023-08-07
Correctly coerce Parquet Int96 timestamps into requested TimeUnits

jaychia:
Addresses the first part of issue #1527 

Instead of always naively parsing Parquet Int96 timestamps into `TimeUnit::Nanosecond`, we match on the requested timeunit and perform timeunit-specific parsing

This makes parsing safer when reading Int96 timestamps that are outside of the range of `timestamp[ns]` (e.g. timestamps with dates like the years 1000 or 3000) instead of the current behavior which is to always parse the timestamps with the Nanosecond timeunit, and overflow.


jaychia:
Also happy to move the `int96_to_i64_*s` [logic](https://github.com/jorgecarleitao/arrow2/pull/1532/files#diff-54a247478b1bed5d57ca4083841b89ef7744d1b037b7a5cc91c15b9c9a99de28R394-R430) over into the parquet2 repo!

src/io/parquet/read/deserialize/simple.rs
```diff
@@ -401,17 +439,14 @@ fn timestamp<'a, I: Pages + 'a>(
...
-            (1, _) => Ok(dyn_iter(iden(iter))),
-            (a, true) => Ok(dyn_iter(op(iter, move |x| x * a))),
-            (a, false) => Ok(dyn_iter(op(iter, move |x| x / a))),
+        let conversion_op = match time_unit {
+            TimeUnit::Nanosecond => int96_to_i64_ns,
+            TimeUnit::Microsecond => int96_to_i64_us,
+            TimeUnit::Millisecond => int96_to_i64_ms,
+            TimeUnit::Second => int96_to_i64_s,
         };
+        let iter = primitive::Iter::new(pages, data_type, num_rows, chunk_size, conversion_op);
```
ritchie46, line 448: I think we should branch this iterator on the time unit. Otherwise the inline doesn't do much as it is still runtime dependable.

Can you move this iter creation and return in the match case?

jaychia:
Second PR is up! #1533

ritchie46 (approved):

jaychia:
@ritchie46 any chance you could retrigger CI - Seems to be failing on some flaky token issues?

```
{'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}
```

sundy-li (approved):

jaychia:
Thanks for the stamp @sundy-li!

I had to make a new PR to fix lints since they were re-enabled last week. Would appreciate some help with launching CI :)

sundy-li:
Merged, will create a new pr to fix the clippy.

PR #1535, opened 2023-08-10
feat: handle unaligned pointers in FFI

ritchie46:
Arrow IPC can produce unaligned pointers and this will not be changed upstream.

This ensures we handle them gracefully. If the pointer is aligned, we use it zero copy. Otherwise we memcopy to a new vec.

See issue upstream: https://github.com/apache/arrow/issues/32276

And report of failure on polars: https://github.com/pola-rs/polars/issues/6315

### DIFF
--- a/src/ffi/array.rs
+++ b/src/ffi/array.rs
@@ -213,13 +213,7 @@ unsafe fn get_buffer_ptr<T: NativeType>(
             must have a non-null buffer {index}"
         )));
     }
-    if ptr.align_offset(std::mem::align_of::<T>()) != 0 {
-        return Err(Error::oos(format!(
-            "An ArrowArray of type {data_type:?}
-            must have buffer {index} aligned to type {}",
-            std::any::type_name::<T>()
-        )));
-    }
+
     // note: we can't prove that this pointer is not mutably shared - part of the safety invariant
     Ok(ptr as *mut T)
 }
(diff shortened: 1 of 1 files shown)


src/ffi/array.rs
```diff
@@ -241,12 +235,21 @@ unsafe fn create_buffer<T: NativeType>(
...
-
     let offset = buffer_offset(array, data_type, index);
-    let bytes = Bytes::from_foreign(ptr, len, BytesAllocator::InternalArrowArray(owner));
+    let ptr: *mut T = get_buffer_ptr(array, data_type, index)?;
 
-    Ok(Buffer::from_bytes(bytes).sliced(offset, len - offset))
+    // We have to check alignment.
+    // This is the zero-copy path.
+    if ptr.align_offset(std::mem::align_of::<T>()) == 0 {
+        let bytes = Bytes::from_foreign(ptr, len, BytesAllocator::InternalArrowArray(owner));
```
ritchie46, line 244: Bytes takes ownership of `InternalArrowArray(owner)` ensuring the memory backed by the pointer is not released.

src/ffi/array.rs
```diff
@@ -241,12 +235,21 @@ unsafe fn create_buffer<T: NativeType>(
...
+        Ok(Buffer::from_bytes(bytes).sliced(offset, len - offset))
+    }
+    // This is the path where alignment isn't correct.
+    // We copy the data to a new vec
+    else {
```
ritchie46, line 249: We copy to a new vec. After this `owner` will go out of scope, calling release on the C FFI, dropping backed memory (if needed).

PR #1544, opened 2023-08-16
fix: slice values in list to fixed-size list cast

orlp:
This issue was encountered in https://github.com/pola-rs/polars/issues/10533. In the `ListArray` to `FixedSizeListArray` cast the values slice is never actually sliced to the offset range.


ritchie46:
Yes!

PR #1545, opened 2023-08-19
Fix `any`/`all` for Kleene logic

stinodego:
#### Changes

* `any` and `all` in the `boolean_kleene` module now return an `Option<bool>`. _(The versions in the `boolean` module are untouched.)_
* Added doctests for both.

### DIFF
--- a/src/compute/boolean_kleene.rs
+++ b/src/compute/boolean_kleene.rs
@@ -234,26 +234,70 @@ pub fn and_scalar(array: &BooleanArray, scalar: &BooleanScalar) -> BooleanArray 
     }
 }
 
-/// Returns whether any of the values in the array is `true`
-pub fn any(array: &BooleanArray) -> bool {
+/// Returns whether any of the values in the array are `true`.
+///
+/// The output is unknown (`None`) if the array contains any null values and
+/// no `true` values.
+///
+/// # Example
+///
+/// ```
+/// use arrow2::array::BooleanArray;
+/// use arrow2::compute::boolean_kleene::any;
+///
+/// let a = BooleanArray::from(&[Some(true), Some(false)]);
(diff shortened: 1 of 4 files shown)


src/compute/boolean_kleene.rs
```diff
@@ -234,26 +234,70 @@ pub fn and_scalar(array: &BooleanArray, scalar: &BooleanScalar) -> BooleanArray
...
+/// assert_eq!(all(&b), Some(false));
+/// assert_eq!(all(&c), None);
+/// ```
+pub fn all(array: &BooleanArray) -> Option<bool> {
     if array.is_empty() {
-        true
-    } else if array.null_count() > 0 {
-        false
+        Some(true)
+    } else if array.validity().is_some() {
```
ritchie46, line 293: Can you guard this on `null_count == 0`? There are sometimes cases where we have a validity even though there are no nulls. (due to slicing).

src/compute/boolean_kleene.rs
```diff
@@ -234,26 +234,70 @@ pub fn and_scalar(array: &BooleanArray, scalar: &BooleanScalar) -> BooleanArray
...
+///
+/// assert_eq!(any(&a), Some(true));
+/// assert_eq!(any(&b), Some(false));
+/// assert_eq!(any(&c), None);
+/// ```
+pub fn any(array: &BooleanArray) -> Option<bool> {
     if array.is_empty() {
-        false
+        Some(false)
     } else if array.validity().is_some() {
```
ritchie46, line 259: See comment below.

src/compute/boolean_kleene.rs
stinodego: Sure!

stinodego: Done, also updated the same check in the regular boolean `any`.

PR #1550, opened 2023-08-20
fix: LargeBinary to LargeList should be taken as a special

reswqa:
It seems that the order is important as `LargeBinary` => `LargeList` is a little special.


reswqa:
I have tested this in https://github.com/pola-rs/polars/pull/10625 and that CI has passed.

reswqa:
@ritchie46, would you mind taking a look at this, thanks!

ritchie46 (approved):

PR #1555, opened 2023-08-30
MutableDictionaryArray rewrite: use values stored in the array instead of the hash->hash map

### LINKED ISSUE #1485: Mutable dictionary array is wrong
It stores the hash as a key, but doesn't do any key comparison on insert. Though the probability of collisions is low, it is not zero, so for certain inputs the keys will collide and produce invalid arrays.

### LINKED ISSUE #1554: Writing chunked dictionary arrays to IPC currently impossible due to difference in key maps?
If you have a `MutableDictionaryArray` which you populate and flush once in a while and then create a new one for the next chunks, on the second chunk you try to write you will get:

> InvalidArgumentError("Dictionary replacement detected when writing IPC file format. Arrow IPC files only support a single dictionary for a given field across all batches.")

The problem is:
- You cannot clear the mutable dictionary array resetting its size
- You cannot create a new dict array with the same keys / key map
- There's a comment in the MutableDictionaryArray definition: `invariant: `keys.len() <= values.len()`; with this invariant, how do you solve the above problem? If you were to fix it and allow sharing key maps, you will inevitably end up with an empty array but non-empty key list. Does this invariant actually break anything anywhere if it's violated?

Am I missing something or is there a way to do it? (will be glad to open a PR if there's any suggestions on what's the proper way to fix it)

Might be somewhat related: https://github.com/jorgecarleitao/arrow2/issues/1485

aldanor:
@ritchie46 @jorgecarleitao Here's one way to fix the incorrect current behaviour of `MutableDictionaryArray`: only rely on values actually stored in the values array and don't rely on hash-hash maps (due to potential hash collisions).

This is almost a rewrite of the whole thing, outer API aside, and gets pretty evil (self-referential) at the lower level, but I believe it's pretty sound.

There's probably bits and pieces that may need to be cleaned, types and methods to be renamed, potentially some docstrings added etc (comments welcome), but I figured I'd push the current version as soon as it's working so as to figure whether something like this would be acceptable.

Bench-wise, current main (10k dict insertions):

```
dict_utf8               time:   [239.70 µs 241.27 µs 242.78 µs]
dict_u64                time:   [123.60 µs 123.78 µs 123.98 µs]
```

This branch:

```
dict_utf8               time:   [151.54 µs 151.75 µs 151.97 µs]
                        change: [-37.274% -36.837% -36.411%] (p = 0.00 < 0.05)
dict_u64                time:   [42.466 µs 42.522 µs 42.581 µs]
                        change: [-65.705% -65.627% -65.547%] (p = 0.00 < 0.05)
```

Fixes #1485
Fixes #1554

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ hash_hasher = "^2.0.3"
 simdutf8 = "0.1.4"
 
 # A Rust port of SwissTable
-hashbrown = { version = "0.14", default-features = false, optional = true }
+hashbrown = { version = "0.14", default-features = false, features = ["ahash"] }
 
 # for timezone support
 chrono-tz = { version = "0.8", optional = true }
@@ -243,7 +243,7 @@ compute_merge_sort = ["itertools", "compute_sort"]
 compute_nullif = ["compute_comparison"]
 compute_partition = ["compute_sort"]
 compute_regex_match = ["regex"]
-compute_sort = ["compute_take", "hashbrown"]
+compute_sort = ["compute_take"]
 compute_substring = []
 compute_take = []
 compute_temporal = []
--- a/src/array/dictionary/mod.rs
+++ b/src/array/dictionary/mod.rs
@@ -20,6 +20,7 @@ mod iterator;
 mod mutable;
 use crate::array::specification::check_indexes_unchecked;
 mod typed_iterator;
+mod value_map;
 
 use crate::array::dictionary::typed_iterator::{DictValue, DictionaryValuesIterTyped};
 pub use iterator::*;
--- a/src/array/dictionary/mutable.rs
+++ b/src/array/dictionary/mutable.rs
@@ -67,15 +67,31 @@ impl<K: DictionaryKey, M: MutableArray> MutableDictionaryArray<K, M> {
     /// # Errors
     /// Errors if the array is non-empty.
     pub fn try_empty(values: M) -> Result<Self> {
-        let value_map = ValueMap::<K, M>::try_empty(values)?;
+        Ok(Self::from_value_map(ValueMap::<K, M>::try_empty(values)?))
+    }
+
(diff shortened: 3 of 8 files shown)


sundy-li (approved):
LGTM

aldanor:
Clippy errors: can fix in a later PR along with some deps updates.

Miri errors: not sure can prove it to miri here that it's not UB.

sundy-li:
Great PR. This could improve grouping performance by low cardinality column data.

ritchie46:
@sundy-li you were too fast. I also wanted to review this one. Especially given that MIRI isn't happy. :/

ritchie46:
@aldanor I think we can achieve the same with a `Hashmap<K, usize>` where the `usize` gets the value in the `MutableArray`. 


IAC This will save a lot of (undsafe) code, and make the values of the hashmap smaller as you don't have two pointers (the ref and the index), but only a single index.


In polars we use that something like this:

```rust
        self.inner_map
            .raw_entry_mut()
            .from_hash(hash, |hash_map_key| {
                  let index = hash_map_key.index;
                 
                  self.get_value(index) == insert_value
                }
            })
```

I think we should go in this direction.

aldanor:
@ritchie46 Sure, sounds like a plan. Tbh I didn't even think of the raw hashmap API since I got too used to the fact that it's nightly-only (but not for hashbrown!).

All the "Indexable" stuff is unfortunately still needed since we have no clean way of connecting "stuff we're pushing" to "stuff we're storing" elsewise.

I'll take a look at the raw hashmap approach now and report back.

aldanor:
(I'll try to look into it asap so maybe don't merge the revert API first, so we don't end up with the same code flipped back and forth 3 times...)

aldanor:
@ritchie46 I think you meant `HashMap<usize, K>` (as opposed to `HashMap<K, usize>`) 🙂

But anyways, it does seem to work, I'll push PR shortly.

PR #1559, opened 2023-09-03
Revert "MutableDictionaryArray rewrite: use values stored in the array instead of the hash->hash map"

ritchie46:
Reverts jorgecarleitao/arrow2#1555

This added a lot of unsafe code that I don't think we need (and miri is not happy). I put this revert ready to get the discussion going.


aldanor:
@sundy-li There is an open 'fix the unsafe part only and keep the rest' PR... https://github.com/jorgecarleitao/arrow2/pull/1561

PR #1561, opened 2023-09-03
2nd (safe) rewrite of MutableDictionaryArray

aldanor:
As suggested by @ritchie46, using hashbrown's raw-entry API (flipped `<K, usize>` to `<usize, K>` though).

Need to be very careful to _not_ use any default map API though since it will break map consistency (like .insert(), .collect(), etc).

Benchmarks are roughly the same as in #1555, not slower, not faster; miri should be happy though.

### DIFF
--- a/src/array/dictionary/value_map.rs
+++ b/src/array/dictionary/value_map.rs
@@ -1,0 +1,207 @@
+use std::borrow::Borrow;
+use std::fmt::{self, Debug};
+use std::hash::{Hash, Hasher};
+use std::pin::Pin;
+use std::ptr::NonNull;
+
+use hashbrown::{Equivalent, HashMap};
+
+use crate::array::Array;
+use crate::{
+    array::indexable::{AsIndexed, Indexable},
+    array::MutableArray,
+    datatypes::DataType,
+    error::{Error, Result},
+};
+
+use super::DictionaryKey;
+
+struct NonNullSend<M: ?Sized>(NonNull<M>);
+
+// safety: these pointers are for internal self-referential purposes to pinned array only
+unsafe impl<M> Send for NonNullSend<M> {}
+unsafe impl<M> Sync for NonNullSend<M> {}
+
+impl<M: ?Sized> From<&M> for NonNullSend<M> {
+    #[inline]
+    fn from(reference: &M) -> Self {
+        Self(NonNull::from(reference))
+    }
+}
+
+struct ValueRef<M> {
+    array: NonNullSend<M>,
+    index: usize,
+}
+
+impl<M> ValueRef<M> {
+    #[inline]
+    pub fn new(array: &Pin<Box<M>>, index: usize) -> Self {
+        Self {
+            array: NonNullSend::from(Pin::get_ref(array.as_ref())),
+            index,
+        }
+    }
+
+    #[inline]
+    pub fn get_array(&self) -> &M {
+        // safety: the invariant of the struct
+        unsafe { self.array.0.as_ref() }
+    }
+
+    #[inline]
+    pub unsafe fn get_unchecked(&self) -> M::Value<'_>
+    where
+        M: Indexable,
+    {
+        self.get_array().value_unchecked_at(self.index)
+    }
+
+    #[inline]
+    pub unsafe fn equals(&self, other: &M::Type) -> bool
+    where
+        M: Indexable,
+        M::Type: Eq,
+    {
+        self.get_unchecked().borrow() == other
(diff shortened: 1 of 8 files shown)


aldanor:
(Don't merge yet, I'll add a few very quick tests)

ritchie46:
Nice! Thanks for the quick turnaround. We might also explore making the index type generic, so that we can support all unsigned indexes. 

But that's definitely not a blocker for this PR.

aldanor:
> We might also explore making the index type generic, so that we can support all unsigned indexes and get better cache coherence.

As in `<K, K>`? (it will fail otherwise anyway upon key conversion). Or, hold on, even `<K, ()>` (i.e. technically, a `HashSet`), because keys and values are equal anyways...

You would need to convert is back and forth to `usize` though to get the actual array index, not quite sure if cache coherence offsets potential extra branching (might quickly check though).

aldanor:
@ritchie46 ok, using `<K, ()>` seems to work, there's two minor quirks though:
- in `RawEntryMutBuilder::from_hash()`, we cannot handle errors other than unwrap results, but I think there we can use `unsafe { key.as_usize() }` since we've already converted that key _from_ usize previously, so it implies it can be unsafely converted back?
- `DictionaryKey` has to derive from `Hash` just so that some HashMap bounds are satisfied (which is a bit dumb because we'll never actually hash it, but it doesn't matter that much since those are just ints). Tbh we might as well just derive `NativeType: Hash` since those are just primitives... (or just the DictionaryKey).

Benches seem to improve a bit further by ~10%:

```
dict_utf8               time:   [141.11 µs 141.44 µs 141.82 µs]
                        change: [-11.502% -11.251% -10.984%] (p = 0.00 < 0.05)
                        Performance has improved.

dict_u64                time:   [35.081 µs 35.169 µs 35.259 µs]
                        change: [-8.4965% -8.1730% -7.8369%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Can clean it up a bit and push it too here if it's of interest.

aldanor:
(Pushed the `<K, ()>` commit as well to see if that makes sense)

aldanor:
Since #1559 has been merged, I'll have to rebase this on main again, ok then...

aldanor:
Ok, I've rebased on revert of the revert, should be ok now.

Tests were added.

Unit-value hashmap added = waiting on feedback from @ritchie46 (if I missed something and there's some gaps we can always revert that part if needed).

Otherwise, nothing else left to do in this PR I think unless there's any particular comments, so should be good to go.

src/array/dictionary/value_map.rs
```diff
@@ -0,0 +1,126 @@
...
+        for index in 0..values.len() {
+            let key = K::try_from(index).map_err(|_| Error::Overflow)?;
+            // safety: we only iterate within bounds
+            let value = unsafe { values.value_unchecked_at(index) };
+            let mut hasher = map.hasher().build_hasher();
+            value.borrow().hash(&mut hasher);
+            let hash = hasher.finish();
+            match map.raw_entry_mut().from_hash(hash, |_| true) {
+                RawEntryMut::Occupied(_) => {
+                    return Err(Error::InvalidArgumentError(
```
ritchie46, line 51: Don't we want to allow it and just use the first occurring? No strong opinion on my part, but it might be handy.

ritchie46:
> Since https://github.com/jorgecarleitao/arrow2/pull/1559 has been merged, I'll have to rebase this on main again, ok then... 🤦

@sundy-li is trigger happy. :see_no_evil: :stuck_out_tongue_winking_eye:

src/array/dictionary/value_map.rs
aldanor: No strong preference either! Can change if you'd prefer it this way. I think my point was, there's cases where you really want to ensure that values are unique (e.g. writing chunks).

But it's actually not the best way of doing it anyway for those use cases... I just realized one part that's missing is a method allowing to clone value map and reset the keys (without taking ownership), perhaps something like `to_empty(&self)`.

sundy-li: I prefer to throw an error if the values are not unique.

aldanor: 1 vote versus zero, leave it as ensuring it's unique then?

Unless there's anything else, should be lgtm I guess.

ritchie46:
Thanks for doing the rewrite @aldanor. Good to go.

ritchie46 (approved):

PR #1563, opened 2023-09-05
Parse PrimitiveLogicalType::Unknown as Arrow Null DataType

jaychia:
This PR adds logic to parse any Parquet fields with `Unknown` LogicalType as the Arrow `Null` DataType.


jaychia:
Also, no unit tests on this yet but would love pointers on how to best test this change.

Should I add a new column in the files generated by `parquet_integration/write_parquet.py`?

PR #1564, opened 2023-09-05
Native hashed-hash for MutableDictionaryArray

aldanor:
Here's yet another refactor of mutable dictionary arrays... tested it on some arrays of 100+M size, this time it seems like it actually works correctly.

There's a somewhat flimsy part (or flimsy invariant rather) in assuming that hash of a u64 hash using `HashHasherNative` will be identical to the hash itself if it's built in an endian-dependent way. I've tested it on M1 (LE) and it's correct, I'm guessing it will be correct on BE as well.

// @ritchie46 

Fixes #1561

### DIFF
--- a/src/array/dictionary/value_map.rs
+++ b/src/array/dictionary/value_map.rs
@@ -1,6 +1,6 @@
 use std::borrow::Borrow;
 use std::fmt::{self, Debug};
-use std::hash::{BuildHasher, Hash, Hasher};
+use std::hash::{BuildHasher, BuildHasherDefault, Hash, Hasher};
 
 use hashbrown::hash_map::RawEntryMut;
 use hashbrown::HashMap;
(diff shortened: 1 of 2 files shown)


src/array/dictionary/value_map.rs
```diff
@@ -15,10 +15,61 @@ use crate::{
...
+/// Invariant: hash of a u64 value is always equal to itself.
+#[derive(Copy, Clone, Default)]
+pub struct HashHasherNative(u64);
+
+impl Hasher for HashHasherNative {
+    #[inline]
+    fn write(&mut self, value: &[u8]) {
+        #[cfg(target_endian = "little")]
+        {
+            for (index, item) in value.iter().enumerate().take(8) {
```
aldanor, line 31: This is also sort of an overkill tbh... because we know that `value.len() == 8` always

aldanor:
(Forgot to clarify: the added test currently fails on main.)

ritchie46:
@aldanor I believe we can make it a little bit simple by having an `IdHashMap` 

```rust
type IdHashmap = hashbrown::HashMap<K, V, IdBuildHasher>;
type IdBuildHasher = BuildHasherDefault<IdHasher>;
```

where `IdHasher` is only ever called with `u64`

```rust
pub struct IdHasher {
    hash: u64,
}

impl Hasher for IdHasher {
    fn finish(&self) -> u64 {
        self.hash
    }
    #[inline]
    fn write_u64(&mut self, i: u64) {
        self.hash = i;
    }
}
```

And then we have a `Key` type:

```rust
// hash + value
#[derive(Eq, Copy, Clone)]
struct Key<T: Copy> {
    hash: u64,
    value: T,
}

impl<T: Copy> Hash for Key<T> {
    fn hash<H: Hasher>(&self, state: &mut H) {
        state.write_u64(self.hash)
    }
}

impl<T: Copy + PartialEq> PartialEq for Key<T> {
    fn eq(&self, other: &Self) -> bool {
        self.value == other.value
    }
}
```

And then we are responsible for initializing the `Key` with the hashed value and the rest is done by the hash map.

aldanor:
@ritchie46 Oh, very good point indeed, I totally forgot about the fact that `write_u64()` exists.

`value == T` is not going to work though, indexed values can have lifetimes (e.g. strings), and now you will need to not only link the 3 types "things we can push", "things we can index", "connecting things we can push to things we can index", but also "things we can push or index connecting to an owned type that we can store aside". The trait hierarchy is already pretty ugly... (look at AsIndexed, Indexable stopgaps).

ritchie46:
> value == T is not going to work as is though, indexed values can have lifetimes (e.g. strings), 

I think we can do with using the index. That's what we do in polars for strings.

```rust
#[derive(Copy, Clone)]
struct Key {
    pub(super) hash: u64,
    pub(super) idx: u32,
}
```

And then the comparison is something like this:

```rust
    table.raw_entry_mut().from_hash(h, |key| {
        // first compare the hash before we incur the cache miss
        key.hash == h && {
            let idx = key.idx as usize;
            unsafe { keys.get_unchecked_release(idx).as_deref() == key_val }
        }
    })
```

aldanor:
@ritchie46 That's almost exactly what we already have here :) (aside from additional hash equality check which I can add too if it improves things)

src/array/dictionary/value_map.rs
```diff
@@ -91,24 +141,22 @@ impl<K: DictionaryKey, M: MutableArray> ValueMap<K, M> {
...
             }) {
-                RawEntryMut::Occupied(entry) => *entry.key(),
+                RawEntryMut::Occupied(entry) => entry.key().key,
                 RawEntryMut::Vacant(entry) => {
                     let index = self.values.len();
                     let key = K::try_from(index).map_err(|_| Error::Overflow)?;
-                    entry.insert_hashed_nocheck(hash, key, ()); // NB: don't use .insert() here!
+                    entry.insert_hashed_nocheck(hash, Hashed { hash, key }, ()); // NB: don't use .insert() here!
                     push(&mut self.values, value)?;
+                    assert_eq!(self.values.len(), index + 1);
```
ritchie46, line 159: Shall we make this a `debug_assert`? Don't see this going wrong.

ritchie46:
> @ritchie46 That's almost exactly what we already have here :) (aside from additional hash equality check which I can add too if it improves things; and with key being of `K` type)
> 
> // it's actually quite hilarious, I think, that I had to go through all circles of hell in the last prs here, introduce more bugs while fixing previous bugs, then optimize it while fixing them and finally arrive to pretty much precisely the polars solution 😆

I see now. Aside from the naming very similar indeed. :) I have left one comment about the assert. Other than that I think this is good to go. :+1:

src/array/dictionary/value_map.rs
aldanor: Good catch; I think this was actually debug_assert_eq in one of the previous iterations (and was meant to be ) but got messed during all the rewrites.

ritchie46 (approved):

Issue #1556, opened 2023-08-30
Incorrect nullability inferred for nested parquet schema


jhorstmann:
I'm trying to read a parquet file that contains a struct inside a list using pola-rs and am getting null values for each element. I think I can track down the issue to the schema conversion from parquet to arrow.

The `parquet_to_arrow_schema` function tries to set the `nullable` flag of `Field` according to the parquet repetition levels. That flag is then used via the `InitNested` `enum` to calculate the level at which data is valid.

My message schema looks like the following:

```
message eventlog {
  REQUIRED group events (LIST) {
    REPEATED group array {
      REQUIRED BYTE_ARRAY event_name (STRING);
      REQUIRED INT64 event_time (TIMESTAMP(MILLIS,true));
    }
  }
}
```

And I would expect all fields having the `is_nullable` flag set to `false`. Instead the `array` field is marked as nullable. I think the issue can also be shown with the example schemas from [parquet-format/LogicalTypes.md](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#nested-types) which are tested in `test_parquet_lists`. The comments there do not match the assertions. For example:

```
        // // List<String> (list nullable, elements non-null)
        // optional group my_list (LIST) {
        //   repeated group element {
        //     required binary str (UTF8);
        //   };
        // }
        {
            arrow_fields.push(Field::new(
                "my_list",
                DataType::List(Box::new(Field::new("element", DataType::Utf8, true))),
                true,
            ));
        }

        // // List<Integer> (nullable list, non-null elements)
        // optional group my_list (LIST) {
        //   repeated int32 element;
        // }
        {
            arrow_fields.push(Field::new(
                "my_list",
                DataType::List(Box::new(Field::new("element", DataType::Int32, true))),
                true,
            ));
        }
```

According to the comment and documentation `element` should not be nullable in both examples.

I do not yet have a standalone test case and example file, but will try to provide one later.

jhorstmann:
Sample file: [eventlog.zip](https://github.com/jorgecarleitao/arrow2/files/12486226/eventlog.zip), generated with the java implementation from `parquet-mr`.

The [`parquet-read` tool from `arrow-rs`](https://github.com/apache/arrow-rs/blob/master/parquet/src/bin/parquet-read.rs) reads this without problems:

```
[...]

PR #1565, opened 2023-09-07
Fix the inferred nullability when converting a nested parquet schema to arrow

jhorstmann:
This allows the `parquet_read` example to correctly read the nested data attached to issue #1556 and also makes several test assertions match the comments above.


jhorstmann:
@ritchie46 or @sundy-li as the parquet expert, could you take a look at this PR? Or should I rather re-open this against pola-rs/nano-arrow?

ritchie46:
> Or should I rather re-open this against pola-rs/nano-arrow?

@jhorstmann could you please do so? That would get the patched in polars as well.

PR #1566, opened 2023-09-08
Add `array::downcast_ref` and `array::downcast_mut`

aldanor:


src/array/mod.rs
```diff
@@ -703,6 +709,92 @@ impl<'a> AsRef<(dyn Array + 'a)> for dyn Array {
     }
 }
 
+mod downcast {
+    use super::{Array, DataType, MutableArray};
+    use std::any::Any;
+
+    /// Arrays that can be downcasted to a concrete type ([`Array`] and [`MutableArray`]).
+    pub trait DowncastableArray {
```
aldanor, line 717: Actually, a more elegant name might be `ArrayAny`, I think.

Cargo.toml
```diff
@@ -100,6 +100,9 @@ odbc-api = { version = "0.36", optional = true }
 # Faster hashing
 ahash = "0.8"
 
+# workaround for pre-polonius borrow checker
+polonius-the-crab = "0.3"
```
ritchie46, line 104: I am really not sure about adding a new dependency for this.

ritchie46 (commented):
I am not sure about this. You can say something like this for any `Option` returning method. To me this feels something that needs to be resolved by the consumer of the arrow2 crate.

Cargo.toml
aldanor: Removed the dependency but that requires a line of unsafe code (however that crate sort of does the same anyway). It's a borrow-checker issue/bug where the mutable borrow proliferates past the return point, so nothing unsound there.

aldanor:
> I am not sure about this. You can say something like this for any `Option` returning method. To me this feels something that needs to be resolved by the consumer of the arrow2 crate.

It's your call, but in my recent experiences as a daily arrow2 user, when you're dealing when reading lots of box-arrays from various sources (e.g. dealing with struct-like data and the like), it's downcasts all over the place. If you want to avoid unwraps, you will have to `.ok_or_else()` every single line to provide context on what just happened. Or just litter the code with unwraps because it's too much hassle otherwise.

The core difference between "just let the user figure it out, it's an option" here is that the array itself (both `Array` and `MutableArray` in fact) actually have access to the context, both "what you were trying to cast to" and "what the actual datatype is", because of `.data_type()`. So, we can assemble a nice error message just once here; chances are, if the user wanted to assemble an error message, they would do almost precisely the same (i.e., use the actual data type of the underlying array). This way, it's just `?` in any Result-like contexts as long as your error type is convertible from arrow2's (which it probably already is). There's not many other Option-yielding core APIs aside from downcasts in the library, this is the only bit really sticking out.

A minor side effect here is that it unifies downcasting of both array/mutable-array, but that's not the most important part.

PR #1567, opened 2023-09-09
fix: More types supports cast to LargeList

reswqa:
More types(e.g. `LargeUtf8`) should also supports cast to `LargeList`. The only special branch is `LargeBinary to LargeList`.


reswqa:
Seems that CI is unstable, but I assume it's not related to this change.

ritchie46:
Great!

PR #1575, opened 2023-10-03
fix: fix deserialization of parquets with large string list columns causing stack overflow

ByteNybbler:
This mirrors a fix I did within polars: https://github.com/pola-rs/polars/pull/11471


sundy-li:
lint is unhappy.

PR #1579, opened 2023-10-13
add new_constant for Bitmap

Dousir9:


sundy-li:
how about naming `new_constant`

sundy-li (approved):

PR #1589, opened 2023-11-06
Added try_new for mutable data types

aryanshomray:
This PR is made to have consistent in try_new() constructors that accept DataType in a similar way, for all mutable array types

### DIFF
--- a/src/array/dictionary/mutable.rs
+++ b/src/array/dictionary/mutable.rs
@@ -2,6 +2,7 @@ use std::hash::Hash;
 use std::sync::Arc;
 
 use crate::array::indexable::{AsIndexed, Indexable};
+use crate::array::specification::{check_indexes, check_indexes_unchecked};
 use crate::{
     array::{primitive::MutablePrimitiveArray, Array, MutableArray, TryExtend, TryPush},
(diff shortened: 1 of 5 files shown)


src/array/dictionary/mutable.rs
```diff
@@ -63,6 +63,22 @@ impl<K: DictionaryKey, M: MutableArray + Default> Default for MutableDictionaryA
...
+    /// Creates an [`MutableDictionaryArray`] from a given keys array and values array.
+    /// # Errors
+    /// Errors if the keys array have duplicates or if the keys are out of bounds of the values array.
+    pub fn try_new(data_type: DataType, keys: MutablePrimitiveArray<K>, values: M) -> Result<Self>
+    where
+        M: Indexable,
+        M::Type: Eq + Hash,
+    {
+        let map = ValueMap::<K, M>::from_values(values)?;
+        Ok(Self {
```
aldanor, line 86: Hm... shouldn't `data_type` be validated here?

aldanor: No, I think what you need to check is that ValueMap's resulting data type is the same as data type being passed in?

